PR #10, opened 2018-10-11
Add configurable flow element crown

### LINKED ISSUE #1: Flow Element Crown
When you click on a flow element on the canvas two icons should appear to the right of the element. These icons are a flow segment and a trash can. 

Clicking on the flow element will allow for drawing a line from the current element to another.

Clicking on the trash can will remove that element.

screenshot 2018-10-08 14 58 22

chipit24:
Fixes #1.

crown


src/mixins/crownConfig.js
```diff
@@ -0,0 +1,68 @@
...
+        button.addTo(this.graph);
+        const { x, width } = this.shape.findView(this.paper).$el[0].getBBox();
+        button.position(x + width + buttonMargin, yOffset, { parentRelative: true });
+      });
+    }
+  },
+  mounted() {
+    this.$nextTick(this.configureCrown);
+  }
+};
```
chipit24: This is a pretty beefy mixin. I noticed a lot of duplicate functionality between the different node components (`task.vue`, `startEvent.vue`, etc.). Creating a base component (which would also include this mixin functionality) would be a nice refactor 👍.

PR #16, opened 2018-10-17
Add Text Annotation as a draggable element

### LINKED ISSUE #4: Text Annotation Element
When I click on the text icon in the flow menu, I want to be able to drag a text annotation onto the canvas. I need to be able to then click on the flow element and edit the text in the inspector pane.

### LINKED ISSUE #13: Add ability to connect elements via an association flow
Utilizing the crown and clicking, the ability to connect elements via association flow (dotted line) needs to be added. Ideally, this is done by clicking the flow icon in a crown on the source object, and then clicking on the target item.

Please refer to the camunda modeler for example of this.

**Association Flow Restrictions**
- Any element in the modeler can be connected to any element but text annotations or lanes

tuantruong625:
Fixes #4 , Fixes #13


tdondich:
1) Fails to resize the text container on the left side. If the text goes multi-line, then it's important that the left side resizes to scale height of the text.  

2) When creating a new text annotation, the text says "New Text Annotation" but clicking on the element shows an empty Annotation Description field.  

3) If the text annotation description is very short, such as "test", the crown is very far right of the text element ( no screenshot provided, reproduction is simply changing a text annotation to say "test" and then unclick and click the element.
modeler 2

PR #22, opened 2018-10-18
Handle sequence flows in parse

### LINKED ISSUE #21: When loading a bpmn file, the operation immediately tries to link all flows
When importing a bpmn file with existing flow elements, the diagram renders but the cursor immediately is in the step of connecting ALL flows.

chipit24:
Fixes #21.

### DIFF
--- a/src/ModelerApp.vue
+++ b/src/ModelerApp.vue
@@ -83,7 +83,7 @@ export default {
       });
     },
     handleUpload(files) {
-      let reader = new FileReader();
+      const reader = new FileReader();
       reader.onloadend = () => {
         this.$refs.modeler.loadXML(reader.result);
         this.$refs.uploadmodal.hide();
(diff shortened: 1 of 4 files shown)


tdondich:
When a bpmn file is imported, the sequence flow elements in the diagram have coordinates.  These coordinates usually lay on a border area of the element in question (bottom middle, or top middle of task, etc).  It's important that when we do parse, we determine which anchor point the coordinate is closest to and use that anchor point when rendering.  This means that elements should have multiple anchor points.

So when parsing, the intention should be that the anchor point used is closest to the specified coordinates in the bpmn file.

When creating new flow elements or when dragging elements around, then the coordinates/anchors for the flow can be updated/changed to the anchor points that are CLOSEST to the target element.

I hope that makes sense.  I parsed a bpmn file and I'm attaching a screenshot of the obvious problems.
modeler 2

chipit24:
I tested a few different approaches and it's looking better now:
screen shot 2018-10-22 at 11 18 39 am

chipit24:
Tweaked router/anchor/connector:
screen shot 2018-10-22 at 3 02 47 pm

PR #28, opened 2018-10-24
Tasks should be able to connect to exclusive gateways

### LINKED ISSUE #20: Tasks should be able to connect to exclusive gateways
Connecting via flow needs to be allowed from task to exclusive gateway.  And exclusive gateway also needs to allow connecting to tasks.

However, the exclusive gateway can only have two outbound flows and only one inbound flow. We should enforce these rules.

@enriquecolosa , do you agree with these rules? What should the behavior be if you attempt to do something against these rules?

tuantruong625:
Fixes #20

Exclusive Gateway Node
 - able to edit the body of text using the form input
 - handle validation rules & error states based on @enriquecolosa comment


enriquecolosa:
@tuantruong625 basic functionality is there but I found some weird behavior, when you try to connect from a gateway to a start event (invalid connection), target element turns red as expected but gateway turns red also and the later keeps this state. I noticed that the flow icon in the crown is red as well.
screen shot 2018-10-25 at 8 49 24 am
Change of colors must affect only to target elements, not source elements or crown icons.

src/components/nodes/sequenceFlow/sequenceFlow.vue
```diff
@@ -20,52 +20,17 @@ export default {
       shape: null,
       definition: null,
       sourceShape: null,
+      target: null,
+      validNodeColor: '#dffdd0',
+      invalidNodeColor: '#fae0e6',
+      defaultNodeColor: '#fff',
+      gatewayDirectionOptions: { Diverging: 'Diverging', Converging: 'Converging' },
```
chipit24: Instead of the `sequenceFlow` component owning this data, I think it should be in the gateway config.

`/src/components/nodes/exclusiveGateway/index.js` still has the hard-coded strings for `"Diverging"` and `"Converging"`. You could place the direction options in there, and export it, so the same values can be used in the gateway component and also imported and used in `sequenceFlow.vue`.

src/components/nodes/sequenceFlow/sequenceFlow.vue
```diff
@@ -92,67 +57,109 @@ export default {
...
         label: { cursor: 'move' },
+
+        ".body": { fill: `${this.defaultNodeColor}`, cursor: 'move' },
+        ".label": { cursor: 'move' },
       });
 
       this.shape.listenTo(this.sourceShape, 'change:position', this.updateWaypoints);
       this.shape.listenTo(targetShape, 'change:position', this.updateWaypoints);
     },
+    checkExclusiveGateway() {
```
chipit24: When someone looks at this code, they may ask "Check what about exclusive gateway?" I think a better function name may be `isValidGatewayConnection`; this also hints that this function will return a boolean. Or you could add a comment here (which you might want to do anyway since gateway validation may seem confusing).

src/components/nodes/sequenceFlow/sequenceFlow.vue
```diff
@@ -92,67 +57,109 @@ export default {
...
+      const definition = this.target.component.node.definition;
+      const gatewayDirection = definition.get('gatewayDirection');
+      const incoming = definition.get('incoming');
+      const outgoing = definition.get('outgoing');
+      const incomingFlowCount = incoming.length;
```
chipit24: Since some variables are only used once, and `get('incoming')` and `get('outgoing')` will always return an array, you can cut down on intermediate variables and this may help with readability.

Also, `Diverging` seems to be the default option, so we can further clean this up by placing the non-default flow first and using an early return.

```
const definition = this.target.component.node.definition;
const gatewayDirection = definition.get('gatewayDirection');
const incomingFlowCount = definition.get('incoming').length;
const outgoingFlowCount = definition.get('outgoing').length;

if (gatewayDirection == this.gatewayDirectionOptions.Converging) {
  return true;
}

if (incomingFlowCount === 0) {
  return true;
}

if (outgoingFlowCount > 0) {
  return true;
}

return false;
```

If there were more than two options, we can worry about being more explicit, and possibly use a `switch` statement, but I think this is nice considering the two options right now.

PR #32, opened 2018-10-29
Feature/30

tdondich:
Big change to how things are registered.  Now there's a registerNode on the modeler component, expecting certain structures that contain control definition, component definition, etc.


src/ModelerApp.vue
```diff
@@ -69,16 +78,17 @@ export default {
     `;
 
     this.$refs.modeler.loadXML(blank);
+    for(var node of initialNodes) {
+      this.$refs.modeler.registerNode(node)
```
chipit24: Instead of reaching into the child through a ref, why not handle this in the `Modeler` component directly?

src/ModelerApp.vue
```diff
@@ -36,7 +36,17 @@ import {
 } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/vue-fontawesome";
 
-import InitialControls from "./controls";
+// Our initial node types
```
chipit24: The variable name `initialNodes`, to me, indicates that this list might be changed (like a "default value"), but it seems to remain constant. Would a more appropriate name be `nodes`?

I try to avoid mixing ES module and CommonJS syntax; maybe this list could be moved into `/src/components/nodes/index.js` and exported as an array from there?

src/ModelerApp.vue
tdondich: The reason being that the modeler component is embedded into bpm4.  In that situation, different nodes can be registered instead of the ones that are part of the package.  This is how overriding the inspector configs for certain nodes when the modeler is used in bpm4 is done.

PR #36, opened 2018-11-02
Reset text annotation height when text input is edited then empty

### LINKED ISSUE #35: Text Annotation height does not reset when deleting input text.
screen shot 2018-11-02 at 10 33 33 am

screen shot 2018-11-02 at 10 34 09 am

To recreate this bug

- Create New Text Annotation
- Modify the body of text using text input
- Select all text in input
- Delete

Results in the height of text annotation node not resetting to the default height.

tuantruong625:
Fixes #35

### DIFF
--- a/src/components/nodes/textAnnotation/textAnnotation.vue
+++ b/src/components/nodes/textAnnotation/textAnnotation.vue
@@ -7,6 +7,10 @@
 import joint from 'jointjs';
 import connectIcon from '@/assets/connect-elements.svg';
 import crownConfig from '@/mixins/crownConfig';
+import { highlightPadding } from '@/mixins/crownConfig';
+import { textAnnotationHeight } from './index';
+
+const labelPadding = 15;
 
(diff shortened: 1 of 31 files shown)


src/components/nodes/textAnnotation/textAnnotation.vue
```diff
@@ -31,13 +31,13 @@ export default {
       return this.shape;
     },
     updateShape() {
-      let bounds = this.node.diagram.bounds;
       const { height } = this.shape.findView(this.paper).getBBox();
+      const refPoints = `25 ${height} 3 ${height} 3 3 25 3`;
+      let bounds = this.node.diagram.bounds;
+      let textAnnotationLength = this.node.definition.get('text').length;
```
chipit24: Since you're not re-assigning these variables, it's safer to use `const` instead of `let`. I'll eventually add this to our ESLint rules.

PR #39, opened 2018-11-06
Exclusive Gateways

### LINKED ISSUE #38: Exclusive Gateways
Exclusive gateways are represented with this shape:
 
image

An exclusive gateway evaluates the state of the business process and, based on the condition, breaks the flow into one of the two or more mutually exclusive paths.

The inspector for the gateway itself should only contain a name and id field.

The sequence flow lines coming off of the gateway are special as they have an additional parameter not present in a normal sequence flow. These flows contain an EXPRESSION. An expression a small bit of code using the FEEL syntax that will evaluate the data received, and make a decision to route to the subsequent elements based on the evaluation. 

The sidebar for these expression sequence flow line should add a required text box labeled Expression. 

screen shot 2018-11-02 at 12 09 28 pm

tuantruong625:
Fixes #38 

**Basic Functionally implemented.**
- Generates correct XML file
- Removed gatewayDirection on gateway nodes
- Dynamically renders expression input when sourceRef is exclusiveGateway node

**Key Modified Files**
- Modeler.vue
- sequenceFlow/index.js
- mixins/crownConfig.js

*High number of file changes because of lint fixes


caleeli:
about the issue reported by @tdondich It was solved

https://youtu.be/PHAiHhqCbl0

PR #54, opened 2018-11-15
Pool / Lane Expanding & Element Overlapping

### LINKED ISSUE #52: Clicking on lane or pool after adding a lane causes lane and pool to expand
Steps to reproduce:
 * Drag a pool onto the grid.
 * Add a lane to the pool.
 * Click on the lane that was just added, and it expands.

tuantruong625:
Fixes #52

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,5 +1,3 @@
 module.exports = {
-  presets: [
-    '@vue/app'
-  ]
-}
+  presets: ['@vue/app'],
+};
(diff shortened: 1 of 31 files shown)


chipit24:
The fixes in here were added as part of other PRs that were based off of this one.

@caleeli You are correct – the lockfile shouldn't change unless packages were changed (updated, added, or removed). The only thing that appeared to be modified was the protocol for the resolved property, which seems to have been a bug.

PR #59, opened 2018-11-22
Feature/57/refactor

chipit24:
* Moved `modeler-start` and `modeler-init` events from `ModelerApp.vue` to `Modeler.vue`; this ensures the "source of truth" for hooking into the modeler stays with the Modeler app component.
* Added globlas (EventBus, apiClient) to better model how the app will work when implemented into the BPM product.
* Passed specific functions to events instead of the entire component instance. This makes the scope of the events more clear, and prevents accidentally modifying the Modeler instance in dangerous ways.
* Broke extensions into their own files to more easily manage (instead of bloating the `ModelerApp.vue` file).


chipit24:
Note: For these changes to work well with `bpm`, the `mounted` hook from `ModelerApp.vue` will have to be removed to prevent nodes from being registered twice:
https://github.com/ProcessMaker/bpm/blob/a3a40af887b7f6922281e0e91124612dfb479885/resources/js/processes/modeler/components/ModelerApp.vue#L58-L63

PR #93, opened 2018-12-04
Association Renders.

### LINKED ISSUE #86: The association is not working properly
It has issues during rendering and are not loaded during the upload of process.

image

tuantruong625:
Fixes #86


src/components/nodes/association/association.vue
```diff
@@ -17,48 +19,55 @@ export default {
...
-          'processmaker-modeler-start-event',
-          'processmaker-modeler-end-event',
-          'processmaker-modeler-exclusive-gateway',
-          'processmaker-modeler-inclusive-gateway',
-          'processmaker-modeler-parallel-gateway',
-        ],
-      },
+      target: null,
+      anchorPadding: 25,
+      laneShape: LaneShape.id,
```
chipit24: There's no need to add this to the component data. The name is also misleading as it's not referring to the jointjs shape/model but a string ID.

Import it above as:
```
import { id as laneId } from '../poolLane';
```

And reference it directly as `laneId` in the component below.

chipit24:
The cursor doesn't reset after completing the association (it always shows the "not allowed" pointer):
cursor_bug

PR #173, opened 2019-01-29
Fix Sequence Flow Rendering issues when positioning connected nodes.

tuantruong625:
Fixes #172 

Removed padding on the anchor points prevents the routers to break which in some cases would cause the overlapping in the flows. Switch from the **manhattan** router option to the **orthogonal** since the options were not needed from it.

Issue #172, opened 2019-01-29
Sequence Flow Visual Rendering Issues.


tuantruong625:
Steps to Reproduce
1. Drag Task
2. Connect Start Event to task
3. Drag Exclusive Gateway
4. Connect Task to Gateway
5. Drag Task
6. Connect Gateway to Task
7. Drag End Event
8. Connect Task to End Event
8. Reposition the end event to line up with the bottom of task

<img width="663" alt="screen shot 2019-01-30 at 2 50 12 pm" src="https://user-images.githubusercontent.com/26545455/52008306-6043ce80-249e-11e9-8188-a94eea11ce95.png">
 

![image](https://user-images.githubusercontent.com/26545455/51917453-935a6500-23ad-11e9-9ee4-2e2a9581329c.png)

[LeaveAbsenceRequest.bpmn.txt](https://github.com/ProcessMaker/modeler/files/2808266/LeaveAbsenceRequest.bpmn.txt)
Upload **LeaveAbsenceRequest.bpmn.txt**  using our modeler to reproduce the overlapping issue with sequence flow.

PR #176, opened 2019-01-30
Recalculate element drop position relative to the paper origin

### LINKED ISSUE #168: Importing xml file renders improperly
(This is different from #141)

1. Open modeler
2. Attach the following file: [bpmnProcess.xml.txt](https://github.com/ProcessMaker/modeler/files/2789405/bpmnProcess.xml.txt)

What you should see:
image

What displays:
image

Unsure why this occurs, all browsers.

tuantruong625:
Root cause: when dropping an element inside a pool in Firefox, the calculated offset is relative to the pool and the drop is relative to the paper origin. Interestingly, Firefox handles the drop event more closely to spec for the offset attributes.

Fix:
Set the origin of the page, and use it to calculate the position of the drop. 

Fixes #168

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -359,8 +359,9 @@ export default {
       const diagram = this.nodeRegistry[type].diagram(this.moddle);
 
       // Handle transform
-      diagram.bounds.x = event.offsetX - this.paper.options.origin.x;
-      diagram.bounds.y = event.offsetY - this.paper.options.origin.y;
+      const paperOrigin = this.paper.localToPagePoint(0, 0);
+      diagram.bounds.x = event.clientX - paperOrigin.x;
+      diagram.bounds.y = event.clientY - paperOrigin.y;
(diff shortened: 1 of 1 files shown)


PMQABen:
Well, it works, but now the flow arrows start to look bad as you compress down the model:
image

chipit24:
@PMQABen This issue is unrelated to sequence flows; the code doesn't modify any of the sequence flow logic.

It fixes the issue described by the video you posted in https://github.com/ProcessMaker/modeler/issues/168. I believe the issue with sequence flows is addressed in another PR: https://github.com/ProcessMaker/modeler/pull/173.

The flow arrows behave the same as they do on the `develop` branch (which is different from the latest published build currently used in bpm–what you seem to be comparing it to). If you think you found a new issue different from https://github.com/ProcessMaker/modeler/issues/172, I think it's best it goes into a new issue to be tracked separately.

PR #185, opened 2019-02-04
Add Test to update node names, create lanes, and create a process.

### LINKED ISSUE #184: Add Additional Test to Modeler
Additional Test

- Updating Name for every node
- Creating Lanes
- Creating a process with task, pools, connecting elements, and exporting a valid xml

To run test
`npm run test`

tuantruong625:
These test will provide coverage for updating node names, creating lanes, and a process.
The process contains tasks, pools, and connecting elements.

Fixes #184


src/components/controls.vue
```diff
@@ -8,8 +8,9 @@
         :key="index"
         :transfer-data="{type: control.type}"
         v-if="control.label.toLowerCase().includes(filterQuery.toLowerCase())"
+        :class="[ control.type ]"
```
chipit24: Was this added so the element can be selected in tests? If so, why not use the same `data-test` attribute as above?

tests/e2e/specs/Modeler.spec.js
```diff
@@ -0,0 +1,216 @@
...
+};
+
+const dragFromSourceToDest = (source, dest, position) => {
+  const dataTransfer = new DataTransfer();
+  cy.get(source).trigger('dragstart', { dataTransfer });
+  cy.get(dest).trigger('dragenter');
+  cy.get(dest).trigger('drop', { offsetX: position.x, offsetY: position.y });
+};
+
+function coordinateGenerator() {
```
chipit24: Could you add a comment for this function? Or, better yet, it may benefit from a slight refactor. It seems it's always used the same way:

```js
{ x: coordinateGenerator(), y: coordinateGenerator() }
```

It could be renamed to `generateRandomPoint` and return an object with the shape `{ x: Number, y: Number }`. And possibly add constraints based on the paper dimensions, as `500` and `150` are magic numbers.

tests/e2e/specs/Modeler.spec.js
```diff
@@ -0,0 +1,216 @@
...
+    <bpmndi:BPMNPlane id="BPMNPlane_1" bpmnElement="Process_1">
+      <bpmndi:BPMNShape id="node_2_di" bpmnElement="node_2">
+        <dc:Bounds x="150" y="150" width="36" height="36" />
+      </bpmndi:BPMNShape>
+    </bpmndi:BPMNPlane>
+  </bpmndi:BPMNDiagram>
+</bpmn:definitions>`;
+};
+
+const dragFromSourceToDest = (source, dest, position) => {
```
chipit24: Here, you're assigning an anonymous arrow function to a variable, while below you're explicitly defining a named function. To keep things consistent, you should use `function dragFromSourceToDest { ...` here as well (or change the below functions to use `const .. = () =>`).

tests/e2e/specs/Modeler.spec.js
```diff
@@ -0,0 +1,216 @@
...
+    cy.get('.modeler').children().should('have.length', 2);
+
+    nodeTypes.forEach(type => {
+      dragFromSourceToDest(
+        type,
+        '.paper-container',
+        { x: coordinateGenerator(), y: coordinateGenerator()},
+      );
+    });
+    cy.get('.modeler').children().should('have.length', 8);
```
chipit24: I think this test should be cleaned up a tiny bit. It's not clear why, when no nodes have been dragged to the paper, that `.modeler` should have 2 children. I think at the very least these numbers should be put into variables:

```js
const emptyChildrenCount = 2;
cy.get('.modeler').children().should('have.length', emptyChildrenCount);
...
cy.get('.modeler').children().should('have.length', emptyChildrenCount + nodeTypes.length);
```

tests/e2e/specs/Modeler.spec.js
```diff
@@ -0,0 +1,216 @@
...
+
+    dragFromSourceToDest(
+      '.processmaker-modeler-pool',
+      '.paper-container',
+      { x: 100, y: 100},
+    );
+    cy.get(poolSelector).click();
+    cy.get(topLaneSeletor).click({ force: true });
+    cy.get(bottomLaneSeletor).click({ force: true });
+
```
chipit24: Extra space.

tests/e2e/specs/Modeler.spec.js
```diff
@@ -0,0 +1,216 @@
...
+    const bottomLaneSeletor = '#v-29';
+
+    dragFromSourceToDest(
+      '.processmaker-modeler-pool',
+      '.paper-container',
+      { x: 100, y: 100},
+    );
+    cy.get(poolSelector).click();
+    cy.get(topLaneSeletor).click({ force: true });
+    cy.get(bottomLaneSeletor).click({ force: true });
```
chipit24: What's `force: true` doing here?

tests/e2e/specs/Modeler.spec.js
```diff
@@ -0,0 +1,216 @@
...
+      '.paper-container',
+      { x: coordinateGenerator(), y: coordinateGenerator()},
+    );
+
+    cy.get('.joint-viewport').find('.joint-type-processmaker-components-nodes-task').click({force: true});
+    typeIntoTextInput('[name=\'name\']', 'testing');
+    cy.get('[name=\'name\']').should('have.value', 'testing');
+  });
+
+  it('Update exclusive gateway name', ()=> {
```
chipit24: In some places it's `()=>` and in others it's `() =>` (space between `()` and `=>` in arrow functions). I think this should match the style in the rest of the app and include the space. This stuff should go into our ESLint config though (thought it was already there) so we don't have to think about it.

tests/e2e/specs/Modeler.spec.js
```diff
@@ -0,0 +1,216 @@
...
+    dragFromSourceToDest(
+      '.processmaker-modeler-exclusive-gateway',
+      '.paper-container',
+      { x: coordinateGenerator(), y: coordinateGenerator()},
+    );
+
+    cy.get('.joint-viewport').find('.joint-type-processmaker-modeler-bpmn-exclusivegateway').click({force: true});
+    typeIntoTextInput('[name=\'name\']', 'testing');
+    cy.get('[name=\'name\']').should('have.value', 'testing');
```
chipit24: I think using a variable here could help clean things up a bit too:

```js
const testString = `testing`;
typeIntoTextInput('[name=\'name\']', testString);
cy.get('[name=\'name\']').should('have.value', testString);
```

tests/e2e/specs/Modeler.spec.js
```diff
@@ -0,0 +1,216 @@
...
+      dragFromSourceToDest(
+        type,
+        '.paper-container',
+        { x: coordinateGenerator(), y: coordinateGenerator()},
+      );
+    });
+    cy.get('.modeler').children().should('have.length', 8);
+  });
+
+  it('Creates top and bottom lane', () => {
```
chipit24: Bit of a nitpick, but I think we should stick to a consistent naming convention for tests. Some are named to fit the phrase "This test will ..." while others are named to fit "This test ...". I like the latter (how you named this test). It could be "Creates top and bottom lanes" or "Can create top and bottom lanes"

tests/e2e/specs/Modeler.spec.js
```diff
@@ -0,0 +1,216 @@
...
+    const taskConnectorSelectorThree = '#v-71';
+    const endEventSelector = '#v-85';
+    const poolSelector = '#v-101';
+
+    cy.get('.modeler').children().should('have.length', 2);
+
+    dragFromSourceToDest(
+      '.processmaker-modeler-task',
+      '.paper-container',
+      { x: 300, y: 200},
```
chipit24: Not using `coordinateGenerator` for these? :p

src/components/controls.vue
tuantruong625: Binding a class to each nodes allows us to select a node from the tool panel. If we used `data-test` to select each node for tests, we would have to manually attach `data-test="nameOfName"` to each node.

tests/e2e/specs/Modeler.spec.js
tuantruong625: Forces the action, disables waiting for actionability.

tests/e2e/specs/Modeler.spec.js
tuantruong625: I use the `x, y` of the shape to connect other shapes. If we used to `generateRandomPoint()` here we would not know what the `x, y` would be for the connection.

PR #190, opened 2019-02-05
Add direction attributes to association flow 

### LINKED ISSUE #181: What are valid association attributes ?
image

Associations currently have a name attribute, however Camunda complains about it being an unknown attribute. The spec says that names aren't properties of associations. Should we remove 'name' from the attribute list in modeler for Associations?

As well, currently modeler doesn't expose the association direction, however the spec says it should exist. Should 'direction' be added?

Finally, is the expression property a valid attribute should it be removed or kept?

tuantruong625:
Fixes #181 

Association Attributes
- None
- One
- Both

**To Test** 
Create an association and connect to any element. Use the select in the inspector panel and choose values.

**Test Coverage**
Created a test for users to select all the different direction attributes


tests/e2e/specs/Modeler.spec.js
```diff
@@ -211,6 +211,45 @@ describe('Modeler', () => {
...
+    const textAnnotation = '#j_4';
+    const associationButton = '#v-25';
+    const associationNode = '#v-56';
+    const directionSelectSelector = '[name=\'associationDirection\']';
+
+    const testDirection = {
+      none: 'none',
+      one: 'one',
+      both: 'both',
+    };
```
chipit24: I think you can use `associationConfig` here as well?

src/mixins/crownConfig.js
```diff
@@ -3,6 +3,7 @@ import trashIcon from '@/assets/trash-alt-solid.svg';
 import pull from 'lodash/pull';
 
 export const highlightPadding = 3;
+import { direction } from '../components/nodes/association/associationConfig';
```
chipit24: I think it looks nicer if you group imports at the top, and you can use the `@` shorthand in the path.

Issue #183, opened 2019-01-31
Sequence Flows Go Through Elements


vehikl-morning-mob:
![image](https://user-images.githubusercontent.com/39924126/52088371-aa4db280-2579-11e9-9c87-8ff87695b440.png)

chipit24:
Quirks with sequence flows exist even in Camunda:
<img width="605" alt="screen shot 2019-02-01 at 10 18 03 am" src="https://user-images.githubusercontent.com/7561061/52142977-55b54080-2628-11e9-8582-031663356710.png">
There is a bit of tweaking that can be done with jointjs, but there will always be edge cases when sequence flows don't look quite right. E.g. I updated the router to help prevent overlapping, but it still falls apart when elements are very close:
<img width="244" alt="screen shot 2019-02-01 at 1 53 51 pm" src="https://user-images.githubusercontent.com/7561061/52143244-f60b6500-2628-11e9-98b6-6900f27454b2.png">
<img width="216" alt="screen shot 2019-02-01 at 1 54 07 pm" src="https://user-images.githubusercontent.com/7561061/52143249-f86dbf00-2628-11e9-9b10-3f0ba3e24bd4.png">

A complete solution may require setting up a custom router, anchors, and connectPoints, as well as precisely defining what the problem is (how flows should appear in all cases). This requires more discussion.

velkymx:
We're most concerned about the overlapping lines, especially from gateways. Make the flow confusing.

PR #199, opened 2019-02-06
Add Parallel Gateway

### LINKED ISSUE #195: Parallel Gateway
As a designer I would like a Parallel Gateway to represent two concurrent tasks in workflow.

https://www.lucidchart.com/pages/bpmn-gateway-types?a=1#section_2

https://docs.camunda.org/manual/7.8/reference/bpmn20/gateways/parallel-gateway/

tuantruong625:
Fixes #195 

Shape with appropriate attributes are included
- id
- name
- gatewayDirection 

Validation will be added as a separate feature.

- require a merge gateway when a parallel gateway is added

_As per @velkymx's comment in https://github.com/ProcessMaker/modeler/issues/195#issuecomment-461488974_


src/components/nodes/parallelGateway/index.js
```diff
@@ -48,6 +50,18 @@ export default {
...
         },
+        {
+          component: 'FormSelect',
+          config: {
+            label: 'Direction',
+            helper: 'Select direction of gateway',
+            name: 'gatewayDirection',
+            options: [
+              { value: `${ gatewayDirection.diverging }`, content: 'Diverging' },
+              { value: `${ gatewayDirection.converging }`, content: 'Converging' },
```
chipit24: Same here re: template strings.

src/components/nodes/gateway/gatewayConfig.js
```diff
@@ -0,0 +1 @@
+export const gatewayDirection = { diverging: 'diverging', converging: 'converging' };
```
chipit24: It seems you export almost the same object in `src/components/nodes/gateway/index.js`:
```js
export const gatewayDirectionOptions = { Diverging: 'Diverging', Converging: 'Converging' };
```

I think this config can be removed from one of these spots.

src/components/nodes/parallelGateway/parallelGateway.vue
```diff
@@ -4,93 +4,13 @@
...
-      ],
-    };
-  },
-  watch: {
-    'node.definition.name'(name) {
-      const { width } = this.shapeView.getBBox();
-      this.shape.attr('.label/text', joint.util.breakText(name, { width }));
-    },
-  },
+  extends: GatewayComponent,
```
chipit24: I think you can do the same refactor (extend) in `inclusiveGateway.vue` as well.

tuantruong625:
Do you mean how the sequence flow overlaps the task element ? If so this applies to the other gateways as well, and we should create a separate issue for it.

velkymx:
> Do you mean how the sequence flow overlaps the task element ? If so this applies to the other gateways as well, and we should create a separate issue for it.

The lines should not overlap. Do you think this looks correct?

velkymx:
Should look more like this:

screenshot 2019-02-12 14 28 52

chipit24:
@velkymx The issue with sequence flows overlapping other elements occurs not just with parallel gateways, but with all flow elements. Previously, we used the [manhattan router](https://resources.jointjs.com/docs/jointjs/v2.2/joint.html#routers.manhattan) in JointJS, which offered obstacle avoidance, but this created other issues.

We created an issue with sequence flows overlapping elements here: https://github.com/ProcessMaker/modeler/issues/183. I think it makes sense to tackle that issue in that ticket and create another PR for that, since it's not directly related to this PR.

Looking at your updated image, we could break the issue down into first modifying flows coming out of gateways. Would be interesting to set up, since there are still edge cases we need to address. Might be easier to let the user click and drag the flows.

PR #191, opened 2019-02-05
Add base tests for undo/redo functionality

chipit24:
Fixes #189.

**To run tests:** `npm test`.

**Important changes:**

This PR refactors some of the testing logic and adds two basic tests for undo/redo that can easily be added to:
 * Can undo and redo adding a task
 * Can undo and redo deleting a task

Helpful methods that can be used across tests have been moved to the `/tests/e2e/support/utils.js` file: `getGraphElements`, `getElementAtPosition`, and `dragFromSourceToDest`.

A new command has been added to `support/commands.js` for loading the modeler:
```js
beforeEach(() => {
  cy.loadModeler();
});
```

PR #208, opened 2019-02-08
Fix bug preventing saving node position to undo/redo stack

### LINKED ISSUE #207: Unable to undo after moving element
How to reproduce:
* Move the start event.
* The undo button is greyed out; moving the start event doesn't seem to push undo history to the undo history.
* Add a task.
* Move the task to a new position, and then move it to another new position.
* Press undo; the task is moved back to it's original position when added; the intermediate move wasn't added to the undo stack.

Expected behaviour:
* Users should be able to undo/redo position changes in all elements.

chipit24:
Fixes #207.

How to test:
* Visit the issue linked above; ensure you can't reproduce the bug.


chipit24:
Currently WIP pending merge of https://github.com/ProcessMaker/modeler/pull/191. Will add test cases for this after that PR is merged.

PR #209, opened 2019-02-08
Render Condition Expression on sequence flow 

### LINKED ISSUE #204: Show Expression on Flow
After a gateway I want to see the expression as part of the canvas label. 

image

tuantruong625:
Fixes #204 

Condition Expressions are rendered on sequence flow.

### DIFF
--- a/src/components/nodes/sequenceFlow/sequenceFlow.vue
+++ b/src/components/nodes/sequenceFlow/sequenceFlow.vue
@@ -9,6 +9,7 @@ import crownConfig from '@/mixins/crownConfig';
 import linkConfig from '@/mixins/linkConfig';
 import get from 'lodash/get';
 import { id as laneId } from '../poolLane';
+import { expressionPosition } from './sequenceFlowConfig';
 
 export default {
   props: ['graph', 'node', 'id', 'moddle', 'nodeRegistry'],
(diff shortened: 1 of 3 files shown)


src/components/nodes/sequenceFlow/sequenceFlow.vue
```diff
@@ -26,6 +27,20 @@ export default {
       return get(this.target, 'component.node.type');
     },
   },
+  watch: {
+    'node.definition.conditionExpression.body'(conditionExpression) {
+      this.shape.labels([{
```
chipit24: The label can be updated with the following function, this way it's position won't be reset when updating:
```js
this.shape.label(0, {
  attrs: {
    text: {
      text: conditionExpression,
    },
  },
});
```

src/components/nodes/sequenceFlow/sequenceFlowConfig.js
```diff
@@ -0,0 +1 @@
+export const expressionPosition = { distance: 0.25, offset: { x: -40, y: -20 } };
```
chipit24: I think this should be commented. The "Label Position" section of https://resources.jointjs.com/tutorial/link-labels describes how these numbers work. Or you can say something like "position label above and near the source of sequence flow".

chipit24:
> doesn't match the issue screen

To clarify, when created, these labels shouldn't overlap with the sequence flow, and should appear halfway between the source and target?

Also, it's currently possible to drag and move the labels (this is default JointJS behaviour); do you want to keep this behaviour?

velkymx:
The label should be on the left in the middle of the flow (note the screenshot in the description). I like that it can be moved, but it's not a requirement (leave it for now). I like the white behind the text, much cleaner than our competition

tuantruong625:
The positioning of the expression label will always remind centred of the sequence flow. 

When the user repositions any connected element the position of the label is inconsistent, as shown in the gif. Having the label position be consistent will require adding ports to the source element to identify which direction the sequence flow is connected too. 

This should be a separate issue and have some discussion on how the label position should be in any case.

conditionexpressposition

PR #232, opened 2019-02-19
Fix z-index with sequence flows

### LINKED ISSUE #231: Sequence flow layering issue with pools
**Steps to reproduce**
1. Drag any element
2. Connect Start Event to element to create a `sequence flow`
3. Drag a pool
 
sequence-flow

tuantruong625:
**Bug**
When dragging a pool onto the paper, the sequences flows are hidden behind the pool.

1. Drag any element
2. Connect Start Event to element to create a `sequence flow`
3. Drag a pool

sequence-flow

**Fix**

sequence-flow-fix

Fixes #231

### DIFF
--- a/src/mixins/crownConfig.js
+++ b/src/mixins/crownConfig.js
@@ -107,6 +107,7 @@ export default {
       }
 
       this.$emit('add-node', {
+        id: 'sequence-flow-button',
         type: 'processmaker-modeler-sequence-flow',
         definition: sequenceLink,
         diagram: this.moddle.create('bpmndi:BPMNEdge'),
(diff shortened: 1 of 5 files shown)


src/mixins/crownConfig.js
```diff
@@ -107,6 +107,7 @@ export default {
       }
 
       this.$emit('add-node', {
+        id: 'sequence-flow-button',
```
chipit24: This can be removed; `addNode` doesn't do anything with `id`, and the "add sequence flow" buttons for task, start event and gateway already have IDs set.

PR #247, opened 2019-02-21
Fix message flow undo/redo

### LINKED ISSUE #244: Message flow creates two "undo steps" instead of one
**To reproduce:**
1. Drag two pools onto the canvas.
2. Connected the two pools with a message flow.
3. Press undo, and the message flow disconnects from the pool and connects to the edge of the paper. The error `TypeError: Cannot read property 'id' of undefined` also appears in the console.

**Expected behaviour:**
In step 3 above, the message flow should be completely removed without error; the undo behaviour should be identical to that of sequence flows.

**Gif of problem:**
message_flow_pool_bug

chipit24:
Fixes #244.

**How to test:**
* Visit the issue linked above; ensure you can't reproduce the bug you that you experiecne the expected behaviour.


chipit24:
Bug fixed, just need to add tests.

PR #254, opened 2019-02-25
Fix startEvent props update after undo

### LINKED ISSUE #251: Unable to change start event properties after undo
**Steps to reproduce:**
1. Select the start event, and change both it's ID and its name.
2. Press `Undo`, and repeat step 1 above.
3. After undo, updates to the node inspector aren't persisted to the element.

**Expected behaviour:**
* You should be able to update all node properties before and after undo/redo.

**Gif of problem:**
undo_bug

chipit24:
Fixes #251.

**How to test:**
* Visit the issue linked above and ensure you cannot reproduce the bug, and that you experience the expected behaviour.

### DIFF
--- a/src/components/inspectors/InspectorPanel.vue
+++ b/src/components/inspectors/InspectorPanel.vue
@@ -106,7 +106,7 @@ export default {
     }, saveDebounce),
     defaultInspectorHandler(value, node, setNodeProp) {
       /* Go through each property and rebind it to our data */
-      for (const key in value) {
+      for (const key in omit(value, ['$type', 'eventDefinitions'])) {
         if (node.definition.get(key) !== value[key]) {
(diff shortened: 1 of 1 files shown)


src/components/inspectors/InspectorPanel.vue
```diff
@@ -106,7 +106,7 @@ export default {
     }, saveDebounce),
     defaultInspectorHandler(value, node, setNodeProp) {
       /* Go through each property and rebind it to our data */
-      for (const key in value) {
+      for (const key in omit(value, ['$type', 'eventDefinitions'])) {
```
caleeli: I think that Array properties (like eventDefinitions) represents child nodes and not attributes, may be this could be improved

PR #257, opened 2019-02-25
Fix bug with node inspector updates

### LINKED ISSUE #255: Most recent update to node props overrides previous update
When properties are changed on a node, a debounced `setNodeProp` function is called:
https://github.com/ProcessMaker/modeler/blob/116735a33607f33d0cbd9e32b3ac720b4d5cc91c/src/components/inspectors/InspectorPanel.vue#L103-L106

If multiple properties are changes within the debounce period (`saveDebounce`), then only the latest update will be applied. While the debounce period is reasonably short (currently 300 ms), this may not be enough time if setting properties programatically, or if the user quickly tabs through inputs.

The issue can be reproduced more easily by temporarily increasing the debounce timeout.

prop_change

chipit24:
Fixes #255.

Visit the issue linked above and ensure you cannot reproduce the issue:
* Click on the start event, and change both its ID and name within 300 ms of each other (type into ID, hit tab and type name right away). Wait 300 ms and ensure neither of the two inputs get reset.

### DIFF
--- a/src/components/inspectors/InspectorPanel.vue
+++ b/src/components/inspectors/InspectorPanel.vue
@@ -117,10 +117,23 @@ export default {
     processNodeInspectorHandler(value) {
       return this.defaultInspectorHandler(omit(value, ['artifacts', 'flowElements', 'laneSets']));
     },
-    setNodeProp: debounce(function(node, key, value) {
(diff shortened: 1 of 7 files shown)


chipit24:
WIP: Will be adding tests.

Issue #263, opened 2019-02-26
Sequence Flow Overlapping with Gateways.


tuantruong625:
We're most concerned about the overlapping lines, especially from gateways. Make the flow confusing.

_Originally posted by @velkymx in https://github.com/ProcessMaker/modeler/issues/183#issuecomment-467498270_

chipit24:
PR merged.

PR #264, opened 2019-02-26
Update sequence flow behaviour with gateways

tuantruong625:
Reference #263 

**Old Behaviour** 

sequence-flow-old

**New Behaviour** 

sequence-flow-new


velkymx:
Looks great.

PR #266, opened 2019-02-27
Inclusive Gateways

### LINKED ISSUE #262: Inclusive Gateways
https://docs.camunda.org/manual/7.6/reference/bpmn20/gateways/inclusive-gateway/

tuantruong625:
**Requirements.**

- [x] Capability to add inclusive gateways to divide or fork the flow

- [x] Capability to add inclusive gateways to merge current flows on a process

- [x] Capability to define the direction of execution for the gateway

- [x] Capability to detect automatically the direction for the gateway based on the connected sequence flows

**Error handling is in development.

Functionality wise the two scenarios in the PRD are covered, the only issue is with the sequence flows and how they are rendered. Which is addressed in https://github.com/ProcessMaker/modeler/pull/264.

Product Requirements Document : [PM4-InclusiveGateway-PRD-270219-1808.pdf](https://github.com/ProcessMaker/modeler/files/2915597/PM4-InclusiveGateway-PRD-270219-1808.pdf)

Fixes #262


danloa:
When using the connector as a  converging one, the flow begins from the behind side of the icon and it's almost impossible to get a straight line, even moving the  element that is next to the connector at different distances.

converging2
converging

tuantruong625:
@danloa I pushed up some changes on how sequence flows behave, but functionally wise inclusive gateways meet the requires in this PR. In your screenshot above, the sequence flow renders the same way with any gateway, and I believe should be addressed in another ticket and not this one.

src/components/nodes/sequenceFlow/sequenceFlow.vue
```diff
@@ -10,6 +9,7 @@ import linkConfig from '@/mixins/linkConfig';
 import get from 'lodash/get';
 import { id as laneId } from '../poolLane';
 import { expressionPosition } from './sequenceFlowConfig';
+import  inclusiveGateway from '../inclusiveGateway/index';
```
chipit24: For the other components, the ID is specified as a non-default export. I think this should be changed to that as well, to keep it consistent:

```
import { id as inclusiveGatewayId } from '../inclusiveGateway/index';
```

This also lets users know, from the imports, that only the ID is required and used, and not the entire config.

src/components/nodes/sequenceFlow/sequenceFlow.vue
```diff
@@ -92,7 +108,7 @@ export default {
     },
     isSourceElementGateway() {
       const sourceShape = this.shape.getSourceElement();
-      return ['bpmn:ExclusiveGateway', 'bpmn:parellelGateway'].includes(sourceShape.component.node.definition.$type);
+      return ['bpmn:ExclusiveGateway', 'bpmn:parellelGateway', 'bpmn:InclusiveGateway'].includes(sourceShape.component.node.definition.$type);
     },
     createLabel() {
       if (!this.node.definition.conditionExpression) {
```
chipit24: This file includes a couple statement of the form:
```js
typeof thing === 'undefined'
```
Since you're touching this file, I think it's worth cleaning up those lines as well:
```js
thing == null
// or
!thing
```

src/components/nodes/sequenceFlow/sequenceFlow.vue
```diff
@@ -66,7 +67,22 @@ export default {
         this.targetIsNotALane() &&
         this.targetIsInSamePool() &&
         this.targetIsNotSource() &&
-        this.validateOutgoing();
+        this.validateOutgoing() &&
+        this.targetInclusiveGateway();
+    },
+    targetInclusiveGateway() {
```
chipit24: I'm not sure this is the best place for this function. This function is called from `isValidTarget`, but it doesn't actually validate anything, it modifies the target if it's an inclusive gateway.

If possible, I think this logic should be moved to the actual inclusive gateway component.

tests/e2e/specs/Modeler.spec.js
```diff
@@ -55,7 +55,7 @@ describe('Modeler', () => {
 
     const endEventPosition = { x: 100, y: 500 };
     dragFromSourceToDest(
-      nodeTypes.evenEvent,
+      nodeTypes.endEvent,
```
chipit24: Good catch 🥇

tests/e2e/specs/ExclusiveGateway.spec.js
```diff
@@ -21,10 +22,10 @@ describe('Exclusive Gateway', () => {
...
-    getElementAtPosition(exclusiveGatewayPosition)
-      .then(() => {
-        typeIntoTextInput('[name=\'name\']', testString);
-        cy.get('[name=\'name\']').should('have.value', testString);
-      });
+    waitToRenderAllShapes();
+    getElementAtPosition(exclusiveGatewayPosition).click();
+
+    typeIntoTextInput('[name=\'name\']', testString);
+    cy.get('[name=\'name\']').should('have.value', testString);
```
chipit24: Escaping single quotes here isn't necessary:
```
'[name=name]'
```

tests/e2e/specs/InclusiveGateway.spec.js
```diff
@@ -0,0 +1,70 @@
...
+  dragFromSourceToDest,
+  typeIntoTextInput,
+  connectNodesWithFlow,
+  waitToRenderAllShapes,
+  getElementAtPosition,
+} from '../support/utils';
+
+import { nodeTypes } from '../support/constants';
+
+describe.only('Inclusive Gateway', () => {
```
chipit24: Forgot the `.only` here.

PR #270, opened 2019-02-28
Add base call activity component

### LINKED ISSUE #260: Subprocess Marker
screenshot 2019-02-26 08 22 59

https://docs.camunda.org/manual/7.6/reference/bpmn20/subprocesses/call-activity/

Inspector should allow for selected another process that is active.

chipit24:
Fixes #260.

Screenshots of feature:

screen shot 2019-02-28 at 3 40 27 pm

screen shot 2019-02-28 at 3 42 41 pm

screen shot 2019-02-28 at 3 34 35 pm


caleeli:
Oh please review the `isExecutable` property in the second process.

When I save the XML I could see that the second process (subprocess) the property  `isExecutable` 
```
...
<bpmn:process id="process_2" isExecutable="false">

```
It should be set to `true` since it will be executed by the **callActivity**

Non executable processes are used only for documenting purposes

chipit24:
@caleeli Should it be possible for the user to modify `isExecutable`, or should this property be set to `true` when targeted by a call activity and set to `false` otherwise?

chipit24:
Regarding my last question, I've decided to leave the `isExecutable` as the default value (`true`).

chipit24:
Blocked, requires clarification:

* Need to know how the "Declare as global" checkbox on Tasks and Script Tasks should behave. Should an additional property be added to the BPMN definitions for these elements, or should toggling the checkbox convert the elements to `bpmn:GlobalTask` and `bpmn:GlobalScriptTask`?
* The "Call activity type" dropdown has two options: `Process` and `Global Task`. Should changing this option set a property on the BPMN definition, or should this value be purely visual (deduced from the `calledElement` type of the Call Activity).
  - Additional suggestion: Since there are only two values, radio buttons may provide better UX.
* Related to the above two points, should an additional property be added to the BPMN definition for Call Activity to store the `Run process asynchronously` boolean value?

@velkymx

PR #271, opened 2019-03-01
Add base bpmnlint plugin and Inclusive Gateway validation

### LINKED ISSUE #268: Inclusive Gateway - Converging/Diverging Constrains are not being validated
The gateway direction (Converging | Diverging) adds constraints on how the Gateway MAY be used.

• Converging: This Gateway MAY have multiple incoming Sequence Flows but MUST have no more than one (1) outgoing Sequence Flow.
• Diverging: This Gateway MAY have multiple outgoing Sequence Flows but MUST have no more than one (1) incoming Sequence Flow.

At this moments that is not being validated.

https://processmaker.slack.com/files/U078GPWPQ/FGL4K5X1B/convergingconstraintnotvalidated.mov

chipit24:
Fixes #268.

The PR adds custom BPMNLint validation. It adds validation rules for the following gateway constrains:
* A **Gateway** with a `gatewayDirection` of `converging` MUST have multiple _incoming_ **Sequence Flows**, but MUST NOT have multiple _outgoing_ **Sequence Flows**.
* A **Gateway** with a `gatewayDirection` of `diverging` MUST have multiple _outgoing_ **Sequence Flows**, but MUST NOT have multiple _incoming_ **Sequence Flows**.

**To test:**
Drag a gateway onto the grid, and check that the above errors correctly appear when validating the diagram.


package.json
```diff
@@ -28,6 +28,7 @@
     "bootstrap-vue": "^2.0.0-rc.11",
     "bpmn-moddle": "^5.1.5",
     "bpmnlint": "^5.0.0",
+    "bpmnlint-plugin-processmaker": "file:processmaker-plugin",
```
velkymx: Where is this package being hosted?

chipit24: Locally; The code is included in this PR, it refers to the `processmaker-plugin` folder.

chipit24: This is how they have it set up in their example:
https://github.com/bpmn-io/bpmnlint-playground/blob/f35cc2df5ad74ddde4ee8e7cef319dbc2af1bf77/package.json#L22

PR #286, opened 2019-03-12
Fix overlapping sequence flows

### LINKED ISSUE #283: Lane overlaps sequence flow
**Steps to reproduce:**
* Drag Task onto grid and connect start event to it with a sequence flow.
* Drag a Pool onto the grid, and add a lane to it (top or bottom, doesn't matter).
* The sequence flow disappears.

**Expected behaviour:**
* Sequence flows should always be visible.

**Gif of issue:**
overlapping_bug

chipit24:
Fixes #283.

**How to test:**
* Drag a task onto the grid and connect the start event to it with a sequence flow.
* Drag a pool onto the grid. Ensure the sequence flow is still visible.
* Add a lane to the pool. Ensure the sequence flow is still visible.

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -54,6 +54,7 @@
       @unsetPools="unsetPools"
       @setPools="setPools"
       @save-state="pushToUndoStack"
+      @set-shape-stacking="setShapeStacking"
     />
   </div>
 </template>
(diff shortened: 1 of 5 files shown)


src/components/Modeler.vue
```diff
@@ -54,6 +54,7 @@
       @unsetPools="unsetPools"
       @setPools="setPools"
```
tuantruong625: Not related to this PR, but the naming for our events are inconsistent. The Vue documentation recommends using kebab-case.

velkymx: I've also noticed this...

chipit24: For reference, it's mentioned the docs here: https://vuejs.org/v2/guide/components-custom-events.html. In our case, it definitely doesn't break anything and can easily be missed, but consistency is something I always advocate for.

Ideally, there's an ESLint rule to enforce this kind of stuff, but I didn't find one :( might look into creating one later.

chipit24: The above references have been fixed in this PR.

PR #298, opened 2019-03-20
Fix broken tests on develop

### LINKED ISSUE #296: Fix Broken Test on Intermediate Message Catch Event
Broken Test

- `Intermediate Message Catch Event Update properties`

tuantruong625:
Fix broken test on develop
- **IntermediateMessageCatch.spec.js**
    - Update Properties

- **Modeler.spec.js**
    - Generates sequential, unique node IDs

Fixes #296


chipit24:
@tuantruong625 The `Generates sequential, unique node IDs` test in `Modeler.spec.js` is failing for me locally both with the `npm test` and `npm run test-ci` commands. I updated Cypress from v3.1.5 to v3.2.0 but this made no difference. Note that I'm using the latest version of Chrome, v73; this might be the difference if tests are passing for you.

PR #303, opened 2019-03-22
Add vue-i18n

### LINKED ISSUE #301: Translation
We currently are translating the php / js version of bpm and we want the modeler to support this. Text strings needs to be added to a translation file that we can import into bpm to translation.

chipit24:
Fixes #301.

This PR adds internationalization using https://kazupon.github.io/vue-i18n/.


chipit24:
Closed in favour of https://github.com/ProcessMaker/modeler/pull/312.

PR #308, opened 2019-04-02
Load call activity process list from API

### LINKED ISSUE #305: Call Activity - list of processes is not displaying the correct titles
When configuring a subprocess (callActivity) the list of processes is not showing the correct texts. In this example I have two pools (named admin and create process) but When configuring the call activity instead of those titles, the default text (in this case Process_1 and Process_2) is displayed.

image

chipit24:
Fixes #305.

**TODO**:
 * The generated XML for a _Call Activity_ seems valid, but results in an error, `The bpm definition is not valid`, when trying to save in BPM. See discussion in issue: https://github.com/ProcessMaker/modeler/issues/305#issuecomment-479166889.

**How to test:**
* Add a _Call Activity_ to the grid.
* Select the _Call Activity_ and the _Processes_ dropdown in the inspector panel should be populated with all processes from the API that have a start event.

Screen Shot 2019-04-02 at 3 12 44 PM
Screen Shot 2019-04-02 at 3 21 15 PM


chipit24:
I've updated this PR so that a sequence flow connected to a call activity will now show a dropdown with the list of start events on the call activity's selected process (`calledElement`).

I'm currently saving this event reference under the `calledElementStartEvent` attribute on the sequence flow, but this is, of course, not spec compliant. @velkymx @caleeli how should we handle this interaction?

Screen Shot 2019-04-05 at 12 18 43 PM

caleeli:
@chipit24  @velkymx I put a comment on the PRD related to that, and how we should store the properties of the call activity:

https://processmaker.atlassian.net/wiki/spaces/PM4/pages/479135669/Call+Activities+Process+Support

```
<callActivity calledElement="globalDBProcessId:localProcessId" pm:startEvent="node_5" />

```

chipit24:
@caleeli So, the Call Activity will have two dropdowns now, right? One for selecting the process and another for selecting the start event? The sequence flow connecting to the start event will then just be a regular sequence flow?

I don't have permission to view the link you posted.

caleeli:
@ chipit24 Yes the sequence flow should be a regular one.
About the CallActivity, I think could be good to have one drop-down to select the process (from the database globalDBProcessId) and a second one to select the inner startEvent (pm:startEvent) and the bpmn process it belongs (localProcessId)

the calledElement should refer to a BPMN process ID
and pm:startEvent should refer to a startEvent ID (from the external process)

chipit24:
The selected start event will remain on the sequence flow connected to the call activity. I'll keep this PR in Blocked until implementation details have been confirmed.

PR #312, opened 2019-04-03
Add i18next integration

chipit24:


nolanpro:
@chipit24 FYI Marco wrapped additional translations in https://github.com/ProcessMaker/modeler/compare/feature/309

chipit24:
Moving this PR to review since it looks like @marcoAntonioNina has mostly done the same work in the [`feature/309`](https://github.com/ProcessMaker/modeler/compare/feature/309) branch.

This PR can be reviewed / merged.

I looked into pulling in translations into the component config files, but it would require reaching into the global Vue instance or exposing the i18n object from bpm. @nolanpro I think your original idea is the way to go: handle the translations for the inspector config inside the vue form renderer.

nolanpro:
@chipit24 could you merge feature/309 into this branch? Thanks

chipit24:
I've merged the branch.

@nolanpro What are your thoughts on removing `searchLabels` / `translatedLabels` from `modeler` and the `inspection` watcher from `vue-form-builder`, and having the translation setup in `vue-form-elements`?

The form elements are where the values are actually rendered, and it's mostly label and placeholder text that needs to be translated. I think it's much more straightforward to place calls to `$t` in those templates than to introduce additional logic in the layers before that.

nolanpro:
@chipit24 yeah that would probably be the right way to do it, but since we need to wrap this up today it might have to wait for a v2/refactor

chipit24:
I've updated the inspector config translation. It now translates the default process node, and it does it on `registerNode`, so it doesn't have to keep track of what's been translated since it just does a single pass on initial load.

I think having it like that will make it easier to refactor later too, if we move the translation into `vue-form-elements`.

PR #359, opened 2019-04-30
Modeler design

### LINKED ISSUE #358: Modeler New Design
Model the design similar to the builder and script editor.
- use vue-bootstrap cards for tools and inspector

tuantruong625:
Fixes #358, fixeds #362.


src/components/controls.vue
```diff
@@ -98,68 +101,49 @@ export default {
 </script>
 
 <style lang="scss">
+.card-header {
```
chipit24: The `card-header` class is used by bootstrap and automatically applied to card header elements. I think it might be worth renaming this class to avoid conflicts and confusion with that bootstrap class. Maybe use `white-card-header` or `bg-white`?

chipit24:
For the updates in this PR to work as expected, https://github.com/ProcessMaker/vue-form-elements/pull/36 needs to be merged and published, and then updated in `spark-modeler` to point to those new updates.

For the new design to work well in `spark`, https://github.com/ProcessMaker/spark/pull/1804 needs to be merged.

Some notes:
* The columns might be different sizes when viewed in `spark`. This is because the bootstrap grid layout / columns are fluid, and will expand and shrink based on the available width. To get around this, we'd need to set a max-height or use custom CSS to hard-code the widths. It works and looks fine this way, but let me know if you want this changed.
* (Related to the above) The padding to the left of the inspector is slightly less than the padding to the right of the control panel. This is to give the inspector more space and also to match the changes made to the screen builder. Padding can easily be adjusted, but the point above must be considered.

tuantruong625:
@velkymx 
What do you think about reducing the font-size for the control labels and adding tool-tips?

Screen Shot 2019-05-10 at 12 57 41 PM

PR #361, opened 2019-05-06
Add missing element warning

### LINKED ISSUE #357: Unsupported Elements
If a BPMN file contains an unsupported element, we'd like throw an error that the BPMN contains unsupported elements.

chipit24:
Fixes #357.

This PR add a warning for unknown elements:
Screen Shot 2019-05-06 at 5 02 43 PM

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -1,14 +1,20 @@
 <template>
   <div class="modeler">
-    <div class="alert-container">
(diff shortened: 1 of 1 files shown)


velkymx:
@chipit24 @ssshake this is ok for now, but we should discuss changes like this before implementation. For example, we show alerts at the top of the page, not the bottom - and it should show in the linter.

Issue #398, opened 2019-05-22
Element Labels not Translating


velkymx:
Default name is not translated when dragged on to the canvas.

![image](https://user-images.githubusercontent.com/535300/58179525-384ec580-7c5d-11e9-85c3-76b038eea80c.png)

PR #408, opened 2019-05-30
Missing translations

### LINKED ISSUE #385: Missing translations - modeler
These need to be wrapped in a translator `{{ $t(...) }}` or `this.$t(...)` and/or added to the en.json in Spark and/or sent to smartling and es.json/fr.json updated in the translations package

- [x] controls.vue - Controls title
- [x] inspectorPanel.vue - Inspector title
- [x] Controls - hover value
- [x] Controls - default name when dragged on to the grid

tuantruong625:
**Missing Translations**

- Control Title
- Inspector Title
- Controls Hover (tooltip)
- Default names of elements

Fixes #385, #398.

### DIFF
--- a/src/components/controls.vue
+++ b/src/components/controls.vue
@@ -23,7 +23,7 @@
         @dragstart="$event.preventDefault()"
         @mousedown="startDrag($event, control.type)"
       >
-        <div class="tool text-truncate ml-1" v-b-tooltip.hover :title="`${$t(control.label)}`">
+        <div class="tool text-truncate ml-1" v-b-tooltip.hover :title="$t(control.label)">
           <img :src="control.icon" class="tool-icon mr-1">
           {{ $t(control.label) }}
         </div>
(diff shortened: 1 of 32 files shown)


src/components/nodes/callActivity/callActivity.vue
```diff
@@ -67,6 +67,7 @@ export default {
   },
   mounted() {
     this.shape = new TaskShape();
+    const defaultName = this.node.definition.get('name');
```
chipit24: I think of "default" as a fall-back value to use when the user supplied or expected value doesn't exist. But if this process is being loaded in spark, then what gets returned from the definition may not be a default value, but something the user initially supplied.
 
If you're aiming to increase readability, I think something like `callActivityName` may be better here.

src/components/controls.vue
```diff
@@ -23,7 +23,7 @@
         @dragstart="$event.preventDefault()"
         @mousedown="startDrag($event, control.type)"
       >
-        <div class="tool text-truncate ml-1" v-b-tooltip.hover :title="control.label">
+        <div class="tool text-truncate ml-1" v-b-tooltip.hover :title="`${$t(control.label)}`">
```
chipit24: You don't need the template string (backticks) here. This can be:
```
:title="$t(control.label)"
```

PR #429, opened 2019-06-12
Unqiue id message event

### LINKED ISSUE #428: messageEventDefinition id invalid
Throws an error when attempting to save with two Intermediate Message Catch Events without changing the default message event ids

**How to reproduce**
- Create a simple process with two Intermediate Message Catch Events
- Attempt to save

**Expected**
- Bpm is valid without any errors

**Cause**
- The default message event id are not unique on creation

tuantruong625:
Message Event IDs are unique on render. Initially, they were the same on render and if the user did not edit it `messageEvent Id` it would throw an error in Spark.

**Spark Video Demo**
https://drive.google.com/open?id=1MAW33gkScD3xrPbEYV2YHn5LVeFd6ZvP

Fixes #428

### DIFF
--- a/src/components/nodes/intermediateMessageCatchEvent/index.js
+++ b/src/components/nodes/intermediateMessageCatchEvent/index.js
@@ -1,6 +1,9 @@
 import component from './intermediateMessageCatchEvent.vue';
 import omit from 'lodash/omit';
 import { configId } from '@/components/inspectors/configId';
+import NodeIdGenerator from '../../../NodeIdGenerator';
+
+const generateMessageEventId = new NodeIdGenerator();
 
 export default {
   id: 'processmaker-modeler-intermediate-message-catch-event',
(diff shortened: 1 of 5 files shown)


src/components/nodes/intermediateMessageCatchEvent/index.js
```diff
@@ -1,6 +1,9 @@
 import component from './intermediateMessageCatchEvent.vue';
 import omit from 'lodash/omit';
 import { configId } from '@/components/inspectors/configId';
+import NodeIdGenerator from '../../../NodeIdGenerator';
+
+const generateMessageEventId = new NodeIdGenerator();
```
chipit24: `generateMessageEventId` sounds like the name of a method. To keep this consistent with how it's named in `Modeler.vue`, I think a name of `messageEventIdGenerator` would be better. Also, since it seems the same instance will always use the same prefix, I think it would be better to pass the prefix into the constructor as well (maybe have the class handle adding the `_` separator as well):

```js
const messageEventIdGenerator = new NodeIdGenerator('message_event');
...
id: messageEventIdGenerator.generateUniqueNodeId(),
```

You should also call `generateUniqueNodeId` and **not** `generateNewNodeId`, to ensure IDs are unique. Because right now, if I drag an Intermediate Message Catch Event and manually change it's Message Event Identifier to `message_event_2`, the next Intermediate Message Event I add will be given the same ID.

PR #434, opened 2019-06-15
Add tooltip crown items

### LINKED ISSUE #395: Add tooltips to crown items

tuantruong625:
**Spark Video Demo**
https://drive.google.com/open?id=1gguXI5g5t-eLatuybQQErl9ph1ag3iTA

Fixes #395

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -1,5 +1,12 @@
 <template>
   <b-row class="modeler h-100">
+    <b-tooltip
+      v-if="tooltipTarget"
+      :key="tooltipTarget.id"
+      :target="getTooltipTarget"
+      :title="tooltipTitle"
+    />
+
     <b-col class="h-100 overflow-hidden controls-column" :class="{ 'ignore-pointer': canvasDragPosition }">
       <controls
         :controls="controls"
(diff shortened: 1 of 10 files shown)


src/components/Modeler.vue
```diff
@@ -180,6 +190,11 @@ export default {
     autoValidate() { this.validateIfAutoValidateIsOn(); },
   },
   computed: {
+    tooltipTitle() {
+      if (this.tooltipTarget) {
+        return this.tooltipTarget.$el.data('title');
+      }
```
ssshake: @tuantruong625  @chipit24  the only thing bugging me about this is what will happen if there is no title? Will it show a tool tip with the text "undefined". 

I'm thinking of a scenario where a new component gets added which doesn't have the title set in crown config. Can there be a safe default, like if no title dont show anything?

tuantruong625: `Crown Button` as a fall back?

Screen Shot 2019-06-19 at 4 50 31 PM

tuantruong625: @ssshake @velkymx Will create a separate issue for how to handle if title is undefined.

PR #436, opened 2019-06-17
Style validation drawer

### LINKED ISSUE #417: Style validation drawer to be consistent with design language.
Screen Shot 2019-06-04 at 10 21 05 AM

@velkymx Will require some discussion.

tuantruong625:
**Previous Implementation**
Screen Shot 2019-06-04 at 10 21 05 AM

**Updated**
Screen Shot 2019-06-17 at 1 56 16 PM

**Spark video demo**
https://drive.google.com/open?id=1zJvyo8QQhQy8lXxYCxT2FEHwmIze5xad

Fixes #417

### DIFF
--- a/src/components/ValidationStatus.vue
+++ b/src/components/ValidationStatus.vue
@@ -128,15 +128,15 @@ export default {
 </script>
 
 <style scoped lang="scss">
-$primary-white: #F0F3F7;
+$primary-white: #f7f7f7;
 $seconadry-grey: #555555;
 $secondary-blue: #3397e1;
-$border-color: #aaaaaa;
+$border-color: rgba(0, 0, 0, 0.125);
 $id-container-width: 6.5rem;
 $message-container-width: 18rem;
 $error-category-width: 1rem;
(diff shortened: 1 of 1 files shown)


src/components/ValidationStatus.vue
```diff
@@ -174,8 +174,10 @@ $button-color: #3BD7FF;
   width: $validation-container-width;
   background-color: $primary-white;
   overflow: auto;
-  margin-bottom: 3rem;
+  margin-bottom: 4.30rem;
   border: 1px solid $border-color;
+  margin-right: 1.20rem;
```
chipit24: Nitpick: If you have trailing 0s, it should be consistent, but it seems we've stuck with not having trailing 0s so I think the above values can be changed to `4.3rem` and `1.2rem`.

velkymx:
@tuantruong625 the drawer shouldn't be floating. It should slide out as though part of the footer.

tuantruong625:
@velkymx 
Screen Shot 2019-06-18 at 11 28 20 AM

PR #445, opened 2019-06-25
Crown Buttons: Hide tooltip if title is undefined

### LINKED ISSUE #442: Title undefined for crown button tooltips
If a title is not provided and returns undefined for the crown button tooltips, how should the tooltip behave?

**Current implementation**
- Default tooltip title is renderer as `Crown Button`

Possible Idea
- Do not display tooltip at all if there is no title provided

tuantruong625:
If crown button titles are undefined; do not show a tooltip.

**Spark Video Demo**
https://drive.google.com/open?id=1-ScaCjrG68R7N-nM4Qc2o3Be9o8k1sRK

**Start Event, sequence flow button's title was set to undefined for testing purposes.

Fixes #442

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -4,7 +4,7 @@
       v-if="tooltipTarget"
       :key="tooltipTarget.id"
       :target="getTooltipTarget"
-      :title="tooltipTitle || ''"
+      :title="tooltipTitle"
     />
 
     <b-col class="h-100 overflow-hidden controls-column" :class="{ 'ignore-pointer': canvasDragPosition }">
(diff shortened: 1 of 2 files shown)


src/components/Modeler.vue
```diff
@@ -4,7 +4,7 @@
       v-if="tooltipTarget"
       :key="tooltipTarget.id"
       :target="getTooltipTarget"
-      :title="tooltipTitle"
+      :title="tooltipTitle || ''"
```
chipit24: The tooltip shouldn't render if an empty string is passed to it, so this can ca left as:
```
 :title="tooltipTitle"
```
And `hideTooltip ` can be removed.

PR #475, opened 2019-07-12
Vertex Hover State

### LINKED ISSUE #474: Show delete button on vertex hover
Currently, deleting a vertex on a sequence flow requires double-clicking on the vertex; this action is not obvious to new users. Replace the vertex "dot" with a larger red delete button that will remove the vertex when clicked.
___
This could be accomplished using the [`Button`](https://resources.jointjs.com/docs/jointjs/v2.2/joint.html#linkTools.Button) link tool:
```js
const buttonTool = new joint.linkTools.Button();
```

tuantruong625:
**Spark Video Demo**
https://drive.google.com/open?id=13OADIhx0TUpWXtrEuN4E-drQeunIS-KS

Fixes #474


src/components/Modeler.vue
```diff
@@ -947,5 +947,19 @@ $cursors: default, not-allowed;
       }
     }
   }
+
+  .vertex > circle:hover {
+    fill:#ED4757;
+    cursor: url('../assets/delete-icon-vertex.png') 0 0, pointer;
+  }
+
+  .link-tool-helper {
```
chipit24: This class doesn't appear to be used anymore

src/components/Modeler.vue
```diff
@@ -947,5 +947,19 @@ $cursors: default, not-allowed;
       }
     }
   }
+
+  .vertex > circle:hover {
```
chipit24: If you replace this selector with only `.joint-marker-vertex { ...` then it still works as expected, and you could remove the extra `vertices` code from `linkConfig.js`.

PR #477, opened 2019-07-15
Allow modifying flow endpoints

### LINKED ISSUE #473: Allow modifying connection point of sequence flows
Allow dragging the connection point (source / target vertices) of a sequence flow; allowing users to choose where the connection enters or leaves the node.
___
This can be accomplished using the `SourceAnchor` and `TargetAnchor` link tools:
```js
const sourceAnchorTool = new joint.linkTools.SourceAnchor();
const targetAnchorTool = new joint.linkTools.TargetAnchor();
```

chipit24:
Fixes #473.

Video of fix: https://www.dropbox.com/s/satsv7p2w6oz389/anchor_ports_working.mov?dl=0.

### DIFF
--- a/src/components/nodes/callActivity/callActivity.vue
+++ b/src/components/nodes/callActivity/callActivity.vue
@@ -6,6 +6,7 @@
 import joint from 'jointjs';
 import connectIcon from '@/assets/connect-elements.svg';
 import crownConfig from '@/mixins/crownConfig';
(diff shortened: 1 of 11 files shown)


chipit24:
> Is there a reason why the other nodes( intermediate event, pool, lane) are missing the port mixin?

That was missed–added now!

PR #508, opened 2019-08-12
Fix start timer event timer expression saving

### LINKED ISSUE #506: Start Timer Event timing control data does not persist
In the inspector panel of the start timer event, the property **Repeat on** selector does not persist its selection when saving.

Screen Shot 2019-08-12 at 11 16 47 AM

**How to reproduce bug**
- drag start timer event
- select a day for **Repeat on**
- Click save

**Current Behaviour**
- Users selected date does not persist

**Expected Behaviours** 
- Users selected date loads

chipit24:
Fixes #506.

### DIFF
--- a/src/components/inspectors/TimerExpression.vue
+++ b/src/components/inspectors/TimerExpression.vue
@@ -60,8 +60,8 @@
             control-class="form-control"
             :format="DateTime.DATE_SHORT"
             phrases='{ok: $t("Save"), cancel: $t("Cancel")}'
-            :vaue="endDate"
-            @input="endDate = $event"
+            :value="endDate"
+            @input="setEndDate"
           />
         </b-form-group>
 
@@ -167,17 +167,15 @@ export default {
       return 'You must select at least one day.';
     },
     iso8606Expression() {
-      const expression = [];
-
-      if (this.hasMultipleWeekdaySelected()) {
-        expression.push(this.startDate);
-
-        this.selectedWeekdays.forEach(day => {
-          expression.push(this.getCycle(this.getWeekDayDate(this.startDate, day)));
-        });
-      } else {
-        expression.push(this.getCycle(this.startDate));
+      if (this.selectedWeekdays.length === 1 && this.sameDay) {
+        return this.getCycle(DateTime.fromISO(this.startDate));
       }
+
+      const expression = [];
+      expression.push(this.startDate);
+      this.selectedWeekdays.forEach(day => {
+        expression.push(this.getCycle(this.getWeekDayDate(day)));
+      });
 
       return expression.join('|');
     },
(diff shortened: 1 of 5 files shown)


dmamills:
[note to check out this code during next mob](https://github.com/ProcessMaker/modeler/blob/develop/src/components/inspectors/InspectorPanel.vue#L127)

src/components/inspectors/TimerExpression.vue
```diff
@@ -295,9 +307,9 @@ export default {
         (this.ends === 'ondate' ? this.endDate : '')
       );
     },
-    getWeekDayDate(date, isoWeekDay) {
+    getWeekDayDate(startDate, isoWeekDay) {
```
chipit24: Problem seems to be with this function; it appears it's adding a day to each selected weekday.

chipit24:
@caleeli We're unclear what the correct timer expression string should be in certain cases when repeat is set to weekly.

When only one day is selected, and that day is the current day (2019-08-14), the expression would be:
```
R/2019-08-14T00:00:00.000-04:00/P1W
```
But when another day is selected, for example, the next day, the expression becomes:
```
2019-08-14T00:00:00.000-04:00|R/2019-08-14T00:00:00.000-04:00/P1W|R/2019-08-15T00:00:00.000-04:00/P1W
```
Note that a start date is appended to the expression now.

Which raises a few questions:

1. Should the start date always be omitted when only one day of the week is selected? Always appending the start date would simplify the code.
2. When multiple days of the week are selected, do the additional dates have to be after the start date?
3. Is the start time ignored? I.e. does it always repeat at the start of the day?

caleeli:
ISO_8601 Defines a <Repeating interval> as: R[n]/<interval>
Where an interval has the form: <start_datetime>/<duration>/<end_datetime>

Since a <Repeating Interval> can not represent a weekly cycle with multiple days, in the core we define a timer expression with multiple Repetitions Intervals, with the following structure:

<start_datetime>|<repeating_interval_1>|<repeating_interval_2>|...<repeating_interval_n>

image

**<start_datetime>** Stores the start datetime from when the group of repeating intervals start. In the image it is 2019-08-13T21:00:00Z

**<repeating_interval_n>** Represents a repetition interval. For example every monday from date A to date B. This way we could use one <repeating_interval> for each day selected.

In the example:
repeating_interval_1 (for Mondays) = R/2019-08-13T21:00:00Z/P1W/2019-08-31T21:00:00Z
repeating_interval_2 (for Wednesdays) = R/2019-08-15T21:00:00Z/P1W/2019-08-31T21:00:00Z
repeating_interval_3 (for Fridays) = R/2019-08-17T21:00:00Z/P1W/2019-08-31T21:00:00Z

So:

1. Yes it could work, appending always the start date even when only one day of the week is selected. The reason why the start date is removed is to keep the standard (ISO8601) because a week cycle of one day of is the same as one <Repeating interval>. 

2. Yes. For multiple days selected, each <Repeating interval> should start the first date of the selected day after or equal to the start date. In the example:

in repeating_interval_1 = 2019-08-13T21:00:00Z is the first Monday of the cycle after or equal the start date
in repeating_interval_2 = 2019-08-15T21:00:00Z is the first Wednesday of the cycle after or equal the start date
in repeating_interval_3 = 2019-08-17T21:00:00Z is the first Friday of the cycle after or equal the start date

3. The time is stored in the <start_datetime> and for each <repeating interval> (start date time and end date if selected)

chipit24:
I ran the exact same tests locally (`npm run test-ci` on node 10.16.0 and Chrome 76) and all the tests pass. The single test that was added in this PR is the only test that is failing on CI. It's failing because it expects the timer expression:
```
R/2019-08-14T05:30:00.000Z/P3W/2019-08-22T05:30:00.000Z
```
To be equal to:
```
R/2019-08-14T09:30:00.000Z/P3W/2019-08-22T09:30:00.000Z
```

The time selected in the date picker is `05:30`, but when converted to UTC, it should be `09:30`. For some reason, when run on the CI server, the selected time is not converted to UTC when it's saved.

This could be due to some edge case behaviour with [`cy.clock`](https://docs.cypress.io/api/commands/clock.html#Syntax) or [`luxon.DateTime`](https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html).

PR #521, opened 2019-08-15
Allow boundary timer events to be repositioned

### LINKED ISSUE #511: Allow repositioning of Boundary Timer Event on Tasks
Right now the Boundary Timer cannot be repositioned from where it was snapped onto the Task. You can only delete the timer and attach a new one.

Allow Boundary Timer to be repositioned only on the Task after embedding.

tuantruong625:
This PR will give users the ability to reposition the boundary timer event and it will snap to the closest anchor point when repositioning. 

**Video of enhancement**
https://drive.google.com/open?id=1yrCQI9WAt3Bor735vICqTJWseAZE-q_m

Fixes #511

### DIFF
--- a/src/components/nodes/boundaryTimerEvent/boundaryTimerEvent.vue
+++ b/src/components/nodes/boundaryTimerEvent/boundaryTimerEvent.vue
@@ -8,6 +8,7 @@ import timerEventIcon from '@/assets/timer-event-icon.svg';
 import boundaryEventSwitcher from '@/mixins/boundaryEventSwitcher';
 import { portGroups } from '@/mixins/portsConfig';
 import portsConfig from '@/mixins/portsConfig';
+import crownConfig from '@/mixins/crownConfig';
 import joint from 'jointjs';
 
 function getPointFromGroup(model, group) {
@@ -45,7 +46,7 @@ function snapToAnchor(coords, model) {
 
 export default {
   extends: BoundaryEvent,
-  mixins: [boundaryEventSwitcher, portsConfig],
+  mixins: [boundaryEventSwitcher, portsConfig, crownConfig],
   watch: {
     'node.definition.cancelActivity'(value) {
       this.renderBoundaryTimer(value);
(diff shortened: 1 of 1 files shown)


src/components/nodes/boundaryTimerEvent/boundaryTimerEvent.vue
```diff
@@ -87,15 +93,16 @@ export default {
...
       this.renderBoundaryTimer(this.node.definition.cancelActivity);
-      this.shape.set('elementMove', false);
     } else {
       this.$emit('remove-node', this.node);
     }
+
+    this.shape.listenTo(this.paper, 'element:pointerup', () => {
+      this.updateSnappingPosition(task);
+      this.updateCrownPosition();
+    });
```
chipit24: This will get called every time _any_ element is clicked. I think you should ensure it's only called for the current element (boundary timer event):

```js
this.shape.listenTo(this.paper, 'element:pointerup', cellView => {
  if (cellView.model !== this.shape) {
    return;
  }

  this.updateSnappingPosition(task);
  this.updateCrownPosition();
});
```

chipit24: This will get called every time _any_ element is clicked. I think you should ensure it's only called for the current element (boundary timer event):

```js
this.shape.listenTo(this.paper, 'element:pointerup', cellView => {
  if (cellView.model !== this.shape) {
    return;
  }

  this.updateSnappingPosition(task);
  this.updateCrownPosition();
});
```

src/components/nodes/boundaryTimerEvent/boundaryTimerEvent.vue
```diff
@@ -87,15 +93,16 @@ export default {
...
-      this.shape.position(x - (width / 2), y - (width / 2));
+      this.updateSnappingPosition(task);
       this.renderBoundaryTimer(this.node.definition.cancelActivity);
-      this.shape.set('elementMove', false);
     } else {
       this.$emit('remove-node', this.node);
     }
+
+    this.shape.listenTo(this.paper, 'element:pointerup', () => {
+      this.updateSnappingPosition(task);
```
chipit24: Since you're passing in task, should this not be called within the `if (task)` block?

PR #524, opened 2019-08-15
Check if container in inspector extension

### LINKED ISSUE #523: Allow inspector extensions to be added outside the default container
registerInspectorExtension() always adds it to the default, first container in the inspector panel. Doesn't allow for accordions. 

Needed for https://github.com/ProcessMaker/processmaker/issues/2244

nolanpro:
Fixes #523 

image

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -333,24 +333,43 @@ export default {
      * component is already registered, it is replaced by the new one.
      */
     registerInspectorExtension(node, config) {
-      const registeredIndex = node.inspectorConfig[0].items.findIndex(item => {
+      this.addToInspector(
+        this.inspectorItemsToAddTo(node, config),
+        this.existingConfigIndex(node, config),
+        config
+      );
+    },
+    addToInspector(inspectorItems, existingIndex, config) {
+      if (existingIndex === -1) {
+        inspectorItems.push(config);
+      } else {
+        inspectorItems[existingIndex] = config;
+      }
+    },
+    existingConfigIndex(node, config) {
+      return this.inspectorItems(node).findIndex(item => {
         return config.id && config.id === item.id;
       });
-        
-      let items;
-      if (config.container) {
-        // if it is a container, don't put it inside the default container
-        items = node.inspectorConfig[0].items;
+    },
+    isContainer(config)
+    {
+      return typeof config.container !== 'undefined';
+    },
+    inspectorItemsToAddTo(node, config) {
+      if (this.isContainer(config)) {
+        return this.inspectorItems(node);
       } else {
-        // if it is not a container, put it in the default container
-        items = node.inspectorConfig[0].items[0].items;
+        // if the config we're adding is not a container, put it in the default container
+        return this.defaultContainerItems(node);
       }
-
-      if (registeredIndex === -1) {
-        items.push(config);
-      } else {
-        items[registeredIndex] = config;
-      }
+    },
+    defaultContainerItems(node)
+    {
+      return this.inspectorItems(node)[0].items;
+    },
+    inspectorItems(node)
+    {
+      return node.inspectorConfig[0].items;
     },
     /**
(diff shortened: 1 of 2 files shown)


src/components/Modeler.vue
```diff
@@ -336,10 +336,20 @@ export default {
...
+        // if it is not a container, put it in the default container
+        items = node.inspectorConfig[0].items[0].items;
+      }
+
       if (registeredIndex === -1) {
-        node.inspectorConfig[0].items[0].items.push(config);
+        items.push(config);
       } else {
-        node.inspectorConfig[0].items[0].items[registeredIndex] = config;
+        items[registeredIndex] = config;
```
chipit24: I think this whole function is a little hard to read and can be refactored. There's a few things I've noticed, and you can let me know what you think :)

Whether or not `config` is a container, they both work with a common section of the `node` that's being modified: `node.inspectorConfig[0].items`. This path is going to exist and be the same for every node, and is the list of accordions for each node (currently every node only has one "default" accordion container, "Configuration"). This could be assigned to a variable to make it easier to read.

src/components/Modeler.vue
```diff
@@ -336,10 +336,20 @@ export default {
       const registeredIndex = node.inspectorConfig[0].items.findIndex(item => {
         return config.id && config.id === item.id;
       });
+        
+      let items;
+      if (config.container) {
```
chipit24: There's two if/else blocks in this function and it may help readability to break these out into separate functions.

I've recently learned about the idea of command query separation; having separate functions for executing an action or retrieving data. And it seems the 1st if/else block can qualify as a query and the 2nd as a command.

nolanpro:
@chipit24 i've done some refactoring here but need to run unit tests. I can't figure out how to get vue-cli-service test:unit tests/unit to run. I know we already have cypress but I don't think this can be tested with cypress (can it?). It needs a unit test runner like the built-in jest. Please add `"test-unit": "vue-cli-service test:unit tests/unit"` to package.json scripts and make sure a test file can `import Modeler from '../../src/components/Modeler.vue';` Thanks

chipit24:
@nolanpro We didn't set up unit tests; if you look at the scripts in `package.json` you'll see we only have commands for our Cypress tests. But that's on our radar and something we want to do soon. 

> Please add "test-unit": "vue-cli-service test:unit tests/unit" to package.json scripts and make sure a test file can import Modeler from '../../src/components/Modeler.vue';

If only it was that easy :p the reason we didn't set up any unit tests initially is because JointJS is so tightly coupled with all our components, and JointJS relies heavily on SVG functions that are not supported in JSDOM. JSDOM is what Jest uses as its virtual DOM when running tests. There was some initial work to refactor the codebase so JointJS can be mocked out in tests, but this was never made a priority.

tests/unit/Modeler.spec.js
```diff
@@ -0,0 +1,8 @@
+import Modeler from '../../src/components/Modeler.vue';
+
+describe('registerInspectorExtension', () => {
+    it('adds config with a container', () => {
+        // Placeholder
+        expect(typeof Modeler.created).toBe('function')
+    })
+})
```
chipit24: You can remove this file for now as it's causing our CI to fail, but you've definitely highlighted our need for unit tests and we should make this a priority going forward!

src/components/Modeler.vue
```diff
@@ -333,25 +333,44 @@ export default {
...
-      if (registeredIndex === -1) {
-        items.push(config);
+    },
+    existingConfigIndex(node, config) {
+      return this.inspectorItems(node).findIndex(item => {
+        return config.id && config.id === item.id;
+      });
+    },
+    isContainer(config)
+    {
```
chipit24: To keep things consistent, place the opening bracket on the same line as the function name.

nolanpro:
@chipit24 ok changes make, seems to be working. Thanks

chipit24:
I fixed the formatting for the other two functions as well ;)
Looks good, thanks! 💯

PR #533, opened 2019-08-20
Recreated ModdleElement for BoundaryEvents

### LINKED ISSUE #510: Uploading an xml with an Interrupting Boundary Timer does not behave as expected
**How to reproduce bug**

* Upload the supplied xml which has an Interrupting Boundary Timer
https://www.dropbox.com/s/b6huiw0yolqsau3/bpmnProcess.xml?dl=0
* Click on the New Boundary Timer Event
* Toggle the Interrupting checkbox

**Expected Behaviour**
The Boundary Timer Event Configuration should have a checked Interrupting Boundary Event Type after uploading. Toggling the Interrupting checkbox should change to the dashed, non-interrupting timer icon.

**Actual Behaviour**
The Interrupting checkbox is unchecked when cancelActivity='true' is not explicitly set in the xml. The xml plugin does not include default values. Toggling between Interrupting / Non-interrupting does not show visually but does appear to save in the xml download.

https://www.dropbox.com/s/oyvi9aocbnw6tok/InterruptingTimerError.mov?dl=0

Alternative ways to reproduce:
------
Scenario 1:
* Place a Task on the screen
* Drag an Intermediate Timer Event onto the Task until it attaches as a Boundary Timer Event
* Click undo/redo
* Attempt to toggle the Interrupting checkbox

Scenario 2:
* Place a Task on the screen in ProcessMaker
* Drag an Intermediate Timer Event onto the Task until it attaches as a Boundary Timer Event
* Save and refresh the page
* Attempt to toggle the Interrupting checkbox

lisa-fehr:
Resolves #510 

**This can be a stand alone from Boundary Timer Events. It only alters the data parsed for Boundary Timer Events.**

Adding cancelActivity after the creation of the ModdleElement made it untoggle-able in the sidebar.

Right now it toggles, but it has to update the definition in sourceRef and targetRef on every node pointing to the recreated ModdleElement.

Another option is using something like`node.definition.id` in linkConfig for getting targetShape and sourceShape, rather than matching the definition exactly. The recreated definition no longer matches.

```php
this.sourceShape = this.graph.getElements().find(element => {
  return element.component && element.component.node.definition === this.node.definition.get('sourceRef');
});
```

Example I used to test: 
Screen Shot 2019-09-03 at 9 57 45 AM

Boundary Events with default of no cancelActivity, cancelActivity false and cancelActivity true:
https://www.dropbox.com/s/3d8a9hyvzmzzbl0/bpmnProcessForInterruptingToggleDefault.xml?dl=0

Boundary Events in pools:
https://www.dropbox.com/s/c7ienbpf51frxcu/boundaryTimersInPools.xml?dl=0

Reference #581

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -652,7 +652,6 @@ export default {
     replaceDefinition(definition, boundaryEvent, process) {
       const definitionIndex = process.flowElements.indexOf(definition);
       this.definitions.get('rootElements').find(currentProcess => currentProcess === process).flowElements[definitionIndex] = boundaryEvent;
-      console.log(this.definitions);
     },
     ensureCancelActivityIsAddedToBoundaryEvents(process) {
       this.getBoundaryEvents(process).forEach(definition => {
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -462,6 +462,7 @@ export default {
     },
     // Parses our definitions and graphs and stores them in our id based lookup model
     parse() {
+      this.addCancelActivityToBoundaryEvents();
       // Get the top level objects
       // All root elements are either bpmn:process or bpmn:collaboration types
       // There should only be one collaboration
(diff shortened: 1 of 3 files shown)


src/components/Modeler.vue
```diff
@@ -586,6 +587,42 @@ export default {
         }
       });
     },
+    addCancelActivityToBoundaryEvents() {
```
JCook21: I think there's an awful lot going on in this method and I'm not sure if the code really reveals its intention that well. Could we refactor this, perhaps into multiple methods, so that the code is clearer?

src/components/Modeler.vue
```diff
@@ -586,6 +588,35 @@ export default {
...
+      });
+      boundaryEvent.$parent = definition.$parent;
+      if (definition.get('outgoing').length) {
+        boundaryEvent.set('outgoing', definition.get('outgoing'));
+      }
+      return boundaryEvent;
+    },
+    replaceDefinition(definition, boundaryEvent, process) {
+      const definitionIndex = process.flowElements.indexOf(definition);
+      this.definitions.get('rootElements').find(currentProcess => currentProcess === process).flowElements[definitionIndex] = boundaryEvent;
```
chipit24: Is there a reason mutating `process` directly doesn't work? It seems the following should work just the same:
```js
process.flowElements[definitionIndex] = boundaryEvent;
```

src/components/Modeler.vue
```diff
@@ -586,6 +588,35 @@ export default {
...
+      return process.flowElements
+        ? process.flowElements.filter(({$type}) => $type === 'bpmn:BoundaryEvent')
+        : [];
+    },
+    createBoundaryEvent(definition) {
+      let boundaryEvent = this.moddle.create('bpmn:BoundaryEvent', {
+        ...definition,
+        cancelActivity: definition.get('cancelActivity'),
+        attachedToRef: definition.get('attachedToRef'),
+      });
```
chipit24: I think we can reuse the same function defined in `/src/components/nodes/boundaryEvent/index.js` to create this definition:
```js
import boundaryEventConfig from './nodes/boundaryEvent';
...
const boundaryEvent = boundaryEventConfig.definition(this.moddle, this.$t);
```

Though we'd have to be careful to re-assign the properties from the original `definition`. The benefit here is that now we have a single place where the boundary event definition is declared.

**Note:** Since `boundaryEvent` isn't reassigned, you can use `const` instead of `let`.

src/components/Modeler.vue
```diff
@@ -586,6 +588,35 @@ export default {
...
+      boundaryEvent.$parent = definition.$parent;
+      if (definition.get('outgoing').length) {
```
chipit24: This is just my opinion, but I prefer to be more explicit with `length` checks. What are your thoughts on changing this to:
```
definition.get('outgoing').length > 0
```

src/components/Modeler.vue
lisa-fehr: Seems to work

lisa-fehr:
I'm hesitant to rebase this with `581-boundary-event` until segment flows are fixed.

src/components/Modeler.vue
```diff
@@ -629,6 +632,36 @@ export default {
         this.renderPaper();
       });
     },
+
+    getBoundaryEvents(process) {
+      return process.flowElements
+        ? process.flowElements.filter(({$type}) => $type === 'bpmn:BoundaryEvent')
+        : [];
```
chipit24: Not sure if I commented on this before 🤔 But I think `process.get('flowElements')` will always return an array, so this whole thing could just be:
```js
return process.get('flowElements').filter(({$type}) => $type === 'bpmn:BoundaryEvent');
```

lisa-fehr: It is `Cannot read property 'filter' of undefined` when you load up a new process in processmaker. So that doesn't seem to be true.

lisa-fehr: Done

PR #540, opened 2019-08-21
Ensure rendering is complete before running view dependant code

### LINKED ISSUE #538: Async rendering: Red error border disappears after undo
This is a partial regression of https://github.com/ProcessMaker/modeler/issues/412.

**Steps to reproduce:**
1. Enable auto validation:
   Screen Shot 2019-08-20 at 4 55 27 PM
2. Modify the start event (drag it to a new position or change its name or ID).
3. Press Undo.

**Current behaviour:**
* Although there is a validation error with the start event, the red border around the start event has disappeared and doesn't re-appear.

**Expected behaviour:**
* As there is a validation error for the start event, the red border around it should remain visible:
  Screen Shot 2019-08-20 at 4 58 14 PM

**Note:**
 * Fixing this bug should fix the failing test for "updates validation after undo/redo" without causing further tests to fail.

chipit24:
Fixes #538.

The issue appeared to be that when the code which sets the error highlighting ran, the view for the shape wasn't fully rendered. There is no way in JointJS to determine if rendering is in progress, and `paper.isFrozen()` returns `false` even though rendering of the views is not compete.

This PR passes a flag to components that can be used to conditionally run view-dependant code after rendering is complete; this resolves the issue for setting the highlighting on the shape.

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -87,6 +87,7 @@
       :moddle="moddle"
       :nodeRegistry="nodeRegistry"
       :root-elements="definitions.get('rootElements')"
+      :isRendering="isRendering"
       @add-node="addNode"
       @remove-node="removeNode"
       @set-cursor="cursor = $event"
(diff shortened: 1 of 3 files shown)


lisa-fehr:
When I have a Start Event and then place a Task, they both have the red outline. When I click undo, Start is still outlined in red but the Task is not.

PR #544, opened 2019-08-21
Added ports to text annotation

### LINKED ISSUE #539: Can undo/redo modifying association flow vertices
This is a partial regression of https://github.com/ProcessMaker/modeler/issues/412.

**Steps to reproduce:**
1. Drag a `New Text Annotation` onto the screen
2. Add an `Association Flow` like pictured to the `Start Event`. It must have a bend. Screen Shot 2019-08-21 at 9 52 37 AM
3. Click `Undo` 

**Current behaviour:**
* The test error is: "Uncaught TypeError: Cannot read property 'x' of undefined.
Screen Shot 2019-08-21 at 9 41 54 AM

In Modeler there is an "Uncaught TypeError: Cannot read property 'x' of undefined"
Screen Shot 2019-08-21 at 9 54 32 AM

**Expected behaviour:**
* The number of vertices should reduce by one and the test should pass. There should be no console errors in Modeler.

**Note:**
 * Fixing this bug should fix the failing test for "Can undo/redo modifying association flow vertices" without causing further tests to fail.

lisa-fehr:
Resolves #539 

I propose adding portsConfig to the Text Annotation. This keeps the connection point of the line when re-doing, and makes the test pass.

### DIFF
--- a/src/components/nodes/textAnnotation/textAnnotation.vue
+++ b/src/components/nodes/textAnnotation/textAnnotation.vue
@@ -6,11 +6,12 @@
 import { shapes, util } from 'jointjs';
 import connectIcon from '@/assets/connect-artifacts.svg';
 import crownConfig from '@/mixins/crownConfig';
+import portsConfig from '@/mixins/portsConfig';
 
 export const maxTextAnnotationWidth = 160;
 export default {
   props: ['graph', 'node', 'id'],
-  mixins: [crownConfig],
+  mixins: [crownConfig, portsConfig],
   data() {
     return {
       shape: null,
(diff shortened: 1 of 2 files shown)


lisa-fehr:
Not sure if passing one test from the UndoRedo spec, revealed a new failing test, or caused the failing test.

test: undo/redo boundary timer event

lisa-fehr:
@chipit24 I did the 'undo/redo boundary timer event' test manually and it didn't need the 2 undos after attaching a Boundary Timer to a Task anymore. I removed the extra undo and it still passes after running all tests. Interesting.

PR #547, opened 2019-08-21
Fixes issue with setting the label for a text annotation introduced by async rendering

### LINKED ISSUE #503: Slow Loads / Optimizations
When rendering this process (attached) the whole modeler becomes unresponsive and the browser starts using lots of CPU cycles. We need to find ways to optimize for larger processes so this isn't an issue going forward. 

[total_accounts_payable_w_f.json.txt](https://github.com/ProcessMaker/modeler/files/3469350/total_accounts_payable_w_f.json.txt)

image

Profile of the process:

image

### LINKED ISSUE #546: Async rendering: adding a new text annotation causes a reference error
Related to #503. Dropping a new text annotation onto a process causes an error.

JCook21:
This fixes #546.

Screen Shot 2019-08-21 at 4 46 09 PM

There's a possibility that fixing this issue and switching to async rendering has made the text in the annotation a little off centre vertically (see screenshot). I think there are three possible ways forward if this is the case:

1. Put more time into this now to see if we can work out what has changed.
2. Log an issue to investigate it further later and close this for now.
3. Leave things as they are as if things are off centre it's only by a very small amount, a few pixels.

@velkymx could you comment on if the vertical alignment has changed and how you would like us to proceed if so?

### DIFF
--- a/src/components/nodes/textAnnotation/textAnnotation.vue
+++ b/src/components/nodes/textAnnotation/textAnnotation.vue
@@ -6,11 +6,8 @@
 import { shapes, util } from 'jointjs';
 import connectIcon from '@/assets/connect-artifacts.svg';
 import crownConfig from '@/mixins/crownConfig';
-import { highlightPadding } from '@/mixins/crownConfig';
 
 export const maxTextAnnotationWidth = 160;
-export const textAnnotationLabelPadding = 15;
-
 export default {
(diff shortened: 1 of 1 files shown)


velkymx:
This is a low priority issue.

JCook21:
> This is a low priority issue.

I'll log an issue for this so we don't lose sight of it.

PR #553, opened 2019-08-22
Implement feature to add markers

### LINKED ISSUE #535: Implement option to add a marker in the upper right corner of elements
Required for PRD: [Vocabularies v2.0](https://processmaker.atlassian.net/wiki/spaces/PM4/pages/568262657/Package+Vocabulary)

image

caleeli:
Resolves #535 

This PR includes:

- A Mixing to add reactive markers to BPMN elements
- Add markers to tasks, script tasks, service tasks, manual tasks, call activity
- Add markers to start events
- Add e2e tests 

image


src/components/Modeler.vue
```diff
@@ -327,6 +327,15 @@ export default {
     highlightNode(node) {
       store.commit('highlightNode', node);
     },
+    /**
+     * Register a mixin into a node component.
+     * Used during "modeler-before-init"
+     *
+     */
+    registerComponentMixin(component, mixin) {
```
chipit24: A suggestion: Since this method is fairly simple, do you think it makes more sense to add it as a parameter to `registerNode`? The one drawback of doing this is that many places may already be using `registerNode`, and modifying the function's signature will require a slight re-write of the places the function is called.

My reason for suggesting this is that, the only thing between `modeler-before-init` and `modeler-init` is a single function call, and the name isn't very meaningful, as it's still part of initialization.

src/components/Modeler.vue
```diff
@@ -327,6 +327,15 @@ export default {
...
+      const mixins = component.mixins === undefined ? component.mixins = [] : component.mixins;
```
chipit24: The intention of this code could be improved. You're doing a strict equality comparison of `component.mixins` to `undefined`, but what if the value happens to be `null`, or `false`, or something else?

Since we're dealing with Vue components, we can safely assume that if `mixins` is defined, it will be an array. However, there is no need for the transient `mixins` variable, as I think it hides that fact that you're mutating a property on an existing object.

I think this reads a little better:
```js
if (!component.mixins) {
  component.mixins = [];
}

component.mixins.push(mixin);
```

The "happy path" (pushing a new mixin onto the `mixins` array) remains clearly on it's own, not indented, and the "special case" of initializing an empty `mixins` array is contained in an `if` block.

src/components/nodes/callActivity/callActivity.vue
```diff
@@ -39,9 +41,10 @@ export default {
       const { height } = this.shape.size();
 
       if (labelHeight + labelPadding !== height) {
-        const newHeight = Math.max(labelHeight + 15, taskHeight);
+        const newHeight = Math.max(labelHeight + 15 + 2 * MARKER_SIZE, taskHeight);
```
chipit24: It looks like 15 was a magic number that should have been cleaned up, but now there is another magic number being added (2) and this whole expression is now harder to understand.

It seems that 15 should be a reference to `labelPadding`, and the expressions should be grouped for better readability:
```js
(labelHeight + labelPadding) + (MARKER_SIZE * 2)
```

Once you do that, you'll see that `labelHeight + labelPadding` is used above as well, and you could assign it a meaningful variable name.

However, this code looks very similar to code in `textAnnotation.vue`; take a look at the `updateNodeText` method in that component.

src/mixins/hasMarkers.js
```diff
@@ -0,0 +1,100 @@
...
+export const MARKER_SIZE = 16;
+export const MARKER_PADDING = 4;
+
+/**
+ * Prepare nodes for the Markers to be inserted inside the shapes.
+ *
+ * @param string selector 
+ */
+export function markersMarkup(selector) {
+  let children = [];
```
chipit24: Could use `const` here since `children` is not re-assigned.

src/components/nodes/task/task.vue
```diff
@@ -7,14 +7,16 @@ import { util } from 'jointjs';
 import connectIcon from '@/assets/connect-elements.svg';
 import crownConfig from '@/mixins/crownConfig';
 import portsConfig from '@/mixins/portsConfig';
+import hasMarkers from '@/mixins/hasMarkers';
+import {MARKER_SIZE} from '@/mixins/hasMarkers';
```
chipit24: Same thing here, regarding formatting.

src/mixins/hasMarkers.js
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Define attributes for the markers.
+ *
+ * @param string name 
+ * @param object base 
+ * @param int length 
+ */
+export function markersAttrs(name, base, inverted = 1) {
+  let attrs = {};
```
chipit24: Same thing here; can use `const`.

src/mixins/hasMarkers.js
```diff
@@ -0,0 +1,100 @@
...
+      // Get shape size
+      const { width, height } = this.shape.size();
+      for (let position in markers) {
+        // Align markers (left-right)
+        let paddingX = this.shape.attr(position + '/ref-padding-x');
+        paddingX = paddingX === undefined ? MARKER_PADDING : paddingX;
+        let paddingY = this.shape.attr(position + '/ref-padding-y');
+        paddingY = paddingY === undefined ? MARKER_PADDING : paddingY;
+
+        if (position.indexOf('Right') !== -1) {
```
chipit24: Since you don't care about the index, use `includes` to make this more readable:
```js
if (position.includes('Right')) {
```

src/mixins/hasMarkers.js
```diff
@@ -0,0 +1,100 @@
...
+      inmediate: true,
+      deep: true,
+      handler() {
+        this.refreshMarkers();
+      },
+    },
+  },
+  methods: {
+    refreshMarkers(markers = this.markers) {
+      // Get shape size
```
chipit24: I'm not sure this comment is necessary; I think it's clear enough from the code what values we're retrieving.

src/mixins/hasMarkers.js
```diff
@@ -0,0 +1,100 @@
...
+      const { width, height } = this.shape.size();
+      for (let position in markers) {
+        // Align markers (left-right)
```
chipit24: Instead of a comment like this, do you think you can break out the relevant code into a meaningful function? Something like `alignMarkersFromLeftToRight(position)`. You can apply the same technique to the other blocks in this function.

src/components/nodes/startEvent/shape.js
```diff
@@ -11,14 +12,26 @@ export default shapes.standard.Circle.extend({
...
       'image': { 'ref-x': 5, 'ref-y': 5, ref: 'circle', width: 26, height: 26 },
+      ...markersAttrs('topLeft', { ref: 'circle', 'ref-y': -20, 'ref-padding-x': 0 }),
+      ...markersAttrs('topCenter', { ref: 'circle', 'ref-y': -20 }),
+      ...markersAttrs('topRight', { ref: 'circle', 'ref-x': 26, 'ref-y': -20, 'ref-padding-x': 0 }, -1),
+      ...markersAttrs('bottomLeft', { ref: 'circle', 'ref-padding-x': 0, 'ref-padding-y': -40 }),
+      ...markersAttrs('bottomCenter', { ref: 'circle', 'ref-padding-y': -40 }),
+      ...markersAttrs('bottomRight', { ref: 'circle', 'ref-x': 26, 'ref-padding-x': 0, 'ref-padding-y': -40 }, -1),
     },
 
   }, shapes.standard.Circle.prototype.defaults),
```
chipit24: You make a call to `this.refreshMarkers` for _Task_ and _Call Activity_, but not for _Start Event_, why is that?

src/components/nodes/task/task.vue
```diff
@@ -39,9 +46,10 @@ export default {
       const { height } = this.shape.size();
 
       if (labelHeight + labelPadding !== height) {
-        const newHeight = Math.max(labelHeight + 15, taskHeight);
+        const newHeight = Math.max(labelHeight + 15 + 2 * MARKER_SIZE, taskHeight);
```
chipit24: Same thing here, about magic numbers and readability.

tests/e2e/specs/Markers.spec.js
```diff
@@ -0,0 +1,62 @@
...
+      .find('image[joint-selector*=topRight]:first')
+      .should('have.attr', 'xlink:href')
+      .and('match', /^\/img\/book.*svg$/);
+    getElementAtPosition(taskPosition)
+      .find('image[joint-selector*=topRight]:nth-child(2)')
+      .should('have.attr', 'xlink:href')
+      .and('match', /^\/img\/clock.*svg$/);
+  });
+
+});
```
chipit24: Nice tests! 💯

PR #574, opened 2019-08-28
Update dependencies

JCook21:
Updated dependencies in line with suggestions from Alan. This has improved things but there are still 2 failing tests locally. Creating this PR to make sure things are the same in CI.


lisa-fehr:
Looks like everything is passing when linking the newest version of screen builder to modeler.

PR #586, opened 2019-08-30
Fix start timer event

### LINKED ISSUE #585: Start timer event config not being saved when periodicity is not set to "week"
**Steps to reproduce:**
1. Drag a _Start Timer Event_ onto the grid.
2. Select the _Start Timer Event_ and change the _Repeat every_ dropbox value (periodicity) to anything other than _week_.
3. Deselect the _Start Timer Event_, and then reselect it.

**Current behaviour:**
The change to the periodicity value isn't persisted. Also, when the periodicity **isn't** set to _week_, changes to any properties other than _Start date_ aren't persisted either.

**Expected behaviour:**
Changes to the periodicity, and all other properties, should be persisted and not be reset when deselecting and reselecting the _Start Timer Event_.

**Video of bug:**
https://www.dropbox.com/s/xigpgx5j2jox17d/TimerStartEventNotSavingConfig.mov?dl=0

### LINKED ISSUE #588: Start timer event selected days not being reset
**Steps to reproduce:**
1. Drag a _Start Timer Event_ onto the grid and highlight it.
2. Expand the _Timing Control_ accordion in the inspector panel and select multiple days for _Repeat on_ (e.g. M, T, W).
3. Change the periodicity from _week_ to _month_.
4. Download the XML and inspect the `timeCycle` property for `timerEventDefinition`.

**Current behaviour:**
The selected weekdays are no longer relevant for the monthly period, but still appear in the timer expression, e.g.:
```
2019-08-30T19:38:00.000Z|R/2019-09-02T19:38:00.000Z/P1M|R/2019-09-03T19:38:00.000Z/P1M|R/2019-09-04T19:38:00.000Z/P1M
```

**Expected behaviour:**
The selected weekdays should either be reset, or ignored when considering the monthly period, e.g.:
```
R/2019-09-04T19:38:00.000Z/P1M
```

chipit24:
Fixes #585.

**Note:** ~There is still another issue related to Start Timer Events which I will be fixing separately: https://github.com/ProcessMaker/modeler/issues/588.~ It would appear that this PR actually fixes #588 as well.

### DIFF
--- a/src/components/inspectors/TimerExpression.vue
+++ b/src/components/inspectors/TimerExpression.vue
@@ -170,7 +170,7 @@ export default {
 
       return 'You must select at least one day.';
     },
-    iso8601FormattedDate() {
+    dateIntervalString() {
       if (this.isWeeklyPeriodSelected) {
         return this.getFormattedDateWithWeekdayIntervals();
       }
@@ -195,7 +195,7 @@ export default {
     },
   },
   watch: {
-    iso8601FormattedDate() {
+    dateIntervalString() {
       this.update();
     },
   },
@@ -209,8 +209,8 @@ export default {
 
       this.selectedWeekdays.forEach(day => {
         const weekDayDate = this.getWeekDayDate(day);
-        const cycle = this.getIso8601FormattedDateString(weekDayDate);
-        dateIntervals.push(cycle);
+        const isoDateString = this.getIso8601FormattedDateString(weekDayDate);
+        dateIntervals.push(isoDateString);
       });
 
       return dateIntervals.join('|');
@@ -247,7 +247,7 @@ export default {
       ];
     },
     update() {
-      this.$emit('input', this.iso8601FormattedDate);
+      this.$emit('input', this.dateIntervalString);
     },
     /**
      * Parse an ISO8601 expression to get the timer configuration
--- a/src/components/inspectors/TimerExpression.vue
+++ b/src/components/inspectors/TimerExpression.vue
@@ -168,15 +168,18 @@ export default {
       return 'You must select at least one day.';
     },
     iso8606Expression() {
-      if (this.selectedWeekdays.length === 1 && this.sameDay) {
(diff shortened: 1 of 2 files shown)


src/components/inspectors/TimerExpression.vue
```diff
@@ -168,15 +168,18 @@ export default {
       return 'You must select at least one day.';
     },
     iso8606Expression() {
-      if (this.selectedWeekdays.length === 1 && this.sameDay) {
+      if ((this.selectedWeekdays.length === 1 && this.sameDay) || this.periodicity !== 'week') {
```
JCook21: Could you extract this into a method? It would make the intent of this code much clearer.

src/components/inspectors/TimerExpression.vue
```diff
@@ -168,15 +168,18 @@ export default {
...
         return this.getCycle(DateTime.fromISO(this.startDate, { zone: 'utc' }));
       }
 
       const expression = [];
       expression.push(this.startDate);
-      this.selectedWeekdays.forEach(day => {
-        expression.push(this.getCycle(this.getWeekDayDate(day)));
-      });
+
+      if (this.periodicity === 'week') {
```
JCook21: Same here: a method would reveal the intention much better.

src/components/inspectors/TimerExpression.vue
```diff
@@ -168,15 +168,18 @@ export default {
       return 'You must select at least one day.';
     },
     iso8606Expression() {
```
JCook21: I think this needs a better name. How about `iso8601FormattedDate`? It's not clear what an 'Expression' is in this context and there's a typo in the name ISO 8606 deals with plastics (https://standards.globalspec.com/std/694731/ISO%208606)...

chipit24:
> This is much better, thank you! If you have the time and inclination `this.cycle()` could use a better name but that's much more minor than the other issues.

I'd definitely like to clean this up. It's not clear what a cycle is, and `getCycle` simply calls `makeCycle`.

PR #602, opened 2019-09-04
Code Cleanup

JCook21:
Extracted method to keep code DRY and removed unused method.

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -113,7 +113,7 @@ import NodeIdGenerator from '../NodeIdGenerator';
 import Process from './inspectors/process';
 import runningInCypressTest from '@/runningInCypressTest';
 
-import { faMapMarked, faMinus, faPlus } from '@fortawesome/free-solid-svg-icons';
+import { faMinus, faPlus } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 
 import { id as poolId } from './nodes/pool';
@@ -226,9 +226,6 @@ export default {
         return invalidIds;
       }, []);
     },
-    mapIcon() {
-      return faMapMarked;
-    },
     plusIcon() {
       return faPlus;
     },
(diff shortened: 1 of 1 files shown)


src/components/Modeler.vue
```diff
@@ -113,7 +113,7 @@ import NodeIdGenerator from '../NodeIdGenerator';
 import Process from './inspectors/process';
 import runningInCypressTest from '@/runningInCypressTest';
 
-import { faMapMarked, faMinus, faPlus } from '@fortawesome/free-solid-svg-icons';
```
JCook21: `faMapMarked` was only used in the removed method below.

src/components/Modeler.vue
```diff
@@ -226,9 +226,6 @@ export default {
         return invalidIds;
       }, []);
     },
-    mapIcon() {
```
JCook21: This method was marked as unused by PHPStorm so I removed it.

src/components/Modeler.vue
```diff
@@ -237,17 +234,17 @@ export default {
     },
   },
   methods: {
+    addWarning(warning) {
```
JCook21: This method extraction removes duplication and keeps things DRY.

JCook21:
@sanjacornelius @ryancooley I think the PR that was merged yesterday to allow warnings to be pushed introduces a subtle bug. Since we built this functionality to only accept an array of warnings when the modeler starts (which is what we agreed to) the code has no concept of adding warning to the existing ones. With the changes that @sanjacornelius made yesterday I think the following scenario is possible:

1. When the modeler loads warnings are loaded which result from the larval app parsing the BPMN file
2. These are immediately populated in the warnings popup
3. Any subsequent event that dispatches 'warnings' will result in a new collection of warnings being displayed, removing the ones added by the backend app.

Since warnings are displayed by dispatching an event we will have no idea which event will get processed last, meaning this will effectively cause a race condition.

@velkymx could you comment on whether you'd like us to spend some time on this? Let's chat about it at our standup. If we want to proceed with this I'd like to add a unit test using Vue test utils to prove that this bug exists before going any further with fixing this.

PR #603, opened 2019-09-04
Refactor 581 boundary event

chipit24:
This PR removed duplicated and unused code, and refactors a couple functions to help improve readability.

### DIFF
--- a/src/components/nodes/boundaryEvent/boundaryEvent.vue
+++ b/src/components/nodes/boundaryEvent/boundaryEvent.vue
@@ -12,7 +12,7 @@ import { id as manualTaskId } from '@/components/nodes/manualTask';
 import { id as scriptTaskId } from '@/components/nodes/scriptTask';
 
 export default {
-  props: ['graph', 'node', 'id'],
+  props: ['graph', 'node'],
   mixins: [crownConfig],
   data() {
     return {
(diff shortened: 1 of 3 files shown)


src/components/nodes/boundaryEvent/boundaryEvent.vue
```diff
@@ -50,50 +50,28 @@ export default {
         .filter(model => model.component)
         .find(model => taskIds.includes(model.component.node.type));
     },
-    updateBoundaryShape(dashLength) {
+    setShapeStrokeDasharry(dashLength) {
```
JCook21: @chipit24 is `Dasharry` a typo? From looking at the body of the method it seems to be a misspelling of `Dasharray`.

Even if you correct the typo, I'm not sure if this is a good name. Can you think of a better one?

src/components/nodes/boundaryEvent/boundaryEvent.vue
```diff
@@ -50,50 +50,28 @@ export default {
...
-        'ref-x': 5,
-        'ref-y': 5,
-        'width': bounds.get('width') - 10,
-        'height': bounds.get('height') - 10,
-      },
-    });
+    const { x, y, width, height } = this.node.diagram.bounds;
+    this.shape.position(x, y);
+    this.shape.resize(width, height);
+    this.shape.attr('label/text', this.node.definition.get('name'));
```
JCook21: Could this all be extracted to a new method? That would make the intention clearer IMO.

PR #605, opened 2019-09-04
Allow boundary events to snap to edges of tasks

### LINKED ISSUE #593: Allow boundary events to snap to edges of tasks
As a user, when I drop a boundary event onto a task it should snap to the nearest connection point on the closest edge.

**Acceptance Criteria**
- Each task edge should have 3 connection points
- When a boundary event is released over a task it snaps to the closest connection point
- Connection points are distributed on the left, right and middle of each side
- Connection points do not overlap at corners
- Connection points are evenly distributed on each side
- Boundary events can be moved between connection points

lisa-fehr:
Related #593 

It snaps to edges but it doesn't have as many connections as required. Need to expand that number.

### DIFF
--- a/src/components/nodes/boundaryEvent/boundaryEvent.vue
+++ b/src/components/nodes/boundaryEvent/boundaryEvent.vue
@@ -11,38 +11,7 @@ import { id as callActivityId } from '@/components/nodes/callActivity';
 import { id as manualTaskId } from '@/components/nodes/manualTask';
 import { id as scriptTaskId } from '@/components/nodes/scriptTask';
 import portsConfig from '@/mixins/portsConfig';
-import { portGroups } from '@/mixins/portsConfig';
-import { g } from 'jointjs';
-
-function getPointFromGroup(model, group) {
-  const { x: shapeX, y: shapeY } = model.position();
-  const { x, y } = Object.values(model.getPortsPositions(group))[0];
-  return new g.Point(shapeX + x, shapeY + y);
-}
-
-function getPortPoints(model) {
-  return portGroups.filter(group => group !== 'absolute').map(group => getPointFromGroup(model, group));
-}
-
-function closestPort(model, anchorReference) {
-  return getPortPoints(model).sort((p1, p2) => {
-    const referencePoint = new g.Point(anchorReference.x, anchorReference.y);
-    return referencePoint.distance(p1) - referencePoint.distance(p2);
-  })[0];
-}
-
-function hasPorts(model) {
-  return Object.values(model.getPortsPositions(portGroups[0])).length > 0;
-}
-
-function snapToAnchor(coords, model) {
-  if (!hasPorts(model)) {
-    const { x, y } = model.position();
-    const { width, height } = model.size();
-    return new g.Point(x - (width / 2), y - (height / 2));
-  }
-  return closestPort(model, coords);
-}
+import { getAnchorCoordinates } from '@/snapToAnchor';
 
 export default {
   props: ['graph', 'node'],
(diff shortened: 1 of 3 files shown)


src/components/nodes/boundaryEvent/boundaryEvent.vue
```diff
@@ -78,6 +111,12 @@ export default {
       this.shape.attr('label/text', this.node.definition.get('name'));
       this.shape.component = this;
     },
+    updateSnappingPosition(task) {
```
JCook21: Would a better name for this be `updateShapePosition()`? From looking at the code it seems that the shape is thing being mutated in this method.

src/components/nodes/boundaryEvent/boundaryEvent.vue
```diff
@@ -10,10 +10,43 @@ import { id as taskId } from '@/components/nodes/task';
...
+}
+
+function snapToAnchor(coords, model) {
+  if (!hasPorts(model)) {
+    const { x, y } = model.position();
+    const { width, height } = model.size();
+    return new g.Point(x - (width / 2), y - (height / 2));
+  }
+  return closestPort(model, coords);
+}
```
JCook21: I really like all of the code in these functions: it seems well factored, named and expressive. I'm wondering why it written as functions inside this module file though. Is there a reason why you didn't want to make them methods of the component? If it's because you think they collectively represent a unit of functionality (which they seem to) how about moving them into another file? You could them just export `snapToAnchor()`, which seems to be the single entry point into this functionality.

src/components/nodes/boundaryEvent/boundaryEvent.vue
```diff
@@ -10,10 +10,43 @@ import { id as taskId } from '@/components/nodes/task';
...
+    const referencePoint = new g.Point(anchorReference.x, anchorReference.y);
+    return referencePoint.distance(p1) - referencePoint.distance(p2);
+  })[0];
+}
+
+function hasPorts(model) {
+  return Object.values(model.getPortsPositions(portGroups[0])).length > 0;
+}
+
+function snapToAnchor(coords, model) {
```
JCook21: This is the one name in this group that I don't like. This function doesn't snap the event to an anchor but it returns a point that the event should be anchored to. Can you think of a better name for this?

lisa-fehr: I tried getAnchorCoordinates. But maybe there is something better.

src/components/nodes/boundaryEvent/boundaryEvent.vue
lisa-fehr: What about making it a mixin?

src/components/nodes/boundaryEvent/boundaryEvent.vue
```diff
@@ -10,10 +10,12 @@ import { id as taskId } from '@/components/nodes/task';
 import { id as callActivityId } from '@/components/nodes/callActivity';
 import { id as manualTaskId } from '@/components/nodes/manualTask';
 import { id as scriptTaskId } from '@/components/nodes/scriptTask';
+import portsConfig from '@/mixins/portsConfig';
+import { getAnchorCoordinates } from '@/snapToAnchor';
```
JCook21: Do we need braces if it's a default export? Doesn't need fixing now, minor point. :)

lisa-fehr:
There is a bug with this where if you click undo/redo, or import a xml with Boundary Events, they start in the wrong location and snap to the correct anchor when you click on the misplaced Boundary Event.

JCook21:
@lisa-fehr I think this is good. Is this good to merge or do we still need to fix a failure before then?

PR #612, opened 2019-09-06
Reposition around the border of the element with 3 connection points per side

### LINKED ISSUE #593: Allow boundary events to snap to edges of tasks

lisa-fehr:
Resolves #593


chipit24:
I think this is now ready for review/merge.

PR #613, opened 2019-09-06
Move boundary events between tasks

### LINKED ISSUE #591: Boundary Events can be moved between Tasks
As a user, boundary events should be draggable between tasks and connectors.

**Acceptance Criteria**
- Can drag boundary events from one task to another
- If the user moves the event over a task type that it cannot be attached to the task will turn red.
- If the user attempts to drop the event onto an invalid task type it will 'snap back' to its starting position.
- Not allowing boundary event to be dragged onto paper will be addressed in the future

chipit24:
Fixes #591.

### DIFF
--- a/src/components/nodes/boundaryEvent/boundaryEvent.vue
+++ b/src/components/nodes/boundaryEvent/boundaryEvent.vue
@@ -108,41 +108,37 @@ export default {
       this.attachBoundaryEventToTask(task);
       this.updateShapePosition(task);
     },
-    listenForValidDropTargets() {
-      this.shape.listenTo(this.paper, 'element:pointerdown', cellView => {
-        if (cellView.model !== this.shape) {
-          return;
-        }
+    attachToValidTarget(cellView) {
+      if (cellView.model !== this.shape) {
+        return;
+      }
 
-        this.validPosition = this.shape.position();
-        this.shape.listenToOnce(this.paper, 'cell:pointerup blank:pointerup', this.moveBoundaryEventIfOverTask);
-      });
+      this.validPosition = this.shape.position();
+      this.shape.listenToOnce(this.paper, 'cell:pointerup blank:pointerup', this.moveBoundaryEventIfOverTask);
     },
     isPoolShape(model) {
       return model.component.node.type === 'processmaker-modeler-pool';
     },
-    listenForInvalidDropTargets() {
-      this.shape.on('change:position', () => {
-        const targetElement = this.graph
-          .findModelsUnderElement(this.shape)
-          .filter(model => model.component)[0];
+    turnInvalidTargetRed() {
+      const targetElement = this.graph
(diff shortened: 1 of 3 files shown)


src/components/nodes/boundaryEvent/boundaryEvent.vue
```diff
@@ -76,32 +81,87 @@ export default {
...
+    },
   },
   async mounted() {
     this.shape = new EventShape();
     this.setShapeProperties();
     this.shape.addTo(this.graph);
     await this.$nextTick();
-    this.attachBoundaryEventToTask();
+    this.listenForValidDropTargets();
+    this.listenForInvalidDropTargets();
```
JCook21: I wonder about the naming of these methods. They're about registering event handlers, could the naming reflect that? I also wonder if having this in multiple methods in this file also seems a bit smelly. Should these be in a separate file or at least consolidated somehow?

src/components/nodes/boundaryEvent/boundaryEvent.vue
```diff
@@ -76,32 +81,87 @@ export default {
...
 
-      this.shape.listenTo(this.paper, 'element:pointerup', cellView => {
+      if (!task) {
+        this.shape.position(this.validPosition.x, this.validPosition.y);
+        this.updateCrownPosition();
+        return;
+      }
+
+      if (task === this.shape.getParentCell()) {
+        this.updateShapePosition(task);
```
JCook21: `attachBoundaryEventToTask` includes a call to `updateShapePosition` on line 94. Is it necessary to do it here and then again in the method?

src/components/nodes/boundaryEvent/boundaryEvent.vue
```diff
@@ -76,32 +81,87 @@ export default {
...
     this.setShapeProperties();
     this.shape.addTo(this.graph);
     await this.$nextTick();
-    this.attachBoundaryEventToTask();
+    this.listenForValidDropTargets();
+    this.listenForInvalidDropTargets();
+    this.toggleInterruptingStyle();
+
+    const task = this.getTaskUnderShape();
+    if (task) {
```
JCook21: Could this null checking be done inside `attachBoundaryEventToTask`? It would keep the code that calls this method cleaner and simplify any future changes to what it means for that method to be allowed to be called. You could use early return to bail if a task is not present.

src/components/nodes/boundaryEvent/boundaryEvent.vue
chipit24: > could the naming reflect that?

I was thinking that "listenFor" in the method name hints at this being related to registering an event listener. What naming convention would you suggest? "registerEventListeners"?

I'll look into consolidating the two functions, since they are very related, and I'll see if there's any logic I can pull out into a separate file.

PR #615, opened 2019-09-09
Boundary Events should not accept incoming flows

### LINKED ISSUE #558: Boundary Timer Event should not have incoming sequence flows
Incoming sequence flows are not compatible with Camunda. Currently, you can add incoming sequence flows to boundary timer events and it should not be accepted.

Reference: https://github.com/ProcessMaker/modeler/issues/581.

lisa-fehr:
Closes #558

### DIFF
--- a/src/components/nodes/boundaryErrorEvent/index.js
+++ b/src/components/nodes/boundaryErrorEvent/index.js
@@ -28,9 +28,6 @@ export default {
       }),
     });
   },
-  validateIncoming() {
-    return false;
-  },
   inspectorConfig: [
     {
       name: 'Boundary Event',
--- a/src/components/nodes/boundaryErrorEvent/index.js
+++ b/src/components/nodes/boundaryErrorEvent/index.js
@@ -28,6 +28,9 @@ export default {
       }),
     });
   },
+  validateIncoming() {
+    return false;
+  },
   inspectorConfig: [
     {
       name: 'Boundary Event',
--- a/tests/e2e/specs/BoundaryTimerEvent.spec.js
+++ b/tests/e2e/specs/BoundaryTimerEvent.spec.js
@@ -232,7 +232,7 @@ describe('Boundary Timer Event', () => {
     getGraphElements().should('have.length', numberOfElementsAfterAddingTasksAndBoundaryEvents);
   });
 
-  it.skip('it can toggle interrupting on Boundary Timer Events', function() {
+  it.skip('can toggle interrupting on Boundary Timer Events', function() {
     const taskPosition = { x: 200, y: 200 };
     dragFromSourceToDest(nodeTypes.task, taskPosition);
 
@@ -256,7 +256,7 @@ describe('Boundary Timer Event', () => {
 
   });
 
-  it.skip('it retains outgoing sequence flows on Boundary Timer Events', function() {
+  it.skip('retains outgoing sequence flows on Boundary Timer Events', function() {
 
     const taskForTimerPosition = { x: 200, y: 200 };
     dragFromSourceToDest(nodeTypes.task, taskForTimerPosition);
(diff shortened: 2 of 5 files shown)


tests/e2e/specs/BoundaryTimerEvent.spec.js
```diff
@@ -285,7 +285,36 @@ describe('Boundary Timer Event', () => {
...
+
+    const taskPosition = { x: 400, y: 300 };
+    dragFromSourceToDest(nodeTypes.task, taskPosition);
+
+    const boundaryEvents = [
+      nodeTypes.boundaryEvent,
+      nodeTypes.boundaryTimerEvent,
+      // @todo enable when added to menu
+      //nodeTypes.boundaryErrorEvent,
+      //nodeTypes.boundarySignalEvent,
```
JCook21: Can we remove these commented out lines? If we need them they'll still be in version control history. Right now they're just noise and are likely to just end up as cruft.

lisa-fehr: If the commit is squashed, would there be a history? Should it be an issue to check these later?

src/components/nodes/boundaryErrorEvent/index.js
```diff
@@ -28,6 +28,9 @@ export default {
       }),
     });
   },
+  validateIncoming() {
+    return false;
+  },
```
chipit24: Going with what @JCook21 said: The goal here is to have the other boundary events extend the base boundary event. So this method could be added only in `/src/components/nodes/boundaryEvent/index.js`, and, when we enable the other boundary events, they should be updated to extend that common base config.

tests/e2e/specs/BoundaryTimerEvent.spec.js
```diff
@@ -285,7 +285,32 @@ describe('Boundary Timer Event', () => {
 
   });
 
-  it('it cannot attach an incoming sequence flow on Boundary Timer Events');
+  it('it cannot attach an incoming sequence flow on Boundary Events', function() {
```
chipit24: I don't think we need to double up on the `it` here. We've been following a naming convention that reads as if the text passed to it follows from that word.

```suggestion
  it('cannot attach an incoming sequence flow on Boundary Events', function() {
```

tests/e2e/specs/BoundaryTimerEvent.spec.js
```diff
@@ -285,7 +285,32 @@ describe('Boundary Timer Event', () => {
...
+    ];
+
+    boundaryEvents.forEach((event) => {
+      const eventPosition = { x: 300, y: 210 };
+      dragFromSourceToDest(event, eventPosition);
+
+      connectNodesWithFlow('sequence-flow-button', taskPosition, eventPosition);
+
+      getElementAtPosition(taskPosition).then(getLinksConnectedToElement).should($links => {
+        expect($links.length).to.eq(0);
```
chipit24: This can be updated to produce a more readable assertion:

```suggestion
        expect($links).to.have.lengthOf(0);
```

tests/e2e/specs/BoundaryTimerEvent.spec.js
```diff
@@ -285,7 +285,32 @@ describe('Boundary Timer Event', () => {
...
+      });
+
+      getElementAtPosition(eventPosition).click().then($endEvent => {
+        getCrownButtonForElement($endEvent, 'delete-button').click();
```
chipit24: The element being worked on here isn't an "end event". I think `$endEvent` should be re-named to something like `$boundaryEvent`.

lisa-fehr: Done. Thanks for the suggestions.

Issue #220, opened 2019-02-12
Message Flows


velkymx:
https://knowhow.visual-paradigm.com/business-process-modeling/bpmn-message-flows/

https://www.researchgate.net/figure/BPMN-20-Messages-and-Message-Flow_fig5_283500761

https://client.igrafx.com/17/en/Content/igrafx/usingbpmn/Connector_Lines_in_BPMN_Diagrams_Reference.htm

https://www.lucidchart.com/pages/bpmn-bpmn-2.0-tutorial?a=0

https://www.visual-paradigm.com/support/documents/vpuserguide/2821/286/56998_sequenceandm.html

https://camunda.com/bpmn/reference/#sequenceflow

<img width="511" alt="screenshot 2019-02-12 08 39 17" src="https://user-images.githubusercontent.com/535300/52651797-b1da5900-2ea1-11e9-8578-f0c7900badbc.png">

- [x] Add Message Flow to crown
- [x] Connect elements with dashed line with outlined arrowhead
- [ ] Update BPMN with specifics
- [ ] Show label next to the line (like the regular flows)

chipit24:
From the BPMN spec:

<img width="783" alt="screen shot 2019-02-12 at 11 19 32 am" src="https://user-images.githubusercontent.com/7561061/52650382-27512400-2eb8-11e9-86a3-bef20a88563b.png">

Connection rules:
* All Message Flows MUST connect two separate Pools. They MAY connect to the Pool boundary or to Flow Objects within the Pool boundary. They MUST NOT connect two objects within the same Pool.
* Activity: 0 or more incoming or outgoing Message Flows.
* Start Event: 0 or more incoming Message Flows; no outgoing Message Flows.
* End Event: no incoming Message Flows; 0 or more outgoing Message Flows.

PR #616, opened 2019-09-09
Allow messages flows to connect to other elements in other pools

### LINKED ISSUE #611: Message Flows
You should be able to connect a message flow from an empty pool to a message catch event or other task types inside a process inside another pool. It is currently not working.

tuantruong625:
The purpose of this PR is to allow pool message flows to connect to elements in different pools.

**How to test**
- Drag two pool onto paper
- Drag a task onto one pool
- Connect with message flow from pool B to task in Pool A

**Behaviour** 
- Elements are connected with a message flow

**Note**
- There is a layering issue with message flows

Fixes #611

### DIFF
--- a/src/components/nodes/messageFlow/messageFlow.vue
+++ b/src/components/nodes/messageFlow/messageFlow.vue
@@ -38,7 +38,6 @@ export default {
       return this.hasTargetType() &&
         this.targetIsNotALane() &&
         this.targetIsNotContainingPool() &&
-        this.targetIsInDifferentPool() &&
         this.targetIsNotSource() &&
         this.validateOutgoing();
     },
@@ -56,20 +55,6 @@ export default {
     },
     targetIsPool() {
       return this.targetType === poolId;
-    },
-    targetIsInDifferentPool() {
-      if (this.targetIsPool()) {
-        return true;
-      }
-
-      if (this.targetIsIntermediateCatchEvent()) {
-        return true;
-      }
-
-      const targetPool = this.target.component.node.pool;
-      const sourcePool = this.sourceShape.component.node.pool;
-
-      return sourcePool != null && sourcePool !== targetPool;
     },
     targetIsNotSource() {
       return this.targetNode.definition.id !== this.sourceNode.definition.id;
--- a/tests/e2e/specs/MessageFlows.spec.js
+++ b/tests/e2e/specs/MessageFlows.spec.js
@@ -54,4 +54,25 @@ describe('Message Flows', () => {
         expect($links.length).to.eq(numberOfMessageFlowsAdded);
       });
   });
+
+  it('Can connect to elements in different pools', () => {
+    const pool1Position = { x: 250, y: 250 };
+    dragFromSourceToDest(nodeTypes.pool, pool1Position);
+
+    const pool2Position = { x: 250, y: 500 };
+    dragFromSourceToDest(nodeTypes.pool, pool2Position);
+
+    const offset = 100;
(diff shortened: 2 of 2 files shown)


src/components/nodes/messageFlow/messageFlow.vue
```diff
@@ -57,20 +56,6 @@ export default {
...
-
-      if (this.targetIsIntermediateCatchEvent()) {
-        return true;
-      }
-
-      const targetPool = this.target.component.node.pool;
-      const sourcePool = this.sourceShape.component.node.pool;
-
-      return sourcePool != null && sourcePool !== targetPool;
-    },
```
chipit24: I don't think these validation rules should be removed. See this comment regarding connection rules: https://github.com/ProcessMaker/modeler/issues/220#issuecomment-462840889.

Looking at the issue you linked to, you should only be adding the following new behaviour (emphasis mine):
> connect a message flow **from an empty pool** to a **message catch event** or **other task types** inside a process inside another pool

Although removing this now allows the connection specified above, it also now allows previously invalid connection. Check out page 42 (page 72 using the page input): https://www.omg.org/spec/BPMN/2.0/PDF/. It shows a list of valid  connection rules for message flows.

src/components/nodes/messageFlow/messageFlow.vue
```diff
@@ -45,6 +45,15 @@ export default {
     targetIsIntermediateCatchEvent(){
       return this.targetNode.definition.$type === 'bpmn:IntermediateCatchEvent';
     },
+    targetTaskTypes(){
```
JCook21: I think this is a bad name. `targetTaskType` seems to suggest this is some sort of getter while this is a query method, returns an answer to a question. I would combine this method with `targetIsIntermediateCatchEvent` since the only place that is called is in line 74 below. This could then read as follows:
`
isValidTarget(type)
return [
         'bpmn:Task', 
         'bpmn:ScriptTask', 
         'bpmn:ManualTask',
         'bpmn:CallActivity',
         'bpmn:ServiceTask',
         'bpmn:IntermediateCatchEvent',
       ].includes(type);
}
`
I would also prefer it if type was passed as an argument since this will make this code less coupled and help with refactoring later.

src/components/nodes/messageFlow/messageFlow.vue
```diff
@@ -45,6 +45,15 @@ export default {
     targetIsIntermediateCatchEvent(){
```
JCook21: If you agree with the comment below, this method can be removed.

src/components/nodes/messageFlow/messageFlow.vue
```diff
@@ -62,7 +71,7 @@ export default {
         return true;
       }
 
-      if (this.targetIsIntermediateCatchEvent()) {
+      if (this.targetIsIntermediateCatchEvent() || this.targetTaskTypes()) {
```
JCook21: If you implement the comment above this can be change to:
```suggestion
      if (this.isValidTarget(this.targetNode.definition.$type)) {
```

src/components/nodes/messageFlow/messageFlow.vue
```diff
@@ -58,11 +65,13 @@ export default {
       return this.targetType === poolId;
     },
     targetIsInDifferentPool() {
+      const targetNodeType = this.targetNode.definition.$type;
```
JCook21: I like this extraction but do you think the definition of the variable should be closer to where it's used? How about having this on line 73?

tuantruong625: That's a good point.

PR #618, opened 2019-09-10
Clicking on boundary event snaps it to another position

### LINKED ISSUE #617: Clicking Boundary Event snaps it to another port/connection
**Expected**
Boundary Event should stay in place when selecting it.

**Actual**
Boundary Event may move up or to the left every time it is clicked.

Example:
https://www.dropbox.com/s/xtj0jhznlphwuy9/clicking-moves-event.mov

lisa-fehr:
Closes #617

Snapping on drag also seems to make it less janky when trying to get it to snap to the right point.
https://www.dropbox.com/s/lh2q7gbuzkcb6bp/dragToSnap.mov

### DIFF
--- a/src/components/nodes/boundaryEvent/boundaryEvent.vue
+++ b/src/components/nodes/boundaryEvent/boundaryEvent.vue
@@ -88,7 +88,7 @@ export default {
       this.toggleInterruptingStyle(this.node.definition.cancelActivity);
       this.updateShapePosition(task);
 
-      this.shape.listenTo(this.paper, 'element:pointerup', cellView => {
+      this.shape.listenTo(this.paper, 'element:pointermove', cellView => {
         if (cellView.model !== this.shape) {
           return;
         }


lisa-fehr:
Any other options? This branch seems to introduce a bug. If you add a point on a connected sequence flow, the sequence flow size becomes large.

https://www.dropbox.com/s/6xickzh07ab1eh0/boundaryFlowError.mov

chipit24:
@lisa-fehr This works much better than before. Not sure what's causing that bug with the sequence flow 🤔 I tried a few different things, to no avail:
 * Commenting out `updateCrownPosition`.
 * Ensuring `updateShapePosition` only runs when a new anchor is returned.
 * Passing in position from the event instead of getting it from shape.
 * Wrapping `updateShapePosition` with calls to `freeze` / `unfreeze`.

Perhaps this is a tradeoff we can discuss with Alan? Improved boundary event dragging functionality at the expensive of introducing a subtle bug that we can break out and work on after?

@JCook21 @tuantruong625 thoughts?

lisa-fehr:
@chipit24 @JCook21 @tuantruong625 I'm wondering if it happens in linkConfig when it updates waypoints. So I might try something with that first.

lisa-fehr:
I would vote that the boundary event snapping to another position on select is less damage than dragging it and potentially having a sequence flow grow so far off screen you can't get to the crown controls.

JCook21:
> I would vote that the boundary event snapping to another position on select is less damage than dragging it and potentially having a sequence flow grow so far off screen you can't get to the crown controls.

This error seems to occur in crownConfig.js on line 252. It looks like the array of buttons includes the point on the attached sequence flow, and updating that in this instance seems to cause the sequence flow positioning issues.

lisa-fehr:
Trying another solution.

PR #626, opened 2019-09-11
Added specialised parser for boundary timer events.

JCook21:
This PR ensures that boundary timer events are rendered as such when loading from XML. Before, adding a boundary timer event, clicking undo and then redo resulted in it being re-added as a base boundary event. I also did a little refactoring to introduce a constant in the code. PHPStorm also cleaned up and reordered imports.


src/setup/defaultNodes.js
```diff
@@ -45,32 +44,42 @@ const nodeTypes = [
   pool,
   poolLane,
   boundaryEvent,
-  boundaryTimerEvent,
```
lisa-fehr: Won't this need to be put back when boundary timer is used? This removes it from the menu.

PR #636, opened 2019-09-12
Fix element position on parse

### LINKED ISSUE #632: Confirm boundary timer events are in the correct place on upload/save
_As a positive:_ it doesn't seem like boundary timer events keep changing position if the xml is downloaded and uploaded multiple times.

**Expected**
On upload (or save in processmaker) the Boundary Timer Event is in the last seen position.

**Actual**
Boundary Timer events seem to connected to a different port than originally saved on.

Original 1:
Screen Shot 2019-09-12 at 12 01 16 PM

Downloaded then uploaded 1:
Screen Shot 2019-09-12 at 12 01 37 PM

Original 2: 
Screen Shot 2019-09-12 at 12 04 00 PM

Downloaded then uploaded 2:
Screen Shot 2019-09-12 at 12 04 14 PM

In processmaker after saving:
https://www.dropbox.com/s/640xsx9gzw5v46w/bounadryEventPlacementSaved.mov?dl=0

chipit24:
Fixes #632.

**Note:**

To fix the issue, I used the centre of the shape as a reference for finding the closest port to snap to. But, when you drag a shape from the controls list, the shape's _top left_ corner will appear under your cursor, which means that the shape's centre will possibly be closer to a different port.

This means that a user might hover directly over a port, but it may snap to the port closer to the right. This fixes the issue of the boundary event re-appearing in a different location, and it's why I added conditional logic to `Modeler.vue` to ensure it's centre is positioned under the cursor.

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -679,9 +679,20 @@ export default {
       diagram.bounds.x = (clientX - paperOrigin.x) / scale.sx;
       diagram.bounds.y = (clientY - paperOrigin.y) / scale.sy;
 
+      if (this.isBoundaryEvent(definition)) {
+        this.setShapeCenterUnderCursor(diagram);
+      }
+
       // Our BPMN models are updated, now add to our nodes
       // @todo come up with random id
       this.addNode({ type: control.type, definition, diagram });
+    },
+    isBoundaryEvent(definition) {
+      return definition.$type === 'bpmn:BoundaryEvent';
+    },
+    setShapeCenterUnderCursor(diagram) {
+      diagram.bounds.x = diagram.bounds.x - (diagram.bounds.width / 2);
+      diagram.bounds.y = diagram.bounds.y - (diagram.bounds.height / 2);
     },
     addNode({ type, definition, diagram }) {
       /*
--- a/src/components/nodes/boundaryEvent/boundaryEvent.vue
+++ b/src/components/nodes/boundaryEvent/boundaryEvent.vue
@@ -85,13 +85,23 @@ export default {
 
       return x !== prevX || y !== prevY;
     },
+    getCenterPosition() {
+      const { x, y } = this.shape.position();
+      const { width, height } = this.shape.size();
+
+      return {
+        x: x + (width / 2),
+        y: y + (height / 2),
+      };
+    },
     updateShapePosition(task) {
       if (!this.hasPositionChanged()) {
         return;
       }
-      const { x, y } = getBoundaryAnchorPoint(this.shape.position(), task);
-      const { width } = this.shape.size();
-      this.shape.position(x - (width / 2), y - (width / 2));
+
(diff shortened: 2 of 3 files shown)


src/components/Modeler.vue
```diff
@@ -679,10 +679,21 @@ export default {
...
 
+      if (this.isBoundaryEvent(definition)) {
+        this.setShapeCenterUnderCursor(diagram);
+      }
+
       // Our BPMN models are updated, now add to our nodes
       // @todo come up with random id
       this.addNode({ type: control.type, definition, diagram });
     },
+    isBoundaryEvent(definition) {
```
JCook21: Is there anyway we could modify the definition object to have this behaviour there? Do we control that object? If so it would be much preferable to having it here in the modeler. I would imaging the code would look something like:
`
isType(type) {
    return this.$type === type;
}
`

src/components/nodes/boundaryEvent/boundaryEvent.vue
```diff
@@ -85,13 +85,23 @@ export default {
 
       return x !== prevX || y !== prevY;
     },
+    getCenterPosition() {
```
JCook21: Is there any way that this would be added to the shape class? Is that a class that we 'own'?

chipit24: Shape is part of JointJS; it's not something we own. The shape may return a JointJS "rect", in which case I could call `centre` on it: https://resources.jointjs.com/docs/jointjs/v3.0/geometry.html#g.Rect.prototype.center. I'll look into that.

I could also grab the position of the centre port and use that. But I'm not sure I want to couple this to the ports implementation.

src/components/Modeler.vue
chipit24: That would be nice, but alas, we also don't own the definition object either; it comes from `bpmn-moddle`.

JCook21: Maybe we should consider ways to mitigate this then. One option is to wrap the definition object, which would also have the benefit of helping to separate us from the dependency. Another would be to extend it using prototypical inheritance, although that wouldn't help to isolate the dependency. What do you think? This behaviour really belongs in the definition...

chipit24: I'm thinking, we already have a sort of wrapper around these items: `node`. Currently, we treat node as a plain object with at least the following properties: `type`, `definition`, `diagram`.

It may be a good idea to wrap up the node functionality into a `Node` class. This node class could handle both returning its type, and functionality related to returning its centre position.

`definition` and `diagram` both come from `bpmn-moddle`, and both refer to the same entity that is eventually rendered on the paper.

JCook21: @chipit24 this sounds like a good plan! Could we begin this here by moving this behaviour (and the behaviour in my other comment) to the node?

chipit24: Yup :)

chipit24:
**Note:** Refactoring–creating a `Node` class–will be introduced in a separate PR.

PR #650, opened 2019-09-17
Fix mini map scale

### LINKED ISSUE #646: Mini-map: Fix mini-map scale on initial load
The mini-map should scale to contain all visible elements on the grid. If there are multiple elements on the grid, the mini-map should zoom out to show all elements. If there are only a few elements on the grid, the mini-map should zoom in to make it easier to see the elements.

Currently, when the modeler initially loads, it does not scale; a small process will be too zoomed out, while a large process will appear cut-off:

Screen Shot 2019-09-17 at 11 58 28 AM

Screen Shot 2019-09-17 at 11 58 15 AM

chipit24:
Fixes #646.

### DIFF
--- a/src/components/MiniPaper.vue
+++ b/src/components/MiniPaper.vue
@@ -1,0 +1,83 @@
+<template>
+  <div class="mini-paper-container" @click="movePaper">
+    <div v-show="isOpen" ref="miniPaper" class="mini-paper" />
+  </div>
+</template>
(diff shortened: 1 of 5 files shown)


chipit24:
Closing to work around CircleCI bug. See https://discuss.circleci.com/t/could-not-parse-object-git-error/21193/34.

PR #657, opened 2019-09-19
Vue code coverage

JCook21:
Added initial configuration to get code coverage for Vue components.


.babelrc
```diff
@@ -0,0 +1,5 @@
+{
+  "plugins": [
+    "istanbul"
```
JCook21: I'm not sure what effect adding this file will have on the existing babel config. Coverage didn't work until I added this though.

JCook21:
See https://1017-150140391-gh.circle-artifacts.com/0/root/project/coverage/lcov-report/index.html for code coverage including vue components.

.babelrc
chipit24: We already have a `babel.config.js` file which specifies `plugins: ['istanbul']`. Is this file required?

JCook21: All I know is that it didn't work until I added this...

chipit24: Interesting. It could be confusing for new developers coming to the project to see two babel config files with a duplicated plugins config. I'm assuming that whatever generates the Vue code coverage is unable to read the `babel.config.js` file, but is able to read the `.babelrc` file.

Looking at the babel docs, at https://babeljs.io/docs/en/configuration#babelconfigjs, it has `babel.config.js` set up to export a function. If you change it to use that format, does it work?

lisa-fehr:
Should it be doing code coverage on the tests folder?

.babelrc
JCook21: Thanks, changing to that format seems to work.

JCook21:
> Should it be doing code coverage on the tests folder?

@lisa-fehr no it shouldn't but I think it's been like this since it was setup. I'll see if there's a way to ignore this directory.

JCook21:
> > Should it be doing code coverage on the tests folder?
> 
> @lisa-fehr no it shouldn't but I think it's been like this since it was setup. I'll see if there's a way to ignore this directory.

Hopefully this is now resolved.

PR #666, opened 2019-09-20
Create ModelerNode class

chipit24:
This PR creates a `ModelerNode` class to abstract all node/definition/diagram related functionality. This refactor is a result of the discussion in https://github.com/ProcessMaker/modeler/pull/636.

### DIFF
--- a/src/ModelerNode.js
+++ b/src/ModelerNode.js
@@ -1,0 +1,60 @@
+import { g } from 'jointjs';
+
+export default class ModelerNode {
+  constructor(type, nodeRegistry, moddle, $t) {
+    this.type = type;
+    this.definition = nodeRegistry[type].definition(moddle, $t);
+    this.diagram = nodeRegistry[type].diagram(moddle);
+  }
+
+  isType(bpmnType) {
+    return this.bpmnType === bpmnType;
+  }
+
+  get bpmnType() {
+    return this.definition.$type;
+  }
+
+  getSize() {
+    return {
+      width: this.diagram.bounds.width,
+      height: this.diagram.bounds.height,
+    };
+  }
+
+  setPosition({ x, y }) {
+    this.diagram.bounds.x = x;
+    this.diagram.bounds.y = y;
+  }
+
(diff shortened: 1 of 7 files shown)


src/ModelerNode.js
```diff
@@ -0,0 +1,60 @@
...
+
+  isType(bpmnType) {
+    return this.bpmnType === bpmnType;
+  }
+
+  get bpmnType() {
+    return this.definition.$type;
+  }
+
+  getSize() {
```
JCook21: I think this could be a Javascript get function.

src/ModelerNode.js
```diff
@@ -0,0 +1,60 @@
...
+    return {
+      width: this.diagram.bounds.width,
+      height: this.diagram.bounds.height,
+    };
+  }
+
+  setPosition({ x, y }) {
```
JCook21: I think both set and get position should use Javascript getter/setter functions.

src/ModelerNode.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  getPosition() {
+    return {
+      x: this.diagram.bounds.x,
+      y: this.diagram.bounds.y,
+    };
+  }
+
+  getCenterPosition() {
```
JCook21: For consistency, I think this should be a javascript getter function to.

src/ModelerNode.js
```diff
@@ -0,0 +1,60 @@
...
+    const { x, y } = this.getPosition();
+    const { width, height } = this.getSize();
+    const rectangle = new g.Rect(x, y, width, height);
+
+    return rectangle.center();
+  }
+
+  setCenterOnPoint({ x, y }) {
```
JCook21: I think this is the only function that I would leave as is, i.e. **not** as a javascript setter. The reason is that this method is carrying out action rather than just being a bare getter and setter.

JCook21:
Closing as so much time has elapsed that it would be difficult to bring this up to date.

PR #672, opened 2019-09-23
Add error boundary event

### LINKED ISSUE #659: Add Error Boundary Event
The error boundary event should look like this:
Screen Shot 2019-09-19 at 2 26 07 PM

```xml
<boundaryEvent id="some_id" attachedToRef="some_task">
  <errorEventDefinition />
</boundaryEvent>
```

Tasks:
 - [ ] Create a separate boundary event component extending the base boundary event.
    * Ensure the correct icon is loaded: `/src/assets/boundary-error-event-icon.svg`.
    * Ensure the BPMN XML only includes `errorEventDefinition` without a `cancelActivity` attribute.
    * Overwrite default config properties with any boundary event specific config, that we know of at this time.
    * Ensure the error boundary event is available in the list of nodes and is correctly parsed on upload or undo/redo (add to `defaultNodes.js`).
    * Out of scope for now: errorRef property of the event.
    * Automated test:
        * Drag event type onto valid drop target.
        * Download XML.
        * Assert that the XML contains the snippet required from the BPMN docs.

chipit24:
Fixes #659.

Note: This PR is dependant on https://github.com/ProcessMaker/modeler/pull/697, and should be merged after that PR is merged into this one.

### DIFF
--- a/src/components/nodes/boundaryEvent/index.js
+++ b/src/components/nodes/boundaryEvent/index.js
@@ -51,14 +51,6 @@ export default {
                 helper: 'The Name of the Boundary Event',
               },
             },
-            {
-              component: 'FormCheckbox',
-              config: {
-                label: 'Interrupting',
-                name: 'cancelActivity',
-                helper: 'Boundary Event Type',
-              },
-            },
           ],
         },
       ],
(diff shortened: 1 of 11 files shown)


src/components/nodes/boundaryEvent/index.js
```diff
@@ -51,14 +51,6 @@ export default {
...
               },
             },
-            {
-              component: 'FormCheckbox',
-              config: {
-                label: 'Interrupting',
-                name: 'cancelActivity',
-                helper: 'Boundary Event Type',
-              },
-            },
```
JCook21: Good catch on moving this into boundary timer events!

PR #681, opened 2019-09-24
Refactor tests

JCook21:
Moved tests that are common to all boundary events to their own test file and refactored these in preparation for more boundary event types.

### DIFF
--- a/tests/e2e/specs/BoundaryEventCommonBehaviour.spec.js
+++ b/tests/e2e/specs/BoundaryEventCommonBehaviour.spec.js
@@ -1,0 +1,276 @@
+import {
+  connectNodesWithFlow,
+  dragFromSourceToDest,
+  getComponentsEmbeddedInShape,
+  getCrownButtonForElement,
+  getElementAtPosition,
+  getLinksConnectedToElement,
+  moveElementRelativeTo,
+  removeIndentationAndLinebreaks,
+  waitToRenderAllShapes,
+} from '../support/utils';
+import { nodeTypes } from '../support/constants';
+
+const boundaryEventSelector = '.main-paper ' +
+  '[data-type="processmaker.components.nodes.task.Shape"] + ' +
+  '[data-type="processmaker.components.nodes.boundaryEvent.Shape"]';
+
+describe('Boundary Event Common Behaviour', () => {
+  beforeEach(() => {
+    cy.loadModeler();
+  });
+
+  it('removes references of itself when inside of a pool and deleting the pool', function() {
+    if (Cypress.env('inProcessmaker')) {
+      this.skip();
+    }
+
+    const startEventPosition = { x: 150, y: 150 };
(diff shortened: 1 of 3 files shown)


tests/e2e/specs/BoundaryEventCommonBehaviour.spec.js
```diff
@@ -0,0 +1,241 @@
...
+} from '../support/utils';
+import { nodeTypes } from '../support/constants';
+
+const boundaryEventSelector = '.main-paper ' +
+  '[data-type="processmaker.components.nodes.task.Shape"] + ' +
+  '[data-type="processmaker.components.nodes.boundaryEvent.Shape"]';
+const boundaryEventPosition = { x: 250, y: 250 };
+const taskPosition = { x: 250, y: 200 };
+
+const boundaryEventData = [
```
tuantruong625: Could `boundaryEventSelector` be moved to constants.js?

tests/e2e/specs/BoundaryTimerEvent.spec.js
```diff
@@ -18,8 +14,8 @@ import { nodeTypes } from '../support/constants';
 import { defaultNodeColor, invalidNodeColor } from '../../../src/components/nodeColors';
 
 const boundaryEventSelector = '.main-paper ' +
-    '[data-type="processmaker.components.nodes.task.Shape"] + ' +
-    '[data-type="processmaker.components.nodes.boundaryEvent.Shape"]';
+  '[data-type="processmaker.components.nodes.task.Shape"] + ' +
+  '[data-type="processmaker.components.nodes.boundaryEvent.Shape"]';
 
 describe('Boundary Timer Event', () => {
   beforeEach(() => {
```
chipit24: I think the _Render a boundary timer event_ test below can be added to the group of common tests as well, since the `boundaryTimerEventXML` it tests is the same as the one you provided for `eventXMLSnippet` in `BoundaryEventCommonBehaviour.spec.js`.

JCook21: @chipit24 this test has now been moved too.

JCook21:
> This is great! 🔥 I just left one small suggestion.
> 
> Something to also keep in mind, is that boundary events may not all have the same valid targets, but your tests assume that they can all attach to a task.
> 
> Currently, it seems the escalation boundary event can only attach to call activities. I'm not sure if you also want to define "valid targets" as part of `boundaryEventData`, or this can be refactored later. If you do it now, then the following two tests can also be pulled into the common tests file:
> 
> * Can drag boundary timer events over valid targets, but not over invalid targets
> * turns target red when it is an invalid drop target, and snaps back to original position
> 
> And then maybe the "_moves to another task when dragged over_" tests can be made into a generic "_moves to another valid drop target when dragged over_".

I think this is a great point but let's deal with this when we have tests for the multiple different types of boundary event.

PR #682, opened 2019-09-24
Keep tests DRY

JCook21:
Moved test setup that was duplicated in every file into one place. Also removed unneeded and duplicate checks to skip a test.

### DIFF
--- a/tests/e2e/specs/AssociationFlows.spec.js
+++ b/tests/e2e/specs/AssociationFlows.spec.js
@@ -1,24 +1,20 @@
 import {
+  connectNodesWithFlow,
   dragFromSourceToDest,
   getElementAtPosition,
   getLinksConnectedToElement,
-  connectNodesWithFlow,
 } from '../support/utils';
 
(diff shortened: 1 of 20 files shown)


tests/e2e/specs/AssociationFlows.spec.js
```diff
@@ -1,24 +1,20 @@
...
-
   it('Change direction of association to none, one and both', () => {
     const directionSelectSelector = '[name=associationDirection]';
     const testDirection = {
-      none:`${ direction.none }`,
-      one: `${ direction.one }`,
-      both:`${ direction.both }`,
+      none: `${direction.none}`,
+      one: `${direction.one}`,
+      both: `${direction.both}`,
```
chipit24: Since you're touching this part of the code, it seems these don't need to be wrapped in template strings. It can be:

```js
const testDirection = {
  none: direction.none,
  one: direction.one,
  both: direction.both,
};
```

But after doing that, it's easy to see that `testDirection` is exactly the same as `direction`; so `testDirection` can be removed and `direction` can be used directly in the assertions.

Issue #662, opened 2019-09-19
Add Message Boundary Event


chipit24:
The message boundary event should look like this (interrupting on the left, non-interrupting on the right):
<img width="274" alt="Screen Shot 2019-09-19 at 4 51 58 PM" src="https://user-images.githubusercontent.com/7561061/65280175-d06e9580-dafd-11e9-8385-0ab0fa1f78b4.png">

And produce the following BPMN XML:
```xml
<!-- Interrupting  -->
<bpmn:boundaryEvent id="boundary_event" attachedToRef="task">
  <bpmn:messageEventDefinition />
</bpmn:boundaryEvent>

<!-- Non-interrupting -->
<bpmn:boundaryEvent id="boundary_event" cancelActivity="false" attachedToRef="task">
  <bpmn:messageEventDefinition />
</bpmn:boundaryEvent>
```

Tasks:
 - [x] Create a separate boundary event component extending the base boundary event.
 - [x] Ensure the correct icon is loaded: `/src/assets/toolpanel/intermediate-mail-event.svg`.
 - [x] Ensure the BPMN XML only includes `messageEventDefinition` with a `cancelActivity` attribute if the _Interrupting_ checkbox in the inspector config is unchecked.
 - [ ] Overwrite default config properties with any boundary event specific config. Pending: waiting for more information.
 - [x] Ensure the message boundary event is available in the list of nodes and is correctly parsed on upload or undo/redo (add to `defaultNodes.js`).
- [x] Ensure the message boundary event can only be dropped onto a call activity.
 - ~~Out of scope: messageRef property of the event.~~
 - [ ] Automated test:
 - [ ] Drag event type onto valid drop target.
 - [ ] Download XML.
 - [ ] Assert that the XML contains the snippet required from the BPMN docs.

PR #684, opened 2019-09-24
662 add message boundary event

sanjacornelius:
This PR adds the base implementation for the Boundary Message Events.

references #662


src/setup/defaultNodes.js
```diff
@@ -42,6 +43,7 @@ const nodeTypes = [
   association,
   pool,
   poolLane,
+  boundaryMessageEvent,
```
chipit24: Remove this from here and place it in the `timerEventNodes` array.

Elements in `nodeTypes` are parsed based on their BPMN type, however, the BPMN type for all the different boundary events is `bpmn:BoundaryEvent`, so the parsing of boundary events was  configured to handle this. The name, `timerEventNodes`, isn't really accurate anymore, maybe you can re-name it to something like `timerAndBoundaryEventNodes.`

src/components/nodes/boundaryMessageEvent/boundaryMessageEvent.vue
```diff
@@ -0,0 +1,29 @@
...
+        fill: '#fff',
+      },
+      image: {
+        width: 20,
+        height: 20,
+        y: 3,
+        x: 3,
+      },
+    });
+  },
```
chipit24: The attributes set in `this.shape.attr` appear to be the same ones used in `intermediateMessageCatchEvent.vue`; do you think you can extract them out into a separate file and use it in both places, to help eliminate duplication?

src/setup/defaultNodes.js
```diff
@@ -43,11 +44,12 @@ const nodeTypes = [
   pool,
   poolLane,
 ];
-const timerEventNodes = [
+const timerAndBoundaryEventNodes = [
   [startTimerEvent, 'bpmn:StartEvent', 'bpmn:TimerEventDefinition'],
   [intermediateTimerEvent, 'bpmn:IntermediateCatchEvent', 'bpmn:TimerEventDefinition'],
   [intermediateMessageCatchEvent, 'bpmn:IntermediateCatchEvent', 'bpmn:MessageEventDefinition'],
   [boundaryTimerEvent, 'bpmn:BoundaryEvent', 'bpmn:TimerEventDefinition'],
+  [boundaryMessageEvent, 'bpmn:BoundaryEvent'],
```
chipit24: @sanjacornelius You're missing just one more parameter here:
```suggestion
  [boundaryMessageEvent, 'bpmn:BoundaryEvent', 'bpmn:MessageEventDefinition'],
```

sanjacornelius: Oops! Thanks!

src/components/nodes/boundaryMessageEvent/index.js
```diff
@@ -0,0 +1,31 @@
+import component from './boundaryMessageEvent.vue';
+import boundaryEventConfig from '../boundaryEvent';
+import merge from 'lodash/merge';
+import cloneDeep from 'lodash/cloneDeep';
+
+boundaryEventConfig.inspectorConfig[0].items[0].items.push(
```
lisa-fehr: @chipit24 Might need your input on the best way to get the interrupting checkbox back into the inspector.

chipit24: Calling `push` here mutates the base boundary event config and causes all boundary events to have the interrupting checkbox.

You can look at how it's done in `/src/components/nodes/boundaryTimerEvent/index.js`; it's a little odd with the empty placeholder objects, but it works for now.

I like using `push` though. So instead of exporting the default config object below, assign it to a variable, call `push` on that, and export that. (The `cloneDeep` ensures we're mutating a copy of the config and not the actual base config).

JCook21:
@sanjacornelius I'm going to close this for now. @velkymx just gave us a PRD for message events yesterday and I think we should review this in that light. I don't think this work is useless now: I just think we should review everything before proceeding. Can we all meet on Tuesday (Monday is Thanksgiving here) to plot out the next steps on this?

PR #686, opened 2019-09-25
Add config for Jest and combined test coverage

JCook21:
PR to add configuration for Jest tests and combined test coverage with Cypress.


package.json
```diff
@@ -4,12 +4,22 @@
...
-    "build-bundle": "vue-cli-service build --target lib --name modeler ./src/components/nodes/index.js",
+    "open-cypress": "TZ=UTC cypress open",
     "lint": "vue-cli-service lint --no-fix",
+    "build-bundle": "vue-cli-service build --target lib --name modeler ./src/components/nodes/index.js",
     "lint-fix": "vue-cli-service lint",
-    "test": "TZ=UTC cypress open",
-    "test-processmaker": "cypress open --env inProcessmaker=true",
-    "test-ci": "TZ=UTC vue-cli-service test:e2e --mode test --headless --browser chrome"
+    "pretest": "rm -rf .nyc_output || true",
+    "test": "npm run test-unit && npm run test-ci",
```
lisa-fehr: Oh, the readme does say this. It may need to be updated as well.

```
# Run the Cypress end-to-end (e2e) test suite
npm test

# Run the Cypress end-to-end (e2e) test suite in headless mode
npm run test-ci
```

JCook21: Good catch! I'll fix it right now.

jest.config.js
```diff
@@ -0,0 +1,33 @@
...
+    '/node_modules/',
+  ],
+  moduleNameMapper: {
+    '^@/(.*)$': '<rootDir>/src/$1',
+  },
+  snapshotSerializers: [
+    'jest-serializer-vue',
+  ],
+  testMatch: [
+    '**/tests/unit/**/*.spec.(js|jsx|ts|tsx)|**/__tests__/*.(js|jsx|ts|tsx)',
```
chipit24: We don't have a `__tests__` folder so I think we can remove the `|` condition in here.

jest.config.js
```diff
@@ -0,0 +1,33 @@
...
+  ],
+  testURL: 'http://localhost/',
```
chipit24: This appears to be the default value, so I don't think we need to explicitly set it, especially since we're not currently relying on it for our tests, and the URL is usually `http://localhost:8080` anyway.

.eslintrc.js
```diff
@@ -24,7 +27,19 @@ module.exports = {
     'vue/html-self-closing': 'error',
     'vue/mustache-interpolation-spacing': 'error',
   },
+
   parserOptions: {
     parser: 'babel-eslint',
   },
+
+  overrides: [
```
chipit24: I'm not sure we need this section, since we already have ESLint files in our test folders (`tests/unit/.eslintrc.js`), and this references a non-existent `__tests__` folder.

JCook21: I'm not sure if it's needed or not. To be safe, I updated it to match the unit tests directory and we'll see what happens.

PR #692, opened 2019-09-25
Move like controls together

lisa-fehr:
This pr addresses processmaker: https://github.com/ProcessMaker/processmaker/pull/2412

Made control order the same as processmaker. ~Checking if the order needs to be changed on the pr in processmaker.~ Based on our standup, this might be a good start.


JCook21:
@lisa-fehr I think this change looks good. What do we need to do to remove the WIP label on it?

PR #697, opened 2019-09-26
Add validation for dragging error boundary events

### LINKED ISSUE #679: Only allow one error boundary event per task
Has to be done after https://github.com/ProcessMaker/modeler/issues/659.

As per the PRD, [PM4-ErrorBoundaryEvent-230919-2108.pdf](https://github.com/ProcessMaker/modeler/files/3648032/PM4-ErrorBoundaryEvent-230919-2108.pdf), only one error boundary event can exist per task.

**Question:**
Should the validation around dragging an error boundary event on a task which already has one be the same as dragging it over any other invalid target? I.e. The cursor will be a "no-drop" cursor, and the boundary event won't render when dropped.

chipit24:
Fixes #679.

Note: This PR is being merged into https://github.com/ProcessMaker/modeler/pull/672.

Todo:
 - [x] Add validation when dragging from controls list into task
 - [x] Add validation when dragging between tasks
 - [x] Add tests

### DIFF
--- a/src/components/nodes/boundaryErrorEvent/boundaryErrorEvent.vue
+++ b/src/components/nodes/boundaryErrorEvent/boundaryErrorEvent.vue
@@ -5,9 +5,25 @@
 <script>
 import BoundaryEvent from '@/components/nodes/boundaryEvent/boundaryEvent';
 import errorBoltIcon from '@/assets/boundary-error-event-icon.svg';
+import validBoundaryEventTargets from '@/components/nodes/boundaryEvent/validBoundaryEventTargets';
+import { getAttachedErrorBoundaryEvents } from '@/targetValidationUtils';
 
 export default {
   extends: BoundaryEvent,
+  methods: {
+    doesNotHaveOtherBoundaryEvents(model) {
+      const boundaryEvents = getAttachedErrorBoundaryEvents(model).filter(boundaryEvent => {
+        return boundaryEvent !== this.shape;
+      });
+
+      return boundaryEvents.length === 0;
+    },
+    isValidBoundaryEventTarget(model) {
+      return model.component &&
+        validBoundaryEventTargets.includes(model.component.node.definition.$type) &&
+        this.doesNotHaveOtherBoundaryEvents(model);
+    },
+  },
   mounted() {
     this.shape.attr('image/xlink:href', errorBoltIcon);
   },
--- a/src/components/nodes/boundaryErrorEvent/boundaryErrorEvent.vue
+++ b/src/components/nodes/boundaryErrorEvent/boundaryErrorEvent.vue
@@ -12,7 +12,7 @@ export default {
   extends: BoundaryEvent,
   methods: {
     doesNotHaveOtherBoundaryEvents(model) {
-      getAttachedErrorBoundaryEvents(model)
+      return getAttachedErrorBoundaryEvents(model)
         .filter(boundaryEvent => {
           return boundaryEvent !== this.shape;
         })
--- a/src/components/nodes/boundaryErrorEvent/boundaryErrorEvent.vue
+++ b/src/components/nodes/boundaryErrorEvent/boundaryErrorEvent.vue
@@ -12,11 +12,11 @@ export default {
   extends: BoundaryEvent,
   methods: {
     doesNotHaveOtherBoundaryEvents(model) {
-      const boundaryEvents = getAttachedErrorBoundaryEvents(model).filter(boundaryEvent => {
-        return boundaryEvent !== this.shape;
-      });
-
-      return boundaryEvents.length === 0;
+      getAttachedErrorBoundaryEvents(model)
+        .filter(boundaryEvent => {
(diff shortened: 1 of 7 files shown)


chipit24:
Interesting, the single test that is failing on CI (`BoundaryTimerEvent.spec.js` - _turns target red when it is an invalid drop target, and snaps back to original position_), is passing when I run the tests locally 😕

src/components/nodes/boundaryErrorEvent/boundaryErrorEvent.vue
```diff
@@ -5,9 +5,25 @@
...
 
 export default {
   extends: BoundaryEvent,
+  methods: {
+    doesNotHaveOtherBoundaryEvents(model) {
+      const boundaryEvents = getAttachedErrorBoundaryEvents(model).filter(boundaryEvent => {
+        return boundaryEvent !== this.shape;
+      });
+
+      return boundaryEvents.length === 0;
```
JCook21: This could be reduced to the following:
```suggestion
      return getAttachedBoundaryEvents(model)
          .filter(boundaryEvent => boundaryEvent !== this.shape)
          .length === 0;
```
No need for the intermediate variable.

src/targetValidationUtils.js
```diff
@@ -18,6 +19,11 @@ export default function getValidationProperties(clientX, clientY, control, paper
   }
 
   if (isDraggingBoundaryEvent(control.bpmnType)) {
+    if (isBoundaryErrorEvent(control.type)) {
```
JCook21: This nested if statement is pretty janky. I'm sure there's a better way of expressing this.

tests/e2e/specs/BoundaryErrorEvent.spec.js
```diff
@@ -1,36 +1,97 @@
...
-  removeIndentationAndLinebreaks,
+  getComponentsEmbeddedInShape,
+  getPositionInPaperCoords,
+  waitToRenderAllShapes,
 } from '../support/utils';
 import { nodeTypes } from '../support/constants';
 
-describe.only('Boundary Error Event', () => {
+describe('Boundary Error Event', () => {
   beforeEach(() => {
```
JCook21: I think this `beforeEach()` call is unnecessary with the other changes that we've introduced recently. Can you remove it?

JCook21:
I also wonder if a collection of validation utils functions is the correct approach now. Would it be better to organize these as a base class, extended for each specific boundary event type?

src/targetValidationUtils.js
```diff
@@ -78,3 +84,25 @@ function isOverBoundaryEventTarget(clientX, clientY, paper, graph) {
       return component && validBoundaryEventTargets.includes(component.node.definition.$type);
     });
 }
+
+function isOverValidBoundaryErrorEventTarget(clientX, clientY, paper, graph) {
+  const target = graph
+    .findModelsFromPoint(getLocalMousePosition(clientX, clientY, paper))
+    .find(({ component }) => component && validBoundaryEventTargets.includes(component.node.definition.$type));
+
+  if (!target  || getAttachedErrorBoundaryEvents(target).length > 0) {
```
lisa-fehr: Are there two spaces after `target` or is that just github?

chipit24: Nope, it's an extra space! 😨

src/targetValidationUtils.js
chipit24: This was discussed in person and I will leave it for now. Splitting it out will mean modifying the `isBoundaryErrorEvent` function and introducing some extra complexity that may not be useful at this moment.

Since the error boundary event validation is a special case, I think this is fine for now, though it can be refactored when more boundary event validation is required.

chipit24:
> Other than the nested if statement remaining, this looks ok.

@lisa-fehr I discussed this with @JCook21 in person, see https://github.com/ProcessMaker/modeler/pull/697#discussion_r329589901. Let me know if you still think this should be changed and if you have an idea how to do it; we can go over it.

src/components/nodes/boundaryErrorEvent/boundaryErrorEvent.vue
```diff
@@ -5,9 +5,25 @@
...
+import { getAttachedErrorBoundaryEvents } from '@/targetValidationUtils';
 
 export default {
   extends: BoundaryEvent,
+  methods: {
+    doesNotHaveOtherBoundaryEvents(model) {
+      return getAttachedErrorBoundaryEvents(model)
+        .filter(boundaryEvent => {
+          return boundaryEvent !== this.shape;
+        })
```
JCook21: This can be simplified to:
```suggestion
        .filter(boundaryEvent => boundaryEvent !== this.shape)
```

JCook21: This can be simplified to:
```suggestion
        .filter(boundaryEvent => boundaryEvent !== this.shape)
```

PR #699, opened 2019-09-27
Fix mini map

### LINKED ISSUE #667: Mini-map display not consistent with main view
# Bug report issue
## Description
When viewing a mini-map, sequence flow lines are not drawn and shown in the same way as in the main process. Sequence flows are often overlapping shapes in the mini-diagram.
## Reproduction steps

1. Upload the BPMN XML file associated with this issues.
2. Open the mini map and compare it with the main diagram.

## Expected behaviour
The mini-map layout should match that of the main paper.
## Actual behaviour
Sequence flows on the mini-map are not displayed in the same way as the main paper.
## Software used
Latest version of the modeler in develop, using Google Chrome browser on MacOS.
## Other information
Screen Shot 2019-09-23 at 10 34 46 AM

[bpmnProcess.xml.txt](https://github.com/ProcessMaker/modeler/files/3642641/bpmnProcess.xml.txt)

JCook21:
Resolves #667. Also applied refactoring to move domain logic out of component.

### DIFF
--- a/src/components/MiniPaper.vue
+++ b/src/components/MiniPaper.vue
@@ -1,16 +1,16 @@
 <template>
-  <div class="mini-paper-container" @click="movePaper">
-    <div v-show="isOpen" ref="miniPaper" class="mini-paper" />
+  <div class="mini-paper-container" @click="movePaper" :class="visibilityClass">
+    <div ref="miniPaper" class="mini-paper" />
   </div>
 </template>
 
 <script>
-import { dia } from 'jointjs';
+import MiniMapManager from './miniPaperManager';
 
 export default {
   data() {
     return {
-      miniPaper: null,
+      miniMapManager: null,
     };
   },
   props: {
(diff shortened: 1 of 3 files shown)


lisa-fehr:
I can still manually move elements around until the sequence flows don't match. But this is a big improvement.
Screen Shot 2019-09-27 at 2 14 25 PM

src/components/MiniPaper.vue
```diff
@@ -21,44 +21,37 @@ export default {
...
 
-    this.paper.on('render:done', this.scaleMiniMapToFitContent);
+    this.paper.on('render:done', () => this.miniMapManager.scaleMiniMap());
+    window.ProcessMaker.EventBus.$on('modeler-change', () => this.miniMapManager.scaleMiniMap());
 
-    window.ProcessMaker.EventBus.$on('modeler-change', this.scaleMiniMapToFitContent);
   },
   beforeDestroy() {
-    this.paper.off('render:done', this.scaleMiniMapToFitContent);
+    this.paper.off('render:done', () => this.miniMapManager.scaleMiniMap());
```
chipit24: This use of `off` is based on https://backbonejs.org/#Events-off. Here, you're passing a new arrow function as the `callback` parameter, which will be different from the function passed to `this.paper.on`. Because of this, I'm not sure if JointJS/Backbone will be able to recognize that you're trying to remove the event listener for the `this.miniMapManager.scaleMiniMap` call.

Looking at https://backbonejs.org/#Events-on, it seems like you can pass a context for the event listener and then retain the reference to the function:
```js
this.paper.on('render:done', this.miniMapManager.scaleMiniMap, this);
// ...
this.paper.off('render:done', this.miniMapManager.scaleMiniMap, this);
```
I'm not even really sure we need to worry about removing this event listener though, since this component will mount once and stay loaded until the user leaves the page. But I think cleaning up event listeners like this is good practise.

JCook21: Let me try it and see if it works.

JCook21: All seems good, the change has been made.

src/components/MiniPaper.vue
```diff
@@ -21,44 +21,37 @@ export default {
...
-      model: this.graph,
-      width: 300,
-      height: 200,
-      interactive: false,
-    });
+    this.miniMapManager = MiniMapManager.factory(this.graph, this.$refs.miniPaper);
 
-    this.paper.on('render:done', this.scaleMiniMapToFitContent);
+    this.paper.on('render:done', this.miniMapManager.scaleMiniMap, this);
+    window.ProcessMaker.EventBus.$on('modeler-change', this.miniMapManager.scaleMiniMap());
```
chipit24: `$on` should accept a function, but you're _calling_ the function and so the function's return value will be passed here instead. This may be the cause of the `Error in event handler` console error.

JCook21: Right, of course. Sorry.

PR #700, opened 2019-09-27
Move boundary event to different pool

### LINKED ISSUE #695: Confirm that boundary event on a task is moved to another process
# Bug report issue
## Description
Double check that moving a task with a boundary event from one pool to another is reflected in the xml.
## Reproduction steps
1. add two pools to the screen
2. add a task with a boundary event to one pool
3. select the task and drag it to the other pool
## Expected behaviour
The task and boundary event should both be in the same process/pool
## Actual behaviour
The task is moved to the new process/pool but the boundary event is still listed under the old process/pool
## Software used
Version 76.0.3809.132 (Official Build) (64-bit), macOs Mojave 10.14.6, v 0.17.0

lisa-fehr:
Closes #695

Task was getting moved to the other pool but the boundary event remains in the previous pool.

I moved all embedded elements because checking the type seemed to stop the pool from enveloping an element that was touching a corner of the pool (caused a failing test).

### DIFF
--- a/src/components/nodes/pool/pool.vue
+++ b/src/components/nodes/pool/pool.vue
@@ -101,6 +101,15 @@ export default {
 
       return this.graph.findModelsInArea(area);
     },
+    moveEmbeddedElements(currentElement, toPool) {
+      this.getElementsUnderArea(currentElement).forEach((element) => {
+        if (element.isEmbeddedIn(currentElement)) {
+          const elementDefinition = element.component.node.definition;
+          pull(this.containingProcess.get('flowElements'), elementDefinition);
+          toPool.component.containingProcess.get('flowElements').push(elementDefinition);
+        }
+      });
+    },
     moveElement(element, toPool) {
       const elementDefinition = element.component.node.definition;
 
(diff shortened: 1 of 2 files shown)


src/components/nodes/pool/pool.vue
```diff
@@ -101,6 +101,15 @@ export default {
...
       return this.graph.findModelsInArea(area);
     },
+    moveEmbeddedElements(currentElement, toPool) {
+      this.getElementsUnderArea(currentElement).forEach((element) => {
+        if (element.isEmbeddedIn(currentElement)) {
+          const elementDefinition = element.component.node.definition;
+          pull(this.containingProcess.get('flowElements'), elementDefinition);
+          toPool.component.containingProcess.get('flowElements').push(elementDefinition);
+        }
+      });
```
JCook21: This could be re-written to be a little clearer:
```suggestion
     this.getElementsUnderArea(currentElement)
     .filter(element => element.isEmbeddedIn(currentElement))
     .map(element => element.component.node.definition)
     .forEach((elementDefinition) => {
           pull(this.containingProcess.get('flowElements'), elementDefinition);
           toPool.component.containingProcess.get('flowElements').push(elementDefinition);
       });
```

Issue #693, opened 2019-09-25
Make the inspector and control columns the same width


lisa-fehr:
The solution for #487 doesn't work well with wider columns.

Might have to reduce the size of the mini-map. Or show the mini-map below the toolbar again and make the toolbar white, so people don't try to interact with the graph through it.

![Screen Shot 2019-09-25 at 5 06 25 PM](https://user-images.githubusercontent.com/6653340/65887488-fd2b7400-e36b-11e9-9ab1-38a0affcbed2.png)

lisa-fehr:
Merged in #710

velkymx:
The modeler is not reflecting this change in the latest builds. 

![image](https://user-images.githubusercontent.com/535300/67419235-3331d180-f581-11e9-8b5a-9213e0d6a88c.png)

lisa-fehr:
@velkymx Is that correct? Event the mini-map is in an old location.

I'm trying to reproduce, but not having success.

chipit24:
I'm with @lisa-fehr. Looks like a much older version of modeler from the screenshot. I tested it on the latest processmaker `develop` branch (with nothing linked in) and it looks good.

velkymx:
Looks like an update didn't happen. Closing

PR #710, opened 2019-09-30
Made controls column the same width as the inspector.

lisa-fehr:
#693 
With less space, had to revert mini-map placement.

Making the columns the same width, gives less room to work with the diagram. This will be solved later.

**Before:**
Screen Shot 2019-09-30 at 11 19 55 AM
**After:**
Screen Shot 2019-09-30 at 4 11 40 PM
Screen Shot 2019-09-30 at 2 25 32 PM

### DIFF
--- a/src/ModelerApp.vue
+++ b/src/ModelerApp.vue
@@ -1,5 +1,5 @@
 <template>
-  <b-container id="modeler-app" class="h-100 container position-relative">
+  <b-container id="modeler-app" class="h-100 container position-relative p-0">
 
     <b-card no-body class="h-100">
       <b-card-header class="d-flex align-items-center header">
--- a/src/components/MiniPaper.vue
+++ b/src/components/MiniPaper.vue
@@ -57,8 +57,8 @@ export default {
 </script>
 
 <style lang="scss">
-$mini-paper-container-top-position: 2.5rem;
-$mini-paper-container-right-position: 0;
+$mini-paper-container-top-position: 1rem;
+$mini-paper-container-right-position: 17.5rem;
 
 .mini-paper-container {
   position: absolute;
--- a/src/components/MiniPaper.vue
+++ b/src/components/MiniPaper.vue
@@ -57,8 +57,8 @@ export default {
 </script>
 
 <style lang="scss">
-$mini-paper-container-top-position: 1rem;
-$mini-paper-container-right-position: 17.5rem;
+$mini-paper-container-top-position: 2.5rem;
+$mini-paper-container-right-position: 0;
 
 .mini-paper-container {
   position: absolute;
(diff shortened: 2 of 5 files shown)


lisa-fehr:
@tuantruong625 @JCook21 @chipit24 any other options for the toolbar overlay issue? I heard turning the background white was an option for quick fixing.

JCook21:
@lisa-fehr it's not clear from the description why you had to revert the mini-map placement. Can you add some more context to this?

tuantruong625:
@lisa-fehr The toolbar should have enough space in the processmaker app. If not we could reducing the margins or reducing the map size.

lisa-fehr:
@JCook21 the diagram area became smaller. It overlapped its own button, which could not be clicked again.

JCook21:
> @JCook21 the diagram area became smaller. It overlapped its own button, which could not be clicked again.

Is there any way we can adjust where the mini map opens so that it doesn't overlap the button without moving the button back?

tuantruong625:
@JCook21 We could reduce the left and right margins on the modeler container to give the paper more space.

chipit24:
I agree with @tuantruong625. I loaded this up in the processmaker app and it appears there's plenty of space for the minimap at the moment. The `ModelerApp.vue` component in processmaker has a `p-0` on the container. You could add this class to the `ModelerApp.vue` component in standalone as well.

lisa-fehr:
@tuantruong625 @chipit24 This is what I see with p-0 on the container. I'll see what other padding I can remove.
Screen Shot 2019-09-30 at 1 51 02 PM

lisa-fehr:
@chipit24 @tuantruong625  @JCook21  I couldn't see how I could remove the padding on the minimum supported screen width. So I had to make the mini-map smaller. But I'm not sure about this.

**Current processmaker:**
Screen Shot 2019-09-30 at 2 22 04 PM

**processmaker with alterations:**
Screen Shot 2019-09-30 at 2 25 32 PM

chipit24:
@lisa-fehr I think that looks good! It's still about the same size as the minimap in Camunda and it works just the same as before.

tuantruong625:
@lisa-fehr @chipit24 @JCook21 Not a big fan of the white background on the tool bar.
What about adding `mw-100` and `p-0` to the parent div of ModelerApp.vue? And the stand-alone modeler will use 100% of the viewport.

chipit24:
@tuantruong625 I don't see a white bar with the latest changes here.

lisa-fehr:
@tuantruong625 oops. Left some code in for the background. 

I'm not sure about expanding the width to 100%. It makes it harder to test the minimum screen width. Unless there isn't one.

PR #712, opened 2019-09-30
Refactor paper

JCook21:
Began work to extract and remove direct references to the JointJS paper from our code.

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -195,7 +195,7 @@ export default {
       this.cursor = null;
     },
     scale(scale) {
-      this.paperManager.paper.scale(scale);
+      this.paperManager.scale = scale;
     },
     currentXML() {
       this.validateIfAutoValidateIsOn();
@@ -614,12 +614,12 @@ export default {
     },
     async renderPaper() {
       await this.$nextTick();
-      this.paperManager.freezePaper();
-      this.isRendering = true;
-      await this.waitForCursorToChange();
-      this.paperManager.addOnceHandler('render:done', () => this.isRendering = false);
-      this.parse();
-      this.paperManager.unfreezePaper();
+      await this.paperManager.performAtomicAction(async() => {
+        this.isRendering = true;
+        await this.waitForCursorToChange();
+        this.paperManager.addOnceHandler('render:done', () => this.isRendering = false);
+        this.parse();
+      });
       this.$emit('parsed');
       this.removeLoaderIfProcessIsEmpty();
     },
@@ -689,7 +689,7 @@ export default {
 
       // Handle transform
       const paperOrigin = this.paperManager.paper.localToPagePoint(0, 0);
-      const scale = this.paperManager.paper.scale();
+      const scale = this.paperManager.scale;
 
       diagram.bounds.x = (clientX - paperOrigin.x) / scale.sx;
       diagram.bounds.y = (clientY - paperOrigin.y) / scale.sy;
@@ -933,7 +933,7 @@ export default {
       this.setShapeStacking(shape);
 
       shape.component.$emit('click');
-    }, this);
+    });
 
     /* Register custom nodes */
     window.ProcessMaker.EventBus.$emit('modeler-start', {
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -51,7 +51,7 @@
       <div ref="paper" data-test="paper" class="main-paper" />
     </b-col>
 
-    <mini-paper :isOpen="miniMapOpen" :paper="paper" :graph="graph" />
+    <mini-paper :isOpen="miniMapOpen" :paperManager="paperManager" :graph="graph" />
 
     <b-col class="pl-0 h-100 overflow-hidden inspector-column" :class="{ 'ignore-pointer': canvasDragPosition }">
       <InspectorPanel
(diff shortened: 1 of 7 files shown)


src/components/Modeler.vue
```diff
@@ -887,59 +885,45 @@ export default {
...
-      perpendicularLinks: true,
-      interactive: this.graph.get('interactiveFunc'),
-      highlighting: {
-        default: { options: { padding: highlightPadding } },
-      },
-    });
-
-    this.paper.translate(168, 20);
+    this.paperManager = PaperManager.factory(this.$refs.paper, this.graph.get('interactiveFunc'), this.graph);
+    this.paper = this.paperManager.paper;
```
JCook21: This reference is still required for now as other components still need the paper object.

src/components/Modeler.vue
```diff
@@ -614,12 +614,12 @@ export default {
...
       await this.$nextTick();
-      this.paper.freeze();
+      this.paperManager.freezePaper();
       this.isRendering = true;
       await this.waitForCursorToChange();
-      this.paper.once('render:done', () => this.isRendering = false);
+      this.paperManager.addOnceHandler('render:done', () => this.isRendering = false);
       this.parse();
-      this.paper.unfreeze();
+      this.paperManager.unfreezePaper();
```
JCook21: I tried to convert this code to use the `performAtomicAction()` method but I couldn't make it work. Something about the async/await behaviour here screwed this up. This is something that can be tackled later.

src/components/Modeler.vue
```diff
@@ -688,8 +688,8 @@ export default {
       const diagram = this.nodeRegistry[control.type].diagram(this.moddle);
 
       // Handle transform
-      const paperOrigin = this.paper.localToPagePoint(0, 0);
-      const scale = this.paper.scale();
+      const paperOrigin = this.paperManager.paper.localToPagePoint(0, 0);
+      const scale = this.paperManager.paper.scale();
```
chipit24: You can use the class getter:
```suggestion
      const scale = this.paperManager.scale;
```

src/components/Modeler.vue
```diff
@@ -195,7 +195,7 @@ export default {
       this.cursor = null;
     },
     scale(scale) {
-      this.paper.scale(scale);
+      this.paperManager.paper.scale(scale);
```
chipit24: Since you have a getter for `scale`, I think adding a setter for it would be nice as well. Then you can have:
```js
this.paperManager.scale = scale;
```

src/components/Modeler.vue
```diff
@@ -949,7 +933,7 @@ export default {
       this.setShapeStacking(shape);
 
       shape.component.$emit('click');
-    });
+    }, this);
```
chipit24: Using an arrow function will bind the value of this implicitly, so I don't think you need to pass `this` as the 3rd argument (just like the event handler above).

src/components/paperManager.js
```diff
@@ -0,0 +1,67 @@
+import { dia } from 'jointjs';
+import { highlightPadding } from '@/mixins/crownConfig';
+
+export default class {
+  constructor(paper) {
+    this._paper = paper;
```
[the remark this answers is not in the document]
JCook21: I'd be interested in doing that. The reason for naming this `_paper` though was to avoid infinite recursion in the getter called paper. Would using the private class field take care of that issue?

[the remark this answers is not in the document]
chipit24: It seems like it would, since the private property would be accessed and assigned using `this.#paper` and not just `this.paper`.

src/components/paperManager.js
```diff
@@ -0,0 +1,67 @@
...
+
+  get scale() {
+    return this._paper.scale();
+  }
+
+  translate(x, y) {
+    this._paper.translate(x, y);
+  }
+
+  addEventHandler(eventName, callback, objectScope) {
```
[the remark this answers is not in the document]
JCook21: Much better name, thanks!

src/components/Modeler.vue
JCook21: I missed this call, I'll change it now. Thanks.

src/components/paperManager.js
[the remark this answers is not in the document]
JCook21: Let's give it a spin and see what happens. :)

src/components/Modeler.vue
chipit24: @JCook21 I forgot to include this in my review. You can make `performAtomicAction` an async function and `await` the callback:
```js
async performAtomicAction(callback) {
  this._paper.freeze();
  await callback(this._paper);
  this._paper.unfreeze();
}
```

Then here you can have:
```js
performAtomicAction(async () => {
  this.isRendering = true;
  await this.waitForCursorToChange();
  this.paperManager.addOnceHandler('render:done', () => this.isRendering = false);
  this.parse();
});
```

src/components/Modeler.vue
JCook21: But won't this mean we pass undefined as the callbackScope? Will that work then with the internal call? I'll try it to see if it still works.

src/components/Modeler.vue
JCook21: Let me try it. I tried something like this earlier but it didn't work. I probably missed something...

src/components/Modeler.vue
chipit24: Not including the argument should be the same as explicitly passing `undefined`.

JCook21: It seems to work!

PR #717, opened 2019-10-01
Added unit test

JCook21:
Added unit test for method the calculates the new position for the paper when the mini map is clicked.

Coverage before:
Screen Shot 2019-10-01 at 2 12 47 PM

Coverage after:
Screen Shot 2019-10-01 at 2 12 54 PM


tests/unit/miniMapManager.spec.js
```diff
@@ -0,0 +1,60 @@
...
+      871,
+      {
+        x: 254.109375,
+        y: 312,
+      },
+      {
+        newX: 314.890625,
+        newY: 123.5,
+      },
+    ],
```
chipit24: I'm noticing some duplicate numbers between this block and the one below it. I wonder if it's worth having these in variables, or using different numbers.

For example, the values being used for `scaleX` and `scaleY` are both `1`, and it's used in the expression `(x * scaleX)` and `(y * scaleY)`, respectively. So by testing this logic with the number 1, it's essentially the same as `(x * scaleX) === x * 1 === x`, and how this function actually reacts to larger or smaller scales may not be fully tested.

JCook21: It's a good point. I got these numbers directly from the debugger in the browser. I can definitely make these constants for the test. I'm not sure that they're ever different from 1 in production though...

chipit24: > I'm not sure that they're ever different from 1 in production though...

They are if the user decides to scale the paper. Pressing the `+` or `-` buttons to the left of "RESET" in the toolbar scales the paper, and hence the value of scale will go up or down.

JCook21: That makes sense. Let me plug in some test cases where scale X and Y are not 1.

PR #718, opened 2019-10-01
Highlight shape rather than shape and label

### LINKED ISSUE #705: The yellow box around the element should be around the element, not including the label
For elements that have a label outside of the shape (e.g. events and gateways), when selecting them on the grid, the highlight outline encloses the entire shape and the label:
Screen Shot 2019-09-30 at 10 13 39 AM
Screen Shot 2019-09-30 at 10 13 35 AM

The outline should only be around the shape and _not_ the label:
Screen Shot 2019-09-30 at 10 15 02 AM
Screen Shot 2019-09-30 at 10 14 48 AM

lisa-fehr:
Closes #705

**Estimated appearance:** Using shape instead of a box.
Screen Shot 2019-10-01 at 4 58 58 PM

Custom highlight should only show when you click on an element and should update when resizing a pool.

**Note:** I did not alter the error highlighting box. It would revert the select highlight to the shape and label on some elements like gateways.

**Current error highlighting:**
Screen Shot 2019-10-03 at 10 18 42 AM

### DIFF
--- a/src/mixins/crownConfig.js
+++ b/src/mixins/crownConfig.js
@@ -55,7 +55,6 @@ export default {
       allowSetNodePosition: true,
       savePositionOnPointerupEventSet: false,
       shape: null,
-      highlightShape: null,
     };
   },
   watch: {
--- a/src/mixins/crownConfig.js
+++ b/src/mixins/crownConfig.js
@@ -26,6 +26,16 @@ const errorHighlighter = {
   },
 };
 
+const defaultHighlighter = {
+  name: 'stroke',
+  options: {
+    attrs: {
+      stroke: '#feb663',
+      'stroke-width': 3,
+    },
+  },
+};
+
 export default {
   props: [
     'highlighted',
@@ -45,15 +55,16 @@ export default {
       allowSetNodePosition: true,
       savePositionOnPointerupEventSet: false,
       shape: null,
+      highlightShape: null,
     };
   },
   watch: {
     highlighted(highlighted) {
       if (highlighted) {
-        this.shapeView.highlight();
+        this.shapeView.highlight(this.selectShape, { highlighter: defaultHighlighter });
         this.addCrown();
       } else {
-        this.shapeView.unhighlight();
+        this.shapeView.unhighlight(this.selectShape, { highlighter: defaultHighlighter });
         this.removeCrown();
       }
     },
@@ -68,6 +79,9 @@ export default {
   computed: {
     shapeView() {
       return this.shape.findView(this.paper);
+    },
+    selectShape() {
+      return this.shapeView.$el.find('[joint-selector=body]');
     },
     isPool() {
       return this.node.type === 'processmaker-modeler-pool';
@@ -236,8 +250,8 @@ export default {
       this.shape.on('change:size', () => {
         if (this.highlighted) {
           /* Ensure the highlight box expands to fit element */
-          shapeView.unhighlight();
-          shapeView.highlight();
+          shapeView.unhighlight(this.selectShape, { highlighter: defaultHighlighter });
+          shapeView.highlight(this.selectShape, { highlighter: defaultHighlighter });
         }
       });
     },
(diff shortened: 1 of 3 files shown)


tests/e2e/specs/BoundaryTimerEvent.spec.js
```diff
@@ -239,24 +238,6 @@ describe('Boundary Timer Event', () => {
...
-            });
-
-            moveElement(task2Position, 300, 500);
-            waitToRenderAllShapes();
-            cy.wrap($boundaryEvent).should($el => {
-              const { left, top } = $el.position();
-              const positionErrorMargin = 2;
-              expect(left).to.not.be.closeTo(initialBoundaryEventPosition.left, positionErrorMargin);
-              expect(top).to.not.be.closeTo(initialBoundaryEventPosition.top, positionErrorMargin);
-            });
```
chipit24: Why where these tests removed? I added them back and they appear to be failing; I don't think the solution to failing tests is to remove them altogether.

It appears this is suppose to be testing that boundary events snap back to their original position, and that moving the original task doesn't change the position of the boundary event which should now be attached to the other task.

It seems the first part of this ("snaps back to original position") may already captured in another test, but not the other part. If this test is unclear, maybe it can be moved into a separate test. What do you think?

src/mixins/crownConfig.js
```diff
@@ -69,6 +79,9 @@ export default {
     shapeView() {
       return this.shape.findView(this.paper);
     },
+    selectShape() {
+      return this.shapeView.$el.find('[joint-selector=body]');
+    },
```
chipit24: I think the name of this could be improved. In my opinion, `selectShape` sounds like a verb (_select_ the shape), or a shape that is of type "select". What about a name like `shapeBody`?

tests/e2e/specs/BoundaryEventCommonBehaviour.spec.js
```diff
@@ -157,6 +157,8 @@ boundaryEventData.forEach(({ type, nodeType, eventXMLSnippet }) => {
 
         waitToRenderAllShapes();
 
+        getElementAtPosition(taskPosition).click();
```
chipit24: Why is this required for the test to pass?

tests/e2e/specs/BoundaryTimerEvent.spec.js
JCook21: Lisa and I discussed this. The test was testing two things and I suggested removing this. I think Lisa noted this and added a task to add a separate (common?) test for this case. The failure also seems to be related to some potentially buggy behaviour around how labels are hidden/re-shown.

src/mixins/crownConfig.js
```diff
@@ -50,10 +60,10 @@ export default {
   watch: {
     highlighted(highlighted) {
       if (highlighted) {
-        this.shapeView.highlight();
+        this.shapeView.highlight(this.selectShape, { highlighter: defaultHighlighter });
         this.addCrown();
       } else {
```
JCook21: Tiny style thing: how would you feel about using an early return and not an else here?

tests/e2e/specs/BoundaryEventCommonBehaviour.spec.js
lisa-fehr: The `$el.position()` below is off by about 69 pixels left if I don't do this. Something different about the highlight shape's left position.

tests/e2e/specs/BoundaryTimerEvent.spec.js
lisa-fehr: I added a task to at least add the label back after dropping a boundary event from one task to another. #723 
That might be related to why the left position is so different, might not.

I can add another task to re-add the test for "if the boundary event is in the expected location after dragging a task".

lisa-fehr: For whatever reason the $boundaryEvent.position() and $el.position() left position is off about 62 pixels. But manually it is working fine.

lisa-fehr: If I click on a task after the boundary event was moved and then store the event position - it looks like I can get the second part of the test to work. I'm not as familiar with the cy.wrap so I was trying to click with the wrong functions before. I could make an issue to split out those tests, or do it now.

PR #719, opened 2019-10-01
Set advanced accordion to last item in inspector

### LINKED ISSUE #709: Advanced Accordion always last
The advanced accordion should always be the last accordion in the inspector panel:

Screen Shot 2019-09-30 at 11 18 43 AM

In some cases, the advanced accordion does not appear last, like for the task component:
Screen Shot 2019-09-30 at 11 19 20 AM

This occurs because calls to `registerInspectorExtension` does not maintain the order of the inspector accordions.

tuantruong625:
**Processmaker video**
https://drive.google.com/open?id=19XYJ2GgGpaP51A7TpGifi27xTO6W9n69

Fixes #709

### DIFF
--- a/src/components/InspectorExtensionManager.js
+++ b/src/components/InspectorExtensionManager.js
@@ -1,0 +1,42 @@
+export default class InspectorExtensionManager {
+  static registerInspectorExtension(node, config) {
+    const inspectorItems = this.getInspectorItems(node, config);
+    this.addInspectorItem(inspectorItems, config);
+    inspectorItems.sort(this.moveAdvancedAccordionToBottom);
+  }
+
+  static isAdvancedAccordion(accordion) {
+    return accordion.name === 'advanced';
+  }
+
+  static moveAdvancedAccordionToBottom(accordionA, accordionB) {
+    if (this.isAdvancedAccordion(accordionA)) {
+      return 1;
+    }
+
+    if (this.isAdvancedAccordion(accordionB)) {
+      return -1;
+    }
+
+    return 0;
+  }
+
+  static addInspectorItem(inspectorItems, config) {
+    const nodeIndex = inspectorItems.findIndex(item => {
+      return config.id && config.id === item.id;
+    });
+    if (nodeIndex === -1) {
+      inspectorItems.push(config);
+      return;
+    }
+    inspectorItems[nodeIndex] = config;
+  }
+
+  static getInspectorItems(node, config) {
+    let inspectorItems = node.inspectorConfig[0].items;
+    if (typeof config.container === 'undefined') {
+      inspectorItems = inspectorItems[0].items;
+    }
+    return inspectorItems;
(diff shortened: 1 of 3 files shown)


src/components/Modeler.vue
```diff
@@ -385,6 +385,8 @@ export default {
       } else {
         inspectorItems[existingIndex] = config;
       }
+
+      inspectorItems.push(...inspectorItems.splice(inspectorItems.findIndex(item => item.config.name === 'advanced'), 1));
```
chipit24: I find this a little hard to read. Do you think you can split it up into multiple lines or a separate function?

Or maybe this line could be a `sort`, something like:
```js
function isAdvancedAccordion(accordion) {
  return accordion.name === 'advanced';
}

function moveAdvancedAccordionToBottom(accordionA, accordionB) {
  if (isAdvancedAccordion(accordionA)) {
    return 1;
  }

  if (isAdvancedAccordion(accordionB)) {
    return -1;
  }

  return 0;
}

inspectorItems.sort(moveAdvancedAccordionToBottom);
```

Then the sorting behaviour could be unit tested. Actually, the whole `addToInspector` function could be pulled out and unit tested (since it has no reference to `this`)–then the sorting functions would simply be implementation details.

JCook21: I agree with @chipit24 that this could be more readable if broken out. I also agree that it would be great to see a unit test for this.

JCook21:
@chipit24 @tuantruong625 @lisa-fehr I made the refactoring so that we could move this story along if needed. Do you think we should add unit tests now or later?

JCook21:
Screen Shot 2019-10-02 at 4 18 18 PM
Actually, looks like it should be done now: we have no coverage for this functionality at all through the cypress tests.

JCook21:
Screen Shot 2019-10-03 at 11 03 20 AM

Now with boar coverage!

src/components/InspectorExtensionManager.js
```diff
@@ -0,0 +1,38 @@
+export default class InspectorExtensionManager {
```
chipit24: If this is only housing static methods, is there any advantage to having them in a class as oppose to exporting them as standalone functions? Similar to how we have our `/tests/e2e/support/utils.js` file set up? (Though that file is growing a bit big now).

Keeping them in a class like this means we always have to import the entire class, and we have the extra overhead/syntax of a class to worry about.

JCook21: I agree and I think the fact that all of the methods are static is the smell that gives this away. Let me change the code to this, it's a better structure.

JCook21:
I've noticed that this doesn't seem to be working properly when linked into processmaker as a whole. I only see the config for name and advanced on all elements. It appears to be find in the standalone modeler though. I'm going to wait until we have the latest of all packages to test further though.

PR #720, opened 2019-10-02
Update call activity to use multiselect

### LINKED ISSUE #706: Inspector select dropdown needs to be vue-multiselect for filtering

chipit24:
Fixes #706.

This PR requires the changes in https://github.com/ProcessMaker/vue-form-elements/pull/105 to be merged and published before merging this PR. Tests may be failing until that happens.

Screen Shot 2019-10-02 at 2 02 48 PM

### DIFF
--- a/src/components/inspectors/InspectorPanel.vue
+++ b/src/components/inspectors/InspectorPanel.vue
@@ -25,6 +25,7 @@ import {
   FormRadioButtonGroup,
   FormAccordion,
   FormDatePicker,
+  FormMultiSelect,
 } from '@processmaker/vue-form-elements';
 import '@processmaker/vue-form-elements/dist/vue-form-elements.css';
 import store from '@/store';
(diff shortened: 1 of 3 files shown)


chipit24:
Just noticed a bug when de-selecting a process in the list:
```
[Vue warn]: Error in v-on handler: "TypeError: Cannot read property 'value' of null"
```

You can hold off on reviewing until I fix this.

JCook21:
@chipit24 I don't think this can be merged until the changes in the upstream component are published. We need to keep our repo in a state where tests are passing.

chipit24:
@JCook21 Yup–I mentioned that in the PR description but I should have kept this in a draft to prevent it from being accidentally merged.

JCook21:
One comment on the screenshot above: the highlight colour on the selected element seems to be green. Does this match the colour scheme of the modeler?

chipit24:
@JCook21 Is there some colour scheme you are referring to? I'm not aware of any colour scheme that this should match, and it appears that the same green is used across all other multiselects.

JCook21:
> @JCook21 Is there some colour scheme you are referring to? I'm not aware of any colour scheme that this should match, and it appears that the same green is used across all other multiselects.

@chipit24 my bad, sorry. I didn't remember seeing that green anywhere else but I must be mistaken.

chipit24:
@JCook21 There are some differences in the bootstrap theme between standalone and processmaker, so it's good that you brought that up, as this PR shouldn't cause a colour change. I checked the colours in processmaker and everything appears as it should.

PR #722, opened 2019-10-02
Fix sequence flow alignment

### LINKED ISSUE #690: Round elements seem to be horizontally off centre
Round elements seem to be horizontally off centre, causing the initial flow element to be not straight

chipit24:
Fixes #690.

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -687,12 +687,9 @@ export default {
       // Now, let's modify planeElement
       const diagram = this.nodeRegistry[control.type].diagram(this.moddle);
 
-      // Handle transform
-      const paperOrigin = this.paperManager.localToPagePoint(0, 0);
-      const scale = this.paperManager.scale;
-
-      diagram.bounds.x = (clientX - paperOrigin.x) / scale.sx;
-      diagram.bounds.y = (clientY - paperOrigin.y) / scale.sy;
+      const { x, y } = this.paperManager.clientToGridPoint(clientX, clientY);
+      diagram.bounds.x = x;
+      diagram.bounds.y = y;
 
       if (this.isBoundaryEvent(definition)) {
         this.setShapeCenterUnderCursor(diagram);
--- a/src/components/nodes/startEvent/startEventConfig.js
+++ b/src/components/nodes/startEvent/startEventConfig.js
@@ -1,0 +1,1 @@
+export const startEventDiameter = 36;
--- a/src/components/nodes/task/index.js
+++ b/src/components/nodes/task/index.js
@@ -1,8 +1,8 @@
 import component from './task.vue';
 import idConfigSettings from '@/components/inspectors/idConfigSettings';
 import nameConfigSettings from '@/components/inspectors/nameConfigSettings';
+import { taskHeight, taskWidth } from './taskConfig';
 
-export const taskHeight = 76;
 export const id = 'processmaker-modeler-task';
 
 export default {
@@ -22,7 +22,7 @@ export default {
     return moddle.create('bpmndi:BPMNShape', {
       bounds: moddle.create('dc:Bounds', {
         height: taskHeight,
-        width: 116,
+        width: taskWidth,
       }),
     });
   },
--- a/tests/unit/paperManager.spec.js
+++ b/tests/unit/paperManager.spec.js
@@ -1,0 +1,31 @@
+import PaperManager from '@/components/paperManager';
+
+describe('Paper Manager', () => {
+  const positionData = [
+    [{ sx: 1, sy: 1 }, { x: 319, y: 77 }, 711, 320, { x: 390, y: 240 }],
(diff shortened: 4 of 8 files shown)


src/components/nodes/task/index.js
```diff
@@ -1,8 +1,8 @@
 import component from './task.vue';
 import idConfigSettings from '@/components/inspectors/idConfigSettings';
 import nameConfigSettings from '@/components/inspectors/nameConfigSettings';
+import { taskHeight, taskWidth } from './taskConfig';
 
-export const taskHeight = 76;
```
JCook21: What's the advantage of moving these constants to a separate file?

src/components/nodes/startEvent/startEventConfig.js
```diff
@@ -0,0 +1 @@
+export const startEventDiameter = 36;
```
JCook21: What's the advantage of having this single constant in a file?

src/components/Modeler.vue
```diff
@@ -687,12 +687,9 @@ export default {
...
 
-      // Handle transform
-      const paperOrigin = this.paperManager.localToPagePoint(0, 0);
-      const scale = this.paperManager.scale;
-
-      diagram.bounds.x = (clientX - paperOrigin.x) / scale.sx;
-      diagram.bounds.y = (clientY - paperOrigin.y) / scale.sy;
+      const { x, y } = this.paperManager.clientToGridPoint(clientX, clientY);
+      diagram.bounds.x = x;
+      diagram.bounds.y = y;
```
JCook21: Nice extraction!

tests/unit/paperManager.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    'Can convert client position to grid position',
+    (scale, paperOrigin, clientX, clientY, expectedPoint) => {
+      paper.scale = () => scale;
+      paper.localToPagePoint = () => paperOrigin;
+
+      const actualPoint = paperManager.clientToGridPoint(clientX, clientY);
+      expect(actualPoint).toEqual(expectedPoint);
+    });
+});
+
```
JCook21: Awesome! 🔥

tests/unit/paperManager.spec.js
```diff
@@ -0,0 +1,31 @@
...
+  let paperManager;
+  let paper;
+
+  beforeEach(() => {
+    paper = {};
+    paperManager = new PaperManager(paper);
+  });
+
+  test.each(positionData)(
```
JCook21: Do you prefer `test` or `it` for these? They're equivalent so it's a matter of preference. I think we should keep it consistent though.

src/components/nodes/task/index.js
JCook21: I've just seen them used in tests, which makes sense. I still wonder about not extracting them to separate files though.

src/components/nodes/startEvent/startEventConfig.js
JCook21: I've just seen them used in tests, which makes sense. I still wonder about not extracting them to separate files though.

JCook21:
I think this seems good to go. What's remaining to be done to take this out of WIP?

tests/unit/paperManager.spec.js
chipit24: Oh, wasn't paying attention to that! I'll change it to use `it` for consistency.

src/components/nodes/startEvent/startEventConfig.js
chipit24: With this, and the task height/width, if I keep them in the same `index.js` file, Cypress complains that it can't parse the file due to the imports:

```bash
SyntaxError: 'import' and 'export' may appear only with 'sourceType: module'
    at EventEmitter.handler (/Users/robertkomaromi/Library/Caches/Cypress/3.4.1/Cypress.app/Contents/Resources/app/packages/server/lib/plugins/util.js:67:29)
    at emitOne (events.js:116:13)
    at EventEmitter.emit (events.js:211:7)
    at ChildProcess.<anonymous> (/Users/robertkomaromi/Library/Caches/Cypress/3.4.1/Cypress.app/Contents/Resources/app/packages/server/lib/plugins/util.js:25:29)
    at emitTwo (events.js:126:13)
    at ChildProcess.emit (events.js:214:7)
    at emit (internal/child_process.js:772:12)
    at _combinedTickCallback (internal/process/next_tick.js:141:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
```

Our babel setup should handle the ES module synatax, so it appears that Cypress doesn't run the test files through babel, or there's some configuration mismatch. I bug into it a little but didn't get far so decided this was the easier option.

I can look into it more, as it would be nicer so keep them in the same file.

JCook21: Interesting. Let's not hold this up for now but we can fix that later if you like.

chipit24: Sounds good; I can fix it later and not let this hold it up right now.

PR #726, opened 2019-10-07
Upgrade vue-form-elements

chipit24:
Upgrade vue-form-elements and fix the integration with the new datepicker.

### DIFF
--- a/src/components/inspectors/InspectorPanel.vue
+++ b/src/components/inspectors/InspectorPanel.vue
@@ -23,7 +23,6 @@ import {
   FormTextArea,
   FormCheckbox,
   FormRadioButtonGroup,
-  FormCodeEditor,
   FormAccordion,
   FormDatePicker,
 } from '@processmaker/vue-form-elements';
@@ -43,7 +42,6 @@ Vue.component('FormSelect', FormSelect);
 Vue.component('FormTextArea', FormTextArea);
 Vue.component('FormCheckbox', FormCheckbox);
 Vue.component('FormRadioButtonGroup', FormRadioButtonGroup);
-Vue.component('FormCodeEditor', FormCodeEditor);
 Vue.component('FormAccordion', FormAccordion);
 Vue.component('FormDatePicker', FormDatePicker);
 Vue.component('VueFormRenderer', VueFormRenderer);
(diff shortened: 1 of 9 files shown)


lisa-fehr:
This fixes the issue I get when I drag on an `intermediate timer event`, select `datetime` and then immediately click `download xml`.

chipit24:
Locally, when I `npm link` the latest `vue-form-elements` into this branch, things works well. But when I pull things into processmaker and perform the following steps, the app freezes:
1. Drag an _Intermediate Timer Event_ onto the grid.
2. Select the event, expand the Timing Control accordion, and select Date/Time for the Type.
3. Click on the date input under _Wait until specified date/time_.
4. The app freezes right away (the datepicker doesn't even open).

src/components/inspectors/InspectorPanel.vue
```diff
@@ -23,7 +23,6 @@ import {
   FormTextArea,
   FormCheckbox,
   FormRadioButtonGroup,
-  FormCodeEditor,
```
chipit24: Vue complained about this component not being found; it appears it's not actually used or defined anywhere.

src/setup/globals.js
```diff
@@ -30,4 +30,9 @@ window.ProcessMaker = {
 
     window.ProcessMaker.EventBus.$emit('alert', window.ProcessMaker.navbar.alerts);
   },
+  // user: {
+  //   timezone: 'America/Recife',
+  //   lane: 'en',
+  //   datetime_format: 'MM/DD/YYYY HH:mm',
+  // },
```
chipit24: I added this for testing but left it commented out, as this changes the displayed format and thus will cause tests to fail.

Should I leave it like this for now? I think it would be nice to add it to a test that confirms that things work in the local timezone and in some other user defined timezone.

lisa-fehr: Seems like an integration test. Wonder how that would work here. @JCook21 ?

chipit24: Since this info exists on the window object, I think we can change that object from the test. The XML generated should be the same, but that formatting the user sees in the input will be different.

JCook21:
@chipit24 I think a new vue-form-elements has been published. Can we move this along now?

chipit24:
Yup, saw it published yesterday. On it!

src/setup/globals.js
chipit24: I'll leave this for now and pull it out into a test in a separate PR.

JCook21: Can it be deleted? I don't think leaving commented code in is ever a good idea...

PR #728, opened 2019-10-08
Reduce size of bundled modeler package

JCook21:
This work reduces the size of the bundled modeler package by about 50%. It has been tested in conjunction with changes in the process maker project and may depend on those.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "build": "vue-cli-service build",
     "open-cypress": "TZ=UTC cypress open",
     "lint": "vue-cli-service lint --no-fix",
-    "build-bundle": "vue-cli-service build --target lib --name modeler ./src/components/nodes/index.js",
+    "build-bundle": "vue-cli-service build --modern --target lib --name modeler ./src/components/nodes/index.js",
     "lint-fix": "vue-cli-service lint",
     "pretest": "rm -rf .nyc_output || true",
     "test": "npm run test-unit && npm run test-ci",
@@ -79,7 +79,10 @@
     "stylelint-config-recommended-scss": "^3.2.0",
     "stylelint-scss": "^3.3.2",
     "vue-loader": "latest",
-    "vue-template-compiler": "^2.5.17"
+    "vue-template-compiler": "^2.5.17",
+    "webpack-bundle-analyzer": "^3.5.2",
+    "webpack-cli": "^3.3.9",
+    "yargs": "^14.2.0"
   },
   "peerDependencies": {
     "@processmaker/screen-builder": "0.19.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -6,7 +6,7 @@
     "build": "vue-cli-service build",
     "open-cypress": "TZ=UTC cypress open",
     "lint": "vue-cli-service lint --no-fix",
-    "build-bundle": "vue-cli-service build --target lib --name modeler ./src/components/nodes/index.js",
+    "build-bundle": "vue-cli-service build --modern --target lib --name modeler ./src/components/nodes/index.js",
```
JCook21: I'm not totally sure this is doing anything for us yet. I'll dig a little more into this.

vue.config.js
```diff
@@ -26,12 +28,32 @@ module.exports = {
...
+      }
+      return externals;
+    })(),
+    plugins: (() => {
+      const argv = yargs
+        .boolean('analyze')
+        .argv;
+      const plugins = [];
+      if (argv.analyze) {
+        plugins.push(new BundleAnalyzerPlugin());
```
JCook21: This allows us to analyze the bundle size if the `--analyze` flag is passed to the build.

vue.config.js
```diff
@@ -26,12 +28,32 @@ module.exports = {
...
+        externals.push(
+          'vue',
+          /^bootstrap\/.+$/,
+          /^@processmaker\/.+$/,
+          'i18next',
+          '@panter/vue-i18next',
+          'jointjs',
+          'luxon',
+          'bpmn-moddle',
+          'lodash',
```
JCook21: Many of these dependencies were also being included in the process maker project. Including them in modeler was just causing bloat so I moved to exclude them..

src/portsUtils.js
```diff
@@ -1,4 +1,4 @@
-import { g } from 'jointjs/dist/joint';
+import { g } from 'jointjs';
```
JCook21: Changing this import allows jointjs to be moved to externals.

JCook21:
> I just suggested removing the `modern` flag for now; you can let me know if that makes sense.

Thanks @chipit24, I'll look some more into it. With this option builds give a warning saying that browser list is the better way to do this so I'm not sure...

package.json
chipit24: I wrote this comment up but looks like I forgot to include it in the review and it was lost:

I looked into the docs and it seems that in order to take advantage of modern mode, we'd have to include the appropriate `module` and `nomodule` scripts where the app is loaded.

However, when I run the build, I get the following warning:

>  WARN  Modern mode only works with default target (app). For libraries or web components, use the browserslist config to specify target browsers.

And it seems the same files are generated whether or not the `--modern` flag is set. So unless we significantly change our build process (switch from a lib build target to the default app build target), this change may only cause some confusion.

PR #729, opened 2019-10-08
Show label after dropping boundary event on another task

### LINKED ISSUE #723: Label should reappear after dropping boundary event on another task
# Bug report issue
## Description
When dropping a boundary event from one task to another, the label doesn't return on embed.
## Reproduction steps
1. Drag two tasks onto the screen
2. Drag a boundary event onto a task
3. Drag the boundary event from that task to the second task
## Expected behaviour
After dragging the boundary event from one task to another task, the boundary event is selected and the label should reappear.
## Actual behaviour
After dragging the boundary event from one task to another task, the boundary event is selected but the label is missing.

**After the boundary event is dropped onto another task:**
Screen Shot 2019-10-03 at 10 13 21 AM

Video: https://www.dropbox.com/s/e72c3dkc24b83vi/label-bug.mov

### LINKED ISSUE #724: Re-add removed part of test from "moves to another task when dragged over"
## Description
Start after #718 is merged

https://github.com/ProcessMaker/modeler/pull/718/files#diff-dd7181926d02b18e486a36f5ff941f91L242
There is a test with two tasks where you drag a boundary event from one task to the other, then drag one task and then the other. The position of the boundary event is calculated with `close to` and `not close to`. But it was off by about 62px left. It needs to be added as a new test, or the position mystery solved (maybe it included the missing label somehow before the highlighter change from box&label to shape).

#723 might help with this issue.

lisa-fehr:
Closes #723 
Closes #724

Added a listener for when the shape is embedded in another parent.

Added a test back that was failing. Due to the label being missing after dragging & dropping a Boundary Event from one task to another.

### DIFF
--- a/src/mixins/hideLabelOnDrag.js
+++ b/src/mixins/hideLabelOnDrag.js
@@ -8,8 +8,9 @@ export default {
 
       this.shape.attr('label/display', 'none');
     },
-    showLabel() {
+    async showLabel() {
       this.shape.attr('label/display', 'initial');
+      await this.$nextTick();
       this.shape.once('change:position', this.hideLabel);
     },
   },
@@ -18,5 +19,6 @@ export default {
 
     this.shape.once('change:position', this.hideLabel);
     this.shape.listenTo(this.paper, 'cell:pointerup blank:pointerup', this.showLabel);
+    this.shape.on('change:parent', this.showLabel);
   },
 };
--- a/src/mixins/hideLabelOnDrag.js
+++ b/src/mixins/hideLabelOnDrag.js
@@ -19,6 +19,6 @@ export default {
 
     this.shape.once('change:position', this.hideLabel);
     this.shape.listenTo(this.paper, 'cell:pointerup blank:pointerup', this.showLabel);
-    this.shape.on('change:parent', this.showLabel);
+    this.shape.once('change:parent', this.showLabel);
   },
 };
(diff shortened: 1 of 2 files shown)


src/mixins/hideLabelOnDrag.js
```diff
@@ -8,8 +8,9 @@ export default {
 
       this.shape.attr('label/display', 'none');
     },
-    showLabel() {
+    async showLabel() {
       this.shape.attr('label/display', 'initial');
+      await this.$nextTick();
```
chipit24: Going to add some context here as it wasn't clear right away why this fixes the problem:

When a boundary event is dragged off a task, it will snap back to its original task, or onto a new task.

When the mouse is released (`pointerup` event), it will show the label but also add an event listener right away, to hide the label again on position change. And we have a race condition!

The event listener is added before the boundary event snaps to that task, causing it to be triggered right away, hiding the label.

src/mixins/hideLabelOnDrag.js
```diff
@@ -18,5 +19,6 @@ export default {
 
     this.shape.once('change:position', this.hideLabel);
     this.shape.listenTo(this.paper, 'cell:pointerup blank:pointerup', this.showLabel);
+    this.shape.on('change:parent', this.showLabel);
```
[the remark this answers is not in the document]
lisa-fehr: Weird. I thought it did more. Currently, the only difference when this line is there, is the label disappears on the first drag. Without it, the label hangs around initially before behaving as expected. In this case, `once` is good enough for the same effect.

Screen_Shot_2019-10-09_at_10_09_41_AM__2_
Screen_Shot_2019-10-09_at_10_10_12_AM__2_

src/mixins/hideLabelOnDrag.js
lisa-fehr: Yup. Sometimes showLabel won, other times it was hideLabel. 👍

PR #734, opened 2019-10-09
Keep all boundary events in position when moving pools

### LINKED ISSUE #731: Boundary Error Events in Pools do not stay in place when the Pool is dragged
# Bug report issue
## Description
Boundary Error Events in pools, do not stay in the correct position when the pool is dragged.
## Reproduction steps
1. add a Boundary Error Event onto a task
2. add a Boundary Timer Event onto the same task
2. add a Pool around the task
3. drag the Pool
## Expected behaviour
The Boundary Error Event should behave the same as the Boundary Timer Event.
## Actual behaviour
The Boundary Error Event position moves off the task when the Pool is moved.
## Software used
Modeler v. 0.17.0
Latest Chrome
## Other information
https://www.dropbox.com/s/x0d2fkd08np0999/errorEventPoolBug.mov?dl=0

lisa-fehr:
Closes #731

Test confirmed that the boundary events were still embedded, but did not confirm that the boundary event was not moving as the pool moved.

### DIFF
--- a/src/components/nodes/pool/pool.vue
+++ b/src/components/nodes/pool/pool.vue
@@ -19,6 +19,8 @@ import { invalidNodeColor, defaultNodeColor, poolColor } from '@/components/node
 import pull from 'lodash/pull';
 import store from '@/store';
 
+const boundaryEventsPattern = /^processmaker-modeler-boundary-\D*-event$/;
+
 const Pool = shapes.standard.Rectangle.define('processmaker.modeler.bpmn.pool', {
   markup: [
     ...shapes.standard.Rectangle.prototype.markup,
@@ -378,7 +380,7 @@ export default {
         .getElements()
         .filter(({ component }) => component && component !== this)
         .forEach(({ component }) => {
-          if (component.node.type === 'processmaker-modeler-boundary-timer-event'){
+          if (component.node.type.match(boundaryEventsPattern)){
             return;
           }
           this.shape.embed(component.shape);
--- a/src/components/nodes/pool/pool.vue
+++ b/src/components/nodes/pool/pool.vue
@@ -19,8 +19,6 @@ import { invalidNodeColor, defaultNodeColor, poolColor } from '@/components/node
 import pull from 'lodash/pull';
 import store from '@/store';
 
-const boundaryEventsPattern = /^processmaker-modeler-boundary-\D*-event$/;
-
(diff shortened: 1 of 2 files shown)


tests/e2e/specs/BoundaryEventCommonBehaviour.spec.js
```diff
@@ -84,7 +84,9 @@ boundaryEventData.forEach(({ type, nodeType, eventXMLSnippet }) => {
         .its('xml')
```
lisa-fehr: Might need input on the test. I was able to find the x/y in the code below. But I had no idea how to extract it:

```php
getElementAtPosition(taskPosition).then(getComponentsEmbeddedInShape).then(events => events[0]).then($event => {
  console.log($event);
  console.log($event[0].attributes.transform.textContent; // transform(232,251)
  console.log($event.position()); // no where close to the actual position
});
```

src/components/nodes/pool/pool.vue
```diff
@@ -378,7 +380,7 @@ export default {
         .getElements()
         .filter(({ component }) => component && component !== this)
         .forEach(({ component }) => {
-          if (component.node.type === 'processmaker-modeler-boundary-timer-event'){
+          if (component.node.type.match(boundaryEventsPattern)){
```
chipit24: I think we can check this the same way we check it in `Modeler.vue`, using the BPMN type:
```js
component.node.definition.$type === 'bpmn:BoundaryEvent'
```
This keeps the check consistent (maybe you can pull it out into a util function) and also doesn't require regex.

tests/e2e/specs/BoundaryEventCommonBehaviour.spec.js
chipit24: `getComponentsEmbeddedInShape` returns an array of jQuery elements, not an event. This is a function we defined and you can see it in `/tests/e2e/support/utils.js`.

You can try running [`getBBox`](https://developer.mozilla.org/en-US/docs/Web/API/SVGGraphicsElement/getBBox) on the underlying SVG elements, instead of `position()`. Then you'd have to convert it to paper coordinates for comparison, which you could use our `getPositionInPaperCoords` util function for.

src/components/nodes/pool/pool.vue
JCook21: 💯 on using a util function!

PR #739, opened 2019-10-10
Upgrade screen builder

chipit24:


cypress.json
```diff
@@ -2,5 +2,5 @@
   "pluginsFile": "tests/e2e/plugins/index.js",
   "baseUrl": "http://localhost:8080",
   "video": false,
-  "numTestsKeptInMemory": 0
+  "numTestsKeptInMemory": 25
```
JCook21: Did you mean to leave this in?

JCook21:
@chipit24 I'm seeing the same test failures in a branch where I'm updating dependencies that you're seeing here. It's weird. I've isolated to lines that Lisa added in the boundary event common tests earlier but I have no idea why it passes in develop but not in feature branches.

cypress.json
chipit24: Nope, oops! Will remove.

PR #741, opened 2019-10-11
Boundary event test revision

lisa-fehr:
Take out the hard coded values in the test. Might be more reliable, unless the allowance is off.

**Note:** You can confirm the test fails by commenting out line 381 in pool.vue (there is another line like this one, but it won't fail the test).
```php
if (component.node.definition.$type === 'bpmn:BoundaryEvent'){
  return;
}
```

### DIFF
--- a/tests/e2e/specs/BoundaryEventCommonBehaviour.spec.js
+++ b/tests/e2e/specs/BoundaryEventCommonBehaviour.spec.js
@@ -110,8 +110,6 @@ boundaryEventData.forEach(({ type, nodeType, eventXMLSnippet, eventXMLSnippetWit
           const taskPosition2 = $task.position();
           expect(boundaryPosition2.top).to.be.closeTo(taskPosition2.top, 1);
           expect(boundaryPosition2.left).to.be.closeTo(taskPosition2.left, 88);
-          console.log($anchor.position());
-          console.log($task.position());
         });
       });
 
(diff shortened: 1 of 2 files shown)


tests/e2e/support/constants.js
```diff
@@ -33,5 +33,4 @@ export const nodeTypes = {
 };
 
 export const boundaryEventSelector = '.main-paper ' +
-  '[data-type="processmaker.components.nodes.task.Shape"] + ' +
```
lisa-fehr: After a pool is moved, many embedded objects appear between the task and the boundary event. We had to remove it.

tests/e2e/specs/BoundaryEventCommonBehaviour.spec.js
```diff
@@ -81,26 +81,44 @@ boundaryEventData.forEach(({ type, nodeType, eventXMLSnippet, eventXMLSnippetWit
...
+      moveElementRelativeTo({ x: 400, y: 400 }, 50, 50);
+      waitToRenderAllShapes();
+      cy.get(boundaryEventSelector).then($boundaryEvent => {
+        cy.get(taskSelector).then($task => {
+          const boundaryPosition = $boundaryEvent.position();
+          const taskPosition = $task.position();
+          expect(boundaryPosition.top).to.be.closeTo(taskPosition.top, 1);
+          expect(boundaryPosition.left).to.be.closeTo(taskPosition.left, 88);
+        });
+      });
```
chipit24: Since these lines are the same as the one above, do you think it would make it more readable to extract it out into a function like `testThatBoundaryEventIsCloseToTask`?

lisa-fehr: Can do. I wasn't sure if the 88 would change. Since the one above and the one below aren't the same difference. If it is failing, it is a large difference at least.

PR #742, opened 2019-10-11
Hide label on embedded elements when dragging parent

### LINKED ISSUE #732: When dragging tasks with Boundary Events the Boundary Event label should disappear too
## Description
When you drag a task, the task label disappears. When you drag a task with a boundary event, the task label disappears but the boundary event label remains.
Screen Shot 2019-10-09 at 10 32 01 AM
Screen_Shot_2019-10-09_at_10_32_07_AM__2_

## Expected behaviour
Hide the label on embedded elements when the parent is dragged.

lisa-fehr:
Closes #732

### DIFF
--- a/src/mixins/hideLabelOnDrag.js
+++ b/src/mixins/hideLabelOnDrag.js
@@ -7,9 +7,11 @@ export default {
       }
 
       this.shape.attr('label/display', 'none');
-      this.shape.getEmbeddedCells().forEach(element => {
-        element.attr('label/display', 'none');
-      });
+      this.shape.getEmbeddedCells()
+        .filter(element => element.attr('label/text'))
+        .forEach((element) => {
+          element.attr('label/display', 'none');
+        });
     },
     async showLabel() {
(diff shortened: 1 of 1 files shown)


src/mixins/hideLabelOnDrag.js
```diff
@@ -7,6 +7,9 @@ export default {
       }
 
       this.shape.attr('label/display', 'none');
+      this.shape.getEmbeddedCells().forEach(element => {
+        element.attr('label/display', 'none');
```
lisa-fehr: Could limit this to boundary events

lisa-fehr: Guess I'd have to do it with a filter on:
```
element.attributes.type === 'processmaker.components.nodes.boundaryEvent.Shape'
```

For now I just hid label if it had text.

lisa-fehr:
@JCook21 I added one more change, so it won't alter generic embedded shapes. I think it is in a better place now.

PR #747, opened 2019-10-16
Add Intermediate Message Throw Event

### LINKED ISSUE #744: Add intermediate message throw event
The intermediate message throw event should generate the following BPMN XML:
```xml
<bpmn:intermediateThrowEvent id="node_1">
</bpmn:intermediateThrowEvent>
```

And should look like this:
Screen Shot 2019-10-15 at 1 31 34 PM

All other properties should be identical to the intermediate message catch event.

lisa-fehr:
Closes #744 

Shared common code with Intermediate Message Catch Event. Can revert if needed.


chipit24:
This looks good 🔥

@lisa-fehr @tuantruong625 @JCook21 I had a thought when reading through the code: I think we should shift away from using `extends` and try and think in terms of component composition, for which the end result may look something like this:
```html
<!-- intermediateMessageThrowEvent.vue -->
<template>
  <intermediate-message-event :icon="intermediateMailSymbol" />
</template>
```

There's an article in the react docs about this: https://reactjs.org/docs/composition-vs-inheritance.html. I'm not sure if I've mentioned that before, or how easy it might be since JointJS is doing the view rendering, and not necessarily Vue. The same argument goes for `mixins`.

src/setup/defaultNodes.js
```diff
@@ -48,6 +49,7 @@ const timerEventNodes = [
   [startTimerEvent, 'bpmn:StartEvent', 'bpmn:TimerEventDefinition'],
   [intermediateTimerEvent, 'bpmn:IntermediateCatchEvent', 'bpmn:TimerEventDefinition'],
   [intermediateMessageCatchEvent, 'bpmn:IntermediateCatchEvent', 'bpmn:MessageEventDefinition'],
+  [intermediateMessageThrowEvent, 'bpmn:IntermediateThrowEvent', 'bpmn:MessageEventDefinition'],
```
JCook21: Can you think of a better name for this array? It's currently called `timerEventNodes` but it doesn't only contain timer events...

lisa-fehr: Good point. I put it in with Catch Events, but it was in the timerEventNodes.

lisa-fehr: @JCook21 I used `customEventNodes`. It matches the wording below for `customParserFactory`.

PR #751, opened 2019-10-17
Add message ref to intermediate message throw event

### LINKED ISSUE #746: Add messageRef attribute to intermediate message throw event
Add a text input in the inspector to specify the message to be thrown:
Screen Shot 2019-10-15 at 1 33 10 PM

lisa-fehr:
Closes #746

Added unit and cypress tests.

Some odd commits got in here after branching off #747, so it needs to be squashed.


src/components/nodes/intermediateMessageThrowEvent/index.js
```diff
@@ -13,8 +13,28 @@ export default merge(cloneDeep(intermediateMessageEventConfig), {
...
+    {
+      items: [
+        {
+          items: [
+            {},
+            {
+              component: 'FormInput',
+              config: {
+                label: 'Message Name',
+                name: 'messageRef',
```
lisa-fehr: Need to get this on the `MessageEventDefinition`

PR #758, opened 2019-10-22
Add message start event

### LINKED ISSUE #737: Add Message Start Event
The message start event should look like this:
Screen Shot 2019-10-10 at 2 39 26 PM

```xml
<bpmn:startEvent id="node_1">
</bpmn:startEvent>
```

Note:
 * It should be green like the other start events.
 * The message start event should have the same _Listen For Messages_ inspector config as the intermediate message catch event.

chipit24:
Fixes #737.


src/assets/toolpanel/Page 1.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="10" viewBox="0 0 16 10" fill="none" xmlns="http://www.w3.org/2000/svg">
```
lisa-fehr: Should the spaces in the filenames be removed? Did you mean to commit the 2 extra files in toolpanel?

src/assets/toolpanel/message-start-event.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="22" height="22" viewBox="0 0 22 22" fill="none" xmlns="http://www.w3.org/2000/svg">
```
lisa-fehr: end-email-event.svg looks like this:
Screen Shot 2019-10-23 at 11 14 52 AM


Should the envelops in the toolpanel be updated to look all the same?

src/assets/toolpanel/Page 1.svg
chipit24: Didn't mean to add this, will remove, thanks!

src/assets/toolpanel/message-start-event.svg
chipit24: Good point; I can update it to keep this consistent.

PR #762, opened 2019-10-23
Added Message End Event

### LINKED ISSUE #738: Add Message End Event
The message and event should look like this:
Screen Shot 2019-10-10 at 2 46 03 PM

```xml
<bpmn:endEvent id="node_1">
</bpmn:endEvent>
```

lisa-fehr:
Closes #738 

Don't think I can add much of a test without a `Message Start Event`

@/assets/toolpanel/end-email-event.svg already existed. I'm not sure if it was supposed to be for something else, but it was unused.


src/components/nodes/endEvent/index.js
```diff
@@ -25,11 +25,6 @@ export default {
       }),
     });
   },
-  /**
-   * Validate whether to accept an outgoing flow to the node
-   *
-   * @param node
-   */
   validateOutgoing() {
```
JCook21: Is there any way this method name can be changed to match the comment? `allowOutgoingFlow()` would be a bette name IMO.

lisa-fehr: I might have to change unrelated code in sequence and message flow.

chipit24:
@lisa-fehr This seems to be missing the other 1/2 of the functionality mentioned in the issue:

> Since this is a throwing event, it will have the same message behaviour as the intermediate message throw event.

I'm expecting it to have the _Message Name_ inspector input.

lisa-fehr:
@chipit24 oops. I was looking at the main issue content.

lisa-fehr:
Tried to balance out placement of the new Message End Event so it is closer to the other End Event. Like Message Start Event is close to other Start Events.
Screen Shot 2019-10-23 at 2 22 37 PM

Issue #755, opened 2019-10-21
Add Error End Event


chipit24:
The error end event “throws” an error, which can be captured in another part of the process. 

The error end event should look like this:
![Screen Shot 2019-10-21 at 1 38 03 PM](https://user-images.githubusercontent.com/7561061/67228792-07d7a700-f408-11e9-929c-2500e86d99a9.png)

And should produce the following BPMN XML:
```xml
<bpmn:endEvent id="node_1">
  <bpmn:errorEventDefinition errorRef="node_1_error" />
</bpmn:endEvent>

<bpmn:error id="node_1_error" name="Cool error" />
```

Since this is a throwing event, it should share the same basic functionality as the intermediate message throw event:
 * A `bpmn:error` element is created when the error end event is dragged into the grid.
 * The name of the error element can be modified through the Error Name inspector input.
 * The error element is deleted when the error end event is deleted.

PR #764, opened 2019-10-23
Add Error End Event

chipit24:
Fixes https://github.com/ProcessMaker/modeler/issues/755.


tests/e2e/support/utils.js
```diff
@@ -234,7 +234,7 @@ export function getNumberOfLinks() {
     .then(({ graph }) => graph.getLinks().length);
 }
 
-export function assertDownloadedXMLMatchesExpected(xmlString) {
+export function assertDownloadedXmlContainsExpected(xmlString) {
```
chipit24: I renamed this to use `Xml` instead of `XML`, as other functions use that casing. I also changed `Matches` to `Contains`, as it better reflects what the method is doing.

tests/e2e/support/utils.js
```diff
@@ -243,3 +243,13 @@ export function assertDownloadedXMLMatchesExpected(xmlString) {
...
+
+export function assertDownloadedXmlDoesNotContainExpected(xmlString) {
+  cy.get('[data-test=downloadXMLBtn]').click();
+  cy.window()
+    .its('xml')
+    .then(removeIndentationAndLinebreaks)
+    .then(xml => {
+      expect(xml).to.not.contain(removeIndentationAndLinebreaks(xmlString));
+    });
+}
```
chipit24: This method is the exact same as the one above it, with the simple addition of a `.not` to the assertion. Not sure if it's worth reducing the duplication for this.

tests/e2e/support/utils.js
lisa-fehr: Aww. I used this in the Message End Event branch.

src/setup/registerNodes.js
```diff
@@ -48,6 +49,7 @@ const nodeTypes = [
 const customEventNodes = [
   [startTimerEvent, 'bpmn:StartEvent', 'bpmn:TimerEventDefinition'],
   [messageStartEvent, 'bpmn:StartEvent', 'bpmn:MessageEventDefinition'],
+  [errorEndEvent, 'bpmn:EndEvent', 'bpmn:ErrorEventDefinition'],
```
lisa-fehr: Can it go at the bottom of the array. So it is closer to the other end events?

src/components/nodes/endEvent/endEvent.vue
```diff
@@ -23,9 +23,8 @@ export default {
     },
   },
   mounted() {
-    // Now, let's add a rounded rect to the graph
-    this.shape = new shapes.standard.Circle();
-    let bounds = this.node.diagram.bounds;
+    this.shape = new StartEventShape();
```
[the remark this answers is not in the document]
chipit24: It's no longer only a start event shape, so I can re-name it to be more generic.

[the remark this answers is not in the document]
chipit24: I changed the name, but I kept it in the same folder as I'm not sure what a better place for it might be.

PR #768, opened 2019-10-25
Dashes in process id break call activity

### LINKED ISSUE #767: Call Activity breaks when selecting process with dash in ID
## Description
When a process's ID contains a dash, the Call Activity is unable to correctly parse the ID of that process.

## Reproduction steps
Drag a Call Activity onto the grid, select it, and in the Process dropdown in the inspector config, select a process which is known to contain a dash (`-`) in its ID.

## Expected behaviour
The process is selected and its ID is assigned to the call activity's `calledElement` attribute.

## Actual behaviour
Selecting the process throws the following error in the console:
```
[Vue warn]: Error in callback for watcher "node.definition.calledElement": "TypeError: Cannot read property 'name' of undefined"
```

## Software used
Latest modeler `develop` branch, Chrome 78, maxOS Mojave 10.14.6.

## Other information
The relevant code is:
https://github.com/ProcessMaker/modeler/blob/2374a7f1136ae6e4484eb46e0a26303a009584c5/src/components/nodes/callActivity/callActivity.vue#L57

lisa-fehr:
Closes #767 

Split out last dashed digit from the rest of the process id

### DIFF
--- a/src/components/nodes/callActivity/callActivity.vue
+++ b/src/components/nodes/callActivity/callActivity.vue
@@ -1,5 +1,5 @@
 <template>
-  <div/>
+  <div />
 </template>
 
 <script>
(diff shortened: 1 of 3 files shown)


chipit24:
With this being fixed locally, I wonder if the backend will have any issues parsing the `calledElement` value if it contains multiple dashes in the ID 🤔 @caleeli do you have any thoughts on this?

JCook21:
Only thing that I can think of is can we add a test for this case? If you can extract this logic it would be perfect for a unit test. This would then give us a guarantee that this bug cannot be re-introduced.

PR #776, opened 2019-10-30
Fix invalid pool colour bug

### LINKED ISSUE #775: Invalid pool colour does not reset when dragging boundary event
# Bug report issue

## Description
When dragging a boundary event, the target of the boundary event turns red if it's invalid. When the boundary event is no longer over the invalid target, the target should return to its default colour. However, when the target is a pool, it does not return to its default colour, but remains red.

## Reproduction steps
1. Drag a pool into the grid.
2. Drag a task onto the pool.
3. Drag a boundary timer event or a boundary error event onto the task.
4. Move the boundary event over the pool until the pool turns red, then release the boundary event.

Note: The order here matters, as dragging the task on boundary event on first, and then the pool, seems to work as expected.

## Expected behaviour
The pool should return to its default colour (grey) when the boundary event is released.

## Actual behaviour
The pool remains red even though the boundary event is no longer being hovered over it.

## Software used
Chrome 78 running on macOS Mojave 10.14.6

chipit24:
Fixes #775.

### DIFF
--- a/src/components/nodes/boundaryEvent/boundaryEvent.vue
+++ b/src/components/nodes/boundaryEvent/boundaryEvent.vue
@@ -125,11 +125,6 @@ export default {
     resetToInitialPosition() {
       this.shape.position(this.validPosition.x, this.validPosition.y);
       this.allowSetNodePosition = true;
-
-      if (this.invalidTargetElement) {
-        resetShapeColor(this.invalidTargetElement);
-        this.invalidTargetElement = null;
-      }
     },
     moveBoundaryEventIfOverTask() {
       const task = this.getTaskUnderShape();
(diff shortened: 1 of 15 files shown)


JCook21:
@chipit24 is this fixed by #779 instead? If so can you close this one?

lisa-fehr:
@JCook21 I checked this bug in #779 and it was not fixed there.

src/components/nodes/boundaryEvent/boundaryEvent.vue
```diff
@@ -125,6 +125,11 @@ export default {
     resetToInitialPosition() {
       this.shape.position(this.validPosition.x, this.validPosition.y);
       this.allowSetNodePosition = true;
+
+      if (this.invalidTargetElement) {
+        resetShapeColor(this.invalidTargetElement);
+        this.invalidTargetElement = null;
+      }
```
chipit24: This code ensures that any previously invalid element has it's colour reset when the boundary event returns to its initial position.

When working with pools, the pool colour wouldn't reset from red to the default pool colour because the logic for handling the colour change didn't detect that the target of the boundary event has changed, as in every case it appeared to be over the pool.

chipit24:
There's a single test failing, "_retains outgoing sequence flows on Boundary Events_", yet all I did was revert the unneeded lockfile change. These test pass locally, so I'll re-run the tests on CI and see if it passes.

PR #778, opened 2019-11-01
forEach error on validation

### LINKED ISSUE #777: forEach error on autovalidate
# Bug report issue
## Description
forEach error in the validation log appears at times
## Reproduction steps
Upload https://github.com/ProcessMaker/modeler/files/3617104/bugProcess.bpmn.txt
Click Auto Validate
## Expected behaviour
There should be no forEach errors
## Actual behaviour
There are forEach errors in the problems log
## Other information
Screen Shot 2019-10-31 at 2 24 11 PM

Seems to be due to `this.removeUnsupportedElementAttributes(definition);` in the Modeler.vue
It is checking for `Documentation` even though it is not in the xml.
Screen Shot 2019-10-31 at 1 56 48 PM
Screen Shot 2019-10-31 at 2 01 05 PM

lisa-fehr:
Closes #777 

When validating, it was looking through a list on the $descriptor.properties. Since we removed documentation and extensionElements from definitions it was able to pass some of the if conditions (`p.name in element` and `p.isMany`), but not find the propertyValue (`element[p.name]`).
Screen Shot 2019-10-31 at 4 38 17 PM

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -132,7 +132,6 @@ import registerInspectorExtension from '@/components/InspectorExtensionManager';
 import initAnchor from '@/mixins/linkManager.js';
 
 const version = '1.0';
-const unsupportedElements = ['documentation', 'extensionElements'];
 
 export default {
   components: {
@@ -514,11 +513,13 @@ export default {
       store.commit('highlightNode', this.processNode);
     },
     removeUnsupportedElementAttributes(definition) {
+      const unsupportedElements = ['documentation', 'extensionElements'];
+
       unsupportedElements.filter(name => definition.get(name))
         .forEach(name => definition.set(name, null));
 
       definition.$descriptor.properties = definition.$descriptor.properties.filter(
-        p => unsupportedElements.indexOf(p.name) === -1);
+        p => !unsupportedElements.includes(p.name));
     },
     setNode(definition, flowElements, artifacts) {
       /* Get the diagram element for the corresponding flow element node. */
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -464,6 +464,7 @@ export default {
         }
 
         /* Add all other elements */
+
         const flowElements = process.get('flowElements');
         const artifacts = process.get('artifacts');
 
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -132,6 +132,7 @@ import registerInspectorExtension from '@/components/InspectorExtensionManager';
 import initAnchor from '@/mixins/linkManager.js';
 
 const version = '1.0';
+const unsupportedElements = ['documentation', 'extensionElements'];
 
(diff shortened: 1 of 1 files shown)


src/components/Modeler.vue
```diff
@@ -514,9 +514,11 @@ export default {
       store.commit('highlightNode', this.processNode);
     },
     removeUnsupportedElementAttributes(definition) {
-      const unsupportedElements = ['documentation', 'extensionElements'];
       unsupportedElements.filter(name => definition.get(name))
         .forEach(name => definition.set(name, null));
+
+      definition.$descriptor.properties = definition.$descriptor.properties.filter(
+        p => unsupportedElements.indexOf(p.name) === -1);
```
JCook21: I think this could be a little clearer as follows:
```suggestion
        p => !unsupportedElements.includes(p.name));
```

src/components/Modeler.vue
```diff
@@ -132,6 +132,7 @@ import registerInspectorExtension from '@/components/InspectorExtensionManager';
 import initAnchor from '@/mixins/linkManager.js';
 
 const version = '1.0';
+const unsupportedElements = ['documentation', 'extensionElements'];
```
JCook21: I'm not sure about moving this constant to here. In general I think variables should be declared as close to where they're used as possible. I think this array is only used in the single method below, right?

src/components/Modeler.vue
```diff
@@ -515,8 +515,12 @@ export default {
     },
     removeUnsupportedElementAttributes(definition) {
       const unsupportedElements = ['documentation', 'extensionElements'];
+
       unsupportedElements.filter(name => definition.get(name))
         .forEach(name => definition.set(name, null));
```
chipit24: It seems that the `documentation` attribute has the "isMany" property set to `true`, and so _bpmn-moddle_ expects it to have a default value of `[]` instead of `null`. I've noticed that when I set the value to `undefined` instead, it correctly sets it to the expected default value.

Digging into the source code, this seems to be the case:
```js
if (isUndefined(value)) {
  // unset the property, if the specified value is undefined;
  // delete from $attrs (for extensions) or the target itself
  if (property) {
    delete target[propertyName];
  } else {
    delete target.$attrs[name];
  }
}
// ...
function isUndefined(val) {
  return typeof val === 'undefined';
}
```

I've made this change locally and it seems to fix the problem, without having to modify `definition.$descriptor.properties`.

chipit24:
```suggestion
        .forEach(name => definition.set(name, undefined));
```

chipit24: Also part of this change, but GitHub currently doesn't support multi line suggestions: Remove the additional filter:
```js
definition.$descriptor.properties = definition.$descriptor.properties.filter(
        p => !unsupportedElements.includes(p.name));
```

PR #779, opened 2019-11-01
Shapes with background color can lose that color on valid or invalid states

### LINKED ISSUE #763: 'Cannot do' red indicator removes background color on elements with a background color
# Bug report issue
## Description
If you try to attach a message flow to a start, end or message end event, the background turns red and then white.
## Reproduction steps
* Place a start, end, and message end event on the paper
* Place a task on the paper
* Try to connect a message flow from the task to the start, end and message end events.
## Expected behaviour
The background on the start event should return to green. The backgrounds on the end and message end events should return to a light red.
## Actual behaviour
The start, end and message end event backgrounds change to white.
## Software used
Lastest Chrome. Modeler version: 0.17.2
## Other information
**Video:** https://www.dropbox.com/s/tsg9xjdwmn3vbd9/backgroundColorDisappears.mov?dl=0

**Note:** The color change is subtle and might not be noticed on some screens.

**Before:**
Screen Shot 2019-10-23 at 10 29 20 AM
**After:**
Screen Shot 2019-10-23 at 10 27 58 AM

lisa-fehr:
Closes #763

**Additionally solved:**

When you drag a boundary event onto a start or end event, the start or end event would lose their background color.

Drag a boundary event onto a start and end event. They should retain their background colors.
Screen Shot 2019-11-01 at 2 38 52 PM

### DIFF
--- a/src/components/nodes/boundaryEvent/boundaryEvent.vue
+++ b/src/components/nodes/boundaryEvent/boundaryEvent.vue
@@ -150,9 +150,6 @@ export default {
         this.allowSetNodePosition = true;
       });
     },
-    isPoolShape(model) {
-      return model.component.node.type === 'processmaker-modeler-pool';
-    },
     turnInvalidTargetRed() {
       if (!this.highlighted) {
         return;
@@ -179,8 +176,7 @@ export default {
       this.invalidTargetElement = targetElement;
     },
     resetShapeColor(shape) {
-      const defaultColor = this.isPoolShape(shape) ? poolColor : defaultNodeColor;
-      shape.attr('body/fill', shape.attr('body/originalFill') || defaultColor);
+      shape.attr('body/fill', shape.attr('body/originalFill') || defaultNodeColor);
     },
   },
   async mounted() {
--- a/src/components/nodes/boundaryEvent/boundaryEvent.vue
+++ b/src/components/nodes/boundaryEvent/boundaryEvent.vue
@@ -180,7 +180,7 @@ export default {
     },
     resetShapeColor(shape) {
       const defaultColor = this.isPoolShape(shape) ? poolColor : defaultNodeColor;
-      shape.attr('body/fill', defaultColor);
+      shape.attr('body/fill', shape.attr('body/originalFill') || defaultColor);
     },
   },
   async mounted() {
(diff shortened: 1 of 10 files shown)


tests/e2e/specs/SequenceFlows.spec.js
```diff
@@ -233,4 +234,23 @@ describe('Sequence Flows', () => {
 
     cy.get('.main-paper [data-type="standard.Link"] [joint-selector="line"]').should('have.attr', 'd', 'M 308 326 L 308 450');
   });
+
+  it('retains original background color when it can and cannot connect to an element', () => {
```
JCook21: Is this test doing two things? If so how do we know what has failed when it fails? I would prefer to see this broken out into separate tests.

src/mixins/linkConfig.js
```diff
@@ -235,6 +220,16 @@ export default {
...
     },
+    isPoolOrLane(model) {
+      if (!model.component) {
+        return false;
+      }
+      return ['processmaker-modeler-lane', 'processmaker-modeler-pool'].includes(model.component.node.type);
+    },
+    resetBodyColor(shape) {
+      const defaultColor = this.isPoolOrLane(shape) ? poolColor : defaultNodeColor;
+      this.setBodyColor(shape.attr('body/originalFill') || defaultColor, shape);
```
JCook21: This seems to be a direct copy and paste from boundaryEvent.vue. Can we avoid that?

src/components/nodes/boundaryEvent/boundaryEvent.vue
```diff
@@ -180,7 +180,7 @@ export default {
     },
     resetShapeColor(shape) {
       const defaultColor = this.isPoolShape(shape) ? poolColor : defaultNodeColor;
-      shape.attr('body/fill', defaultColor);
+      shape.attr('body/fill', shape.attr('body/originalFill') || defaultColor);
```
JCook21: I'd love to see this simplified. With this change we have three ways of arriving at the default colour: the pool colour, the default colour or the 'originalFill' property. Could we add an originalFill property to the poolShape? This could then be simplified to just this line:
```suggestion
      shape.attr('body/fill', shape.attr('body/originalFill') || defaultColor);
```
What do you think?

JCook21: I'd love to see this simplified. With this change we have three ways of arriving at the default colour: the pool colour, the default colour or the 'originalFill' property. Could we add an originalFill property to the poolShape? This could then be simplified to just this line:
```suggestion
      shape.attr('body/fill', shape.attr('body/originalFill') || defaultColor);
```
What do you think?

Issue #782, opened 2019-11-04
Add ability to collapse and expand right (inspector config) sidebar


chipit24:
Add a button to collapse the inspector panel.

![Screen Shot 2019-11-04 at 2 04 21 PM](https://user-images.githubusercontent.com/7561061/68149563-15af2100-ff0c-11e9-83b4-4723e1eb5144.png)

Questions:
* How should the collapse of the panel be initiated? Should there be a button to the right of the panel? **Answered** in #781 
There is one icon to collapse, it should be in the bar to the left of the mini-map. 
> Upon click on the icon, the inspector should collapse completely, the tools should just show icons.

* Should it be collapsed by default? **Answered** in #781 

> By default sidebars are expanded.

Acceptance criteria:
 * You should still be able to drop elements onto the canvas in the extra space produced by the collapsed panel. The grid in this area should also still be visible and interactive.
* The panel should be completely hidden when collapsed.

PR #784, opened 2019-11-05
Collapse control panel

### LINKED ISSUE #781: Add ability to collapse and expand left (element controls) sidebar
Add a button to collapse the element controls sidebar.

Image from iOS

Questions:
* How should the collapse of the panel be initiated? Should there be a button to the right of the panel?
* Should the filter input be hidden or disabled when the sidebar is collapsed?
* Should it be collapsed by default?

Acceptance criteria:
* The element controls panel will remain expanded by default (we can assume for now to get started).
* It should still be possible to drag elements onto the canvas with the panel collapsed.
* You should still be able to drop elements onto the canvas in the extra space produced by the collapsed panel. The grid in this area should also still be visible and interactive.
* You should still be able to scroll the list of controls if it overflows the bottom of the view.
* Tooltips should still appear when hovering over a control.

Note: In the screenshot the collapsed panel was implemented by reducing the width of the panel container.

tuantruong625:
The purpose of this PR is to give users the feature to collapse the controls panel to display the icons only.

Note: The filter is hidden when the user selects to display icons only.

Screen Shot 2019-11-05 at 2 56 43 PM

Todo: ~Fix layout issue with tooltip in processmaker~, remove `...` when collapsed.

Fixes #781

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -7,10 +7,10 @@
       :title="tooltipTitle"
     />
 
-    <b-col class="h-100 overflow-hidden controls-column" :class="[{ 'ignore-pointer': canvasDragPosition, 'controls-column-compressed' : !panelsOpen }]" data-test="controls-column">
+    <b-col class="h-100 overflow-hidden controls-column" :class="[{ 'ignore-pointer': canvasDragPosition, 'controls-column-compressed' : panelsCompressed }]" data-test="controls-column">
       <controls
         :controls="controls"
-        :panelsOpen="panelsOpen"
+        :panelsCompressed="panelsCompressed"
         :style="{ height: parentHeight }"
         :invalidDrop="validateDropTarget"
         :allowDrop="allowDrop"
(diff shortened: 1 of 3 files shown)


src/components/Modeler.vue
```diff
@@ -42,10 +43,18 @@
           <span class="btn btn-sm btn-secondary scale-value">{{ Math.round(scale*100) }}%</span>
         </div>
 
-        <button class="btn btn-sm btn-secondary mini-map-btn ml-auto" data-test="mini-map-btn" @click="miniMapOpen = !miniMapOpen">
-          <font-awesome-icon v-if="miniMapOpen" :icon="minusIcon" />
-          <font-awesome-icon v-else :icon="mapIcon" />
-        </button>
+        <div class="btn-group btn-group-sm mr-2" role="group" aria-label="Third group">
+          <button class="btn btn-sm btn-secondary ml-auto" data-test="panels-btn" @click="panelsCompressed = !panelsCompressed">
+            <font-awesome-icon v-if="panelsCompressed" :icon="expandIcon" />
```
lisa-fehr: Noticed this should be 
```suggestion
            <font-awesome-icon v-if="panelsCompressed" :icon="compressIcon" />
            <font-awesome-icon v-else :icon="expandIcon" />
```

src/components/Modeler.vue
```diff
@@ -42,10 +43,18 @@
...
+            <font-awesome-icon v-else :icon="compressIcon" />
```
JCook21: Same comment as in Lisa's review: this can be made into a ternary operator and collapsed into a single element.

src/components/Modeler.vue
```diff
@@ -42,10 +43,18 @@
...
+          </button>
+
+          <button class="btn btn-sm btn-secondary mini-map-btn ml-auto" data-test="mini-map-btn" @click="miniMapOpen = !miniMapOpen">
+            <font-awesome-icon v-if="miniMapOpen" :icon="minusIcon" />
+            <font-awesome-icon v-else :icon="mapIcon" />
```
JCook21: Same as above: a ternary operator would take away the need for two elements here.

src/components/Modeler.vue
lisa-fehr: Changes made in #782 
Might be easier to merge into this pr after that one is in develop?"

PR #785, opened 2019-11-05
Highlight dropped element automatically

### LINKED ISSUE #783: Highlight element automatically when dropped onto the grid
When an element is dragged onto the grid, it should be highlighted by default.

Acceptance criteria:
* Highlighted by default without having to select it manually.
* The crown config should be open.
* The inspector panel should have the element's config loaded.
* Clicking away should de-select the element as normal.

chipit24:
Fixes #783.

### DIFF
--- a/src/mixins/crownConfig.js
+++ b/src/mixins/crownConfig.js
@@ -55,10 +55,13 @@ export default {
       allowSetNodePosition: true,
       savePositionOnPointerupEventSet: false,
       shape: null,
+      crownConfigured: false,
     };
   },
   watch: {
     highlighted(highlighted) {
+      this.configureCrown();
+
       if (highlighted) {
         this.shapeView.highlight(this.shapeBody, { highlighter: defaultHighlighter });
         this.addCrown();
@@ -188,6 +191,12 @@ export default {
       });
     },
     configureCrown() {
+      if (this.crownConfigured) {
+        return;
+      }
+
+      this.crownConfigured = true;
+
       if (!this.crownConfig) {
         this.crownConfig = [];
       }
--- a/src/mixins/crownConfig.js
+++ b/src/mixins/crownConfig.js
@@ -55,13 +55,10 @@ export default {
       allowSetNodePosition: true,
       savePositionOnPointerupEventSet: false,
       shape: null,
-      crownConfigured: false,
     };
   },
   watch: {
     highlighted(highlighted) {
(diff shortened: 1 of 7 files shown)


src/mixins/crownConfig.js
```diff
@@ -188,6 +191,12 @@ export default {
       });
     },
     configureCrown() {
+      if (this.crownConfigured) {
+        return;
+      }
+
+      this.crownConfigured = true;
+
```
JCook21: This seems a little weird. What do these lines achieve and why are they needed?

tests/unit/nodeManager.spec.js
```diff
@@ -0,0 +1,143 @@
+import { addIdToNodeAndSetUpDiagramReference, addNodeToProcess, getTargetProcess } from '@/components/nodeManager';
```
JCook21: I really like that you've done the code extraction and added unit tests! Great job. 💯

src/mixins/crownConfig.js
chipit24: It's to prevent this function from running twice. Running it twice duplicates the crown buttons. Without highlight on drop, this method was only called once when the element was clicked. But with the crown already visible on drop, I don't want this function called again if a user clicks the element.

I'll see if I can clean this up.

chipit24: I removed `crownConfigured` and made a single call to `configureCrown` within `setUpCrownConfig`. This may have some performance implications (as the call to `configureCrown` was deferred to help with performance on initial load), but as long as tests continue to pass, performance issues will just have to be re-assessed in the future; I think this is a fair trade-off.

package.json
```diff
@@ -118,11 +118,10 @@
   ],
   "nyc": {
     "extension": [
-      ".js",
       ".vue"
     ],
-    "exclude": [
-      "tests/**/*.js"
+    "include": [
+      "src/**"
```
JCook21: Why is this change needed?

chipit24:
FYI: I force pushed to remove the test commits I pushed up and to pull in the latest changes. The code should be the exact same as when initially reviewed.

chipit24:
Added WIP label as I work on getting this feature to work in Safari.

chipit24:
Tested in processmaker core on macOS in Chrome and Safari, and in Edge on Windows.

PR #786, opened 2019-11-06
Update dependencies

JCook21:
This is an update to the modeler dependencies in an effort to help with packaging the modeler using composer. Some further work will need to be done in core when the next version of modeler is released to update dependencies like jointjs there.


src/components/Modeler.vue
```diff
@@ -215,6 +215,7 @@ export default {
       if (this.tooltipTarget) {
         return this.tooltipTarget.$el.data('title');
       }
+      return undefined;
```
JCook21: After updates to eslint this was flagged as a compilation error since computed properties need to return a value.

src/components/controls.vue
```diff
@@ -15,7 +15,6 @@
 
     <b-list-group flush class="overflow-auto w-auto">
       <b-list-group-item v-for="(control, index) in controlItems"
-        v-if="control.label.toLowerCase().includes(filterQuery.toLowerCase())"
```
JCook21: This was flagged by eslint as an error following the updates. The suggestion was to fix the computed property to do the filtering for us. Since this is the only place that property is used the fix below was simple to make.

babel.config.js
```diff
@@ -3,7 +3,7 @@ module.exports = function(api) {
 
   return {
     presets: ['@vue/app'],
-    plugins: ['istanbul', '@babel/plugin-proposal-private-methods'],
+    plugins: ['@babel/plugin-proposal-private-methods'],
```
JCook21: It seems that `Istanbul` is installed along with `vue-cli-utils` now. Leaving it here was causing a duplicate plugin warning and causing the unit tests to not run. The only solution was remove it from this array.

PR #787, opened 2019-11-06
Collapse / Expand inspector panel

### LINKED ISSUE #782: Add ability to collapse and expand right (inspector config) sidebar

lisa-fehr:
Closes #782 

This has a copy of code from #784

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -58,24 +58,22 @@
 
     <mini-paper :isOpen="miniMapOpen" :paperManager="paperManager" :graph="graph" />
 
-    <transition name="inspector">
-      <b-col
-        v-show="!panelsCompressed"
-        class="pl-0 h-100 overflow-hidden inspector-column"
-        :class="[{ 'ignore-pointer': canvasDragPosition, 'inspector-column-compressed' : panelsCompressed }]"
-        data-test="inspector-column"
-      >
-        <InspectorPanel
-          ref="inspector-panel"
-          :style="{ height: parentHeight }"
-          :nodeRegistry="nodeRegistry"
-          :moddle="moddle"
-          :processNode="processNode"
-          @save-state="pushToUndoStack"
-          class="h-100"
-        />
-      </b-col>
-    </transition>
+    <b-col
+      v-show="!panelsCompressed"
+      class="pl-0 h-100 overflow-hidden inspector-column"
+      :class="[{ 'ignore-pointer': canvasDragPosition, 'inspector-column-compressed' : panelsCompressed }]"
+      data-test="inspector-column"
+    >
+      <InspectorPanel
+        ref="inspector-panel"
+        :style="{ height: parentHeight }"
+        :nodeRegistry="nodeRegistry"
+        :moddle="moddle"
+        :processNode="processNode"
+        @save-state="pushToUndoStack"
+        class="h-100"
+      />
+    </b-col>
     <component
       v-for="node in nodes"
       :is="node.type"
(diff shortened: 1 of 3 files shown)


src/components/Modeler.vue
```diff
@@ -42,29 +42,42 @@
           <span class="btn btn-sm btn-secondary scale-value">{{ Math.round(scale*100) }}%</span>
         </div>
 
-        <button class="btn btn-sm btn-secondary mini-map-btn ml-auto" data-test="mini-map-btn" @click="miniMapOpen = !miniMapOpen">
-          <font-awesome-icon v-if="miniMapOpen" :icon="minusIcon" />
-          <font-awesome-icon v-else :icon="mapIcon" />
-        </button>
+        <div class="btn-group btn-group-sm mr-2" role="group" aria-label="Third group">
```
chipit24: Nitpick: Instead of having the aria labels be "First group", "Second group" and "Third group", I think more meaningful names like "Undo/redo controls", "Zoom controls" and "Additional controls" would be better.

src/components/Modeler.vue
```diff
@@ -42,29 +42,42 @@
...
+          <button class="btn btn-sm btn-secondary ml-auto" data-test="panels-btn" @click="panelsCompressed = !panelsCompressed">
+            <font-awesome-icon v-if="panelsCompressed" :icon="expandIcon" />
+            <font-awesome-icon v-else :icon="compressIcon" />
```
chipit24: These two could be combined into one element, no?
```html
<font-awesome-icon :icon="panelsCompressed  ? expandIcon : compressIcon" />
```

src/components/Modeler.vue
```diff
@@ -243,6 +257,12 @@ export default {
     minusIcon() {
       return faMinus;
     },
+    expandIcon() {
+      return faExpand;
+    },
+    compressIcon() {
+      return faCompress;
+    },
```
chipit24: These don't need to be computed properties; they will never change. They can go into the `data` object.

src/components/Modeler.vue
JCook21: In fairness, I think this was copied from the lines 52 and 53 below. Maybe that could be edited too to use a ternary operator?

tests/e2e/specs/Modeler.spec.js
```diff
@@ -405,4 +405,25 @@ describe('Modeler', () => {
       .find('>text')
       .should('have.css', 'display', 'block');
   });
+
+  it('can collapse and expand inspector panel', () => {
```
JCook21: Can we break this into two separate tests? A failure here would make it harder to understand if collapsing or expanding the panels is broken...

chipit24:
@lisa-fehr I think this is good to merge, and https://github.com/ProcessMaker/modeler/pull/784 can be updated with this code, as you suggested.

PR #794, opened 2019-11-11
Fail build if peer deps not met

### LINKED ISSUE #796: Fail build if peer dependencies are invalid

JCook21:
Fixes #796.

This PR introduces a new npm command, `check-peer-deps`. This command simply runs `npm ls`, which exist with code 1 if peer dependencies are not met. This command has also been added as a pre-step to `npm run test-ci`, which will cause CircleCI builds to break if peer dependencies are not in order. I also updated i18next to be consistent with Screen Builder and took the opportunity to update Cypress to the latest version.


lisa-fehr:
I get the same error locally as CircleCI does.

package.json
```diff
@@ -13,6 +13,7 @@
     "lint-fix": "vue-cli-service lint",
     "pretest": "rm -rf .nyc_output || true",
     "test": "npm run test-unit && npm run test-ci",
+    "check-peer-deps": "npm ls",
```
chipit24: It seems like `npm ls` generates a very long output; you could set the [`depth`](https://docs.npmjs.com/cli/ls.html#depth) to limit this, and it seems the behaviour would still be the same:
```console
npm ls --depth=0
```

PR #795, opened 2019-11-11
Update crown config

### LINKED ISSUE #789: Update crown config styling using HTML and CSS.
Acceptance criteria:
* Update crown config to use html & css
* Add colour background to crown config (#5096db)
* Icons will use a horizontal layout

Note: The dropdown menu will not be implemented in this sprint.

image

lisa-fehr:
Closes #789 

I split the highlight logic into its own mixin and away from the new crown config.

Much of the code in the crown config component was copied from the crown mixin.


src/components/crownConfig.vue
```diff
@@ -0,0 +1,275 @@
...
+      @click="addSequence()"
+      class="crown-config__icon"
+      :title="$t('Sequence Flow')"
+    />
+    <message-flow-button
+      @click="addMessageFlow()"
+      class="crown-config__icon"
+      :title="$t('Message Flow')"
+    />
+    <delete-button
```
lisa-fehr: Might need to tweak this. I thought it was using the svg, but it seems to not work without the classes: 'svg-inline--fa fa-trash-alt fa-w-14'

tests/e2e/specs/BoundaryTimerEvent.spec.js
```diff
@@ -156,7 +156,7 @@ describe('Boundary Timer Event', () => {
     dragFromSourceToDest(nodeTypes.task, task2Position);
 
     const boundaryTimerEventSelector = '.main-paper ' +
-      '[data-type="processmaker.components.nodes.task.Shape"] ~ ' +
+      '[data-type="newCrown"] ~ ' +
```
lisa-fehr: Temporary solution to get tests passing.

lisa-fehr:
Still need to convert Call Activity task. It did not extend the basic task.

src/components/crownConfig.vue
```diff
@@ -0,0 +1,273 @@
...
+  ],
+  watch: {
+    highlighted() {
+      this.showCrown = this.highlighted;
+    },
+    shape() {
+      if (this.highlighted) {
+        this.showCrown = true;
+        this.configurePoolLane();
+        this.setUpCrownConfig();
```
chipit24: In the original crown config, these two functions are only called once on mount. And even though this is in a watcher, `shape` should never change, and that seems to be the case here, as this watcher seems to only be called once after dragging the shape onto the grid.

Maybe break this out into a separate function that is only called once in the mounted hook?

src/components/nodes/task/task.vue
```diff
@@ -1,24 +1,44 @@
...
+      :node="node"
+      :nodeRegistry="nodeRegistry"
+      :moddle="moddle"
+      :collaboration="collaboration"
+      :process-node="processNode"
+      :plane-elements="planeElements"
+      @remove-node="$emit('remove-node', $event)"
+      @add-node="$emit('add-node', $event)"
+      @save-state="$emit('save-state', $event)"
+    />
```
chipit24: I really like how this is a separate component now instead of a mixin 🔥

src/components/crownConfig.vue
```diff
@@ -0,0 +1,273 @@
...
+      defaultHighlighter: {
+        name: 'stroke',
+        options: {
+          attrs: {
+            stroke: '#feb663',
+            'stroke-width': 3,
+          },
+        },
+      },
+      /* allowSetNodePosition is used to prevent setting a node position outside of a pool */
```
chipit24: Since this seems to be moved to `highlightConfig`, is it still needed here?

src/components/crownConfig.vue
lisa-fehr: I had issues where it was erroring due to shape not being there. Is there a nice way to call it once, knowing that shape is initially null?

src/components/crownConfig.vue
lisa-fehr: There was logic for both highlighting and saving the position in the original setUpCrownConfig. I ended up having one in highlightConfig and crownConfig. I believe this is needed to position it correctly after undo/redo. Which didn't seem to be highlight related.

src/components/crownConfig.vue
lisa-fehr: I guess I can try next tick

src/components/crownConfig.vue
chipit24: I think I highlighted an extra line here; I was referring to removing `defaultHighlighter` since it's not used in this file, but appears to be moved over to the new highlightConfig mixin.

lisa-fehr: Whew. Thanks. That can go.

PR #801, opened 2019-11-12
Fix duplicate diagram IDs

### LINKED ISSUE #774: Duplicated ID issue
# Bug report issue
## Description
When updating the provided bpmn in a certain way, it duplicates node_17_di.
https://github.com/ProcessMaker/modeler/files/3617104/bugProcess.bpmn.txt
## Reproduction steps
You must upload the provided bpmn and then edit it from this:
Screen Shot 2019-10-30 at 10 20 14 AM
to this:
Screen Shot 2019-10-30 at 10 33 28 AM

The Sequence Flow is changed from **Report Expense** to **Before Dynaform**

Then it does duplicate **node_17_di** in the xml.
## Expected behaviour
Change the Start Sequence Flow to **Before Dynaform** and it should not duplicate **node_17_di**
## Actual behaviour
**node_17_di** is duplicated when the sequence flow is changed from **Report Expense** to **Before Dynaform**
## Software used
Latest Chrome. Modeler v 0.17.5.

chipit24:
Fixes #774.

Code is already in place for ensuring nodes do not receive duplicate IDs; this PR prevents added nodes from receiving a duplicate _**diagram**_ ID.


src/NodeIdGenerator.js
```diff
@@ -1,33 +1,55 @@
...
+      diagramId = this.#generateDiagramId();
+    }
+
+    return [definitionId, diagramId];
   }
 
-  generateNewNodeId(prefix = nodeIdPrefix) {
-    const id = prefix + this.counter;
-    this.counter++;
+  #generateDefinitionId = () => {
```
chipit24: I had to use private properties here instead of private methods as there appears to be bugs with `babel-eslint` when it comes to using private methods that I could not fix by updating the relevant dependencies. See https://github.com/babel/babel-eslint/issues/728.

Issue #803, opened 2019-11-13
Add new crown to Boundary Events


lisa-fehr:
**Acceptance Criteria:**
Should only have a trash icon and sequence flow.

Covers: 
- Boundary Timer
- Boundary Error 
- Boundary Message
- Boundary Escalation

![Screen Shot 2019-11-13 at 9 53 23 AM](https://user-images.githubusercontent.com/6653340/68774699-7f5db800-05fb-11ea-9d5d-0e44f847015c.png)

Related: #789

Issue #802, opened 2019-11-13
Add new crown to Call Activity


lisa-fehr:
**Acceptance Criteria:**
Should only have a trash icon and a sequence flow.

![Screen Shot 2019-11-13 at 9 44 06 AM](https://user-images.githubusercontent.com/6653340/68773814-3822f780-05fa-11ea-8b0f-f964f682a190.png)

Related: #789

PR #811, opened 2019-11-13
Add new crown to Boundary Events

lisa-fehr:
I ended up removing the altered data-type = newCrown. It was used as a selector in tests and was now the same as tasks.

Can merge into #802 or switch the base to develop.

Closes: #803

### DIFF
--- a/src/components/crownConfig.vue
+++ b/src/components/crownConfig.vue
@@ -88,9 +88,10 @@ export default {
   },
   computed: {
     style() {
-      const { x, y } = this.shape.findView(this.paper).getBBox();
+      const { x, y, width: shapeWidth } = this.shape.findView(this.paper).getBBox();
       const width = (this.node.diagram.bounds.width * this.paper.scale().sx);
-      return 'top:' + (y - 45) + 'px;left:' + (x + width - 20) + 'px;cursor:pointer';
+      const widthOffsetFromLabel = shapeWidth - width > 3 ? (shapeWidth / 2) - 15 : 0;
+      return 'top:' + (y - 45) + 'px;left:' + ((x + width - 20) + widthOffsetFromLabel) + 'px;cursor:pointer';
     },
     isValidMessageFlowSource() {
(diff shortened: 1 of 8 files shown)


src/components/crownConfig.vue
```diff
@@ -88,9 +88,10 @@ export default {
   },
   computed: {
     style() {
-      const { x, y } = this.shape.findView(this.paper).getBBox();
+      const { x, y, width: shapeWidth } = this.shape.findView(this.paper).getBBox();
       const width = (this.node.diagram.bounds.width * this.paper.scale().sx);
-      return 'top:' + (y - 45) + 'px;left:' + (x + width - 20) + 'px;cursor:pointer';
+      const widthOffsetFromLabel = shapeWidth - width > 3 ? (shapeWidth / 2) - 15 : 0;
```
lisa-fehr: Ugh. The label moves the x to the left when dragging a boundary event. This calculation got it as close to the original placement as I could get. Any way to get just the shape?

chipit24: @lisa-fehr You can get the width from `this.shape`:
```js
const shapeWidth = this.shape.size().width;
```

That should return the original size even if the label extends outside the shape bounds.

tests/e2e/support/utils.js
```diff
@@ -92,7 +97,7 @@ export function getPositionInPaperCoords(position) {
 }
 
 export function getCrownButtonForElement($element, crownButton) {
-  if ($element.data('type') === 'newCrown') {
+  if (newCrown.includes($element.data('type'))) {
```
chipit24: Just a suggestion, but I think this might read better using a function, like:
```js
if (hasNewCrownConfig($element)) { ...
```

src/components/crownConfig.vue
lisa-fehr: The width is fine. It is the x/y that is including the label and moving it to the left of the shape.

PR #825, opened 2019-11-14
Added new crown to Pool Lanes

### LINKED ISSUE #820: Update lane to use new crown
**Acceptance Criteria:**
Should have trash icon only.

Covers: 
- Lanes

Related: #789

lisa-fehr:
Closes #820

### DIFF
--- a/src/components/crownConfig.vue
+++ b/src/components/crownConfig.vue
@@ -72,6 +72,7 @@ export default {
         'processmaker-modeler-end-event',
         'processmaker-modeler-error-end-event',
         'processmaker-modeler-message-end-event',
+        'processmaker-modeler-lane',
       ],
     };
   },
@@ -83,7 +84,7 @@ export default {
       /* Use nextTick to ensure this code runs after the component it is mixed into mounts.
        * This will ensure this.shape is defined. */
 
-      this.configurePoolLane();
+      //this.configurePoolLane();
 
       if (this.isRendering) {
         this.paper.once('render:done', this.setUpCrownConfig);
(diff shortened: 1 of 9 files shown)


src/components/crownConfig.vue
```diff
@@ -143,6 +148,9 @@ export default {
         diagram: this.moddle.create('bpmndi:BPMNEdge'),
       });
     },
+    remove() {
+      return this.isLane ? this.removePoolLaneShape() : this.removeShape();
```
lisa-fehr: This logic is in a mixin. Maybe there is a better way.

src/components/nodes/boundaryEvent/boundaryEvent.vue
```diff
@@ -11,6 +11,7 @@
     :process-node="processNode"
     :plane-elements="planeElements"
     :is-rendering="isRendering"
+    :allow-set-node-position="allowSetNodePosition"
```
lisa-fehr: Need to add this to pools too, but that is another pr.

lisa-fehr:
Going to add a few tests that seem to be missing for pool lanes.

lisa-fehr:
@chipit24 I tried to add a test for resizing pool lanes. I got a data-test attribute on the 4 corners but there ends up being 6 instances of each corner. So I had to put that on hold.

src/components/crownConfig.vue
JCook21: I think for this the inheritance seems to be reversed. Currently `crownConfig` extends `poolLaneCrownConfig`, which means crown config has to know about pool lanes. Would it be possible to have `poolLaneCrownConfig` extend `crownConfig` instead? It would then be possible to have a single `remove()` method on `crownConfig` which `poolLaneCrownConfig` can then override to add the specific behaviour. This would also mean that a generic component doesn't need to know anything about the specific behaviour of the pool lane.

src/mixins/poolLaneCrownConfig.js
```diff
@@ -0,0 +1,89 @@
+export default {
+  props: [
+    'node',
+    'graph',
+  ],
+  computed: {
+    isLane() {
+      return this.node.type === 'processmaker-modeler-lane';
+    },
```
JCook21: If we could reverse the inheritance this would not be necessary I think.

src/components/crownConfig.vue
```diff
@@ -22,7 +22,7 @@
       :title="$t('Message Flow')"
     />
     <delete-button
```
lisa-fehr: Here's a thought. What if this had a default delete button and action and it could be replaced with a delete-button slot?

lisa-fehr: Ok, the remove-node emit on pool lane could be weird.

src/components/crownConfig.vue
lisa-fehr: It doesn't seem to work in that order. Or there are things in the crownConfig I need to make sure run first.

JCook21: Can you be more specific? Have you already tried this? What things need to run first? What errors did you see?

lisa-fehr: Think I got it. Thanks for the suggestion.

lisa-fehr: In the crownConfig it updates things like plane elements, artifacts, and flow elements.

I may have an issue to solve. Elements aren't embedding in the pool.

src/mixins/poolLaneCrownConfig.js
lisa-fehr: Unfortunately. I think it contains logic the pool needs to add lanes into it.

JCook21: Shouldn't this be on the pool or the lane then? I don't see how this logic is related to the crown config but I may be missing something.

lisa-fehr:
@JCook21 I'm not certain I can move the pool logic out when the pool still has the old crown in this branch.

PR #829, opened 2019-11-18
Update crown for pool elements

### LINKED ISSUE #819: Update pool to use new crown
**Acceptance Criteria:**
Should have add-lane-above, add-lane-bottom, sequence flow, and trash icon

Screen Shot 2019-11-14 at 12 39 43 PM

Covers: 
- Pool

Related: #789

chipit24:
Fixes #819.

### DIFF
--- a/src/components/crownConfig.vue
+++ b/src/components/crownConfig.vue
@@ -189,7 +189,7 @@ export default {
         return;
       }
 
-      const { x, y, width } = shapeView.getBBox();
+      const { x, y, width } = shapeView.getBBox({ useModelGeometry: !this.isTextAnnotation });
 
       this.style = {
         top: `${y - 45}px`,
--- a/src/components/crownConfig.vue
+++ b/src/components/crownConfig.vue
@@ -1,5 +1,6 @@
 <template>
   <div class="crown-config" :style="style" v-if="showCrown">
+    <slot/>
     <sequence-flow-button
       v-if="isValidSequenceFlowSource"
       @click="addSequence"
@@ -72,6 +73,7 @@ export default {
         'processmaker-modeler-end-event',
         'processmaker-modeler-error-end-event',
         'processmaker-modeler-message-end-event',
+        'processmaker-modeler-pool',
       ],
     };
   },
@@ -94,10 +96,10 @@ export default {
   },
   computed: {
     style() {
-      const { x, y, width: shapeWidth } = this.shape.findView(this.paper).getBBox();
-      const width = (this.node.diagram.bounds.width * this.paper.scale().sx);
-      const widthOffsetFromLabel = shapeWidth - width > 3 ? (shapeWidth / 2) - 15 : 0;
-      return 'top:' + (y - 45) + 'px;left:' + ((x + width - 20) + widthOffsetFromLabel) + 'px;cursor:pointer';
+      const { x, y } = this.shape.findView(this.paper).getBBox();
+      const width = this.shape.size().width * this.paper.scale().sx;
+
+      return 'top:' + (y - 45) + 'px;left:' + (x + width - 20) + 'px;cursor:pointer';
     },
     isValidMessageFlowSource() {
       return this.validMessageFlowSources.includes(this.node.type);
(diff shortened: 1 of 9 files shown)


src/mixins/resizeConfig.js
```diff
@@ -486,7 +482,7 @@ export default {
     },
     setPointAttributes(point, cursorDirection) {
       point.attr({
-        root: { display: 'none' },
+        root: { display: 'initial' },
```
chipit24: This was causing the "resize squares" to not display initally when the pool was dragged onto the grid. Changing it to `'initial'` fixes this.

src/components/crownConfig.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <div class="crown-config" :style="style" v-if="showCrown">
+    <slot/>
```
chipit24: Decided to introduce a slot to add component specific buttons (in this case, the add lane buttons).

src/components/crownConfig.vue
```diff
@@ -204,7 +182,25 @@ export default {
...
+      const { x, y, width } = shapeView.getBBox();
+
+      this.style = {
+        top: `${y - 45}px`,
+        left: `${x + width - 20}px`,
+        cursor: 'pointer',
+      };
+    },
     setUpCrownConfig() {
+      this.paper.on('render:done scale:changed translate:changed', this.repositionCrown);
```
chipit24: I triggered custom events on the paper to handle certain changes that required repositioning the crown. I found this allowed me to get an actual width from `getBBox`.

src/components/nodes/pool/pool.vue
```diff
@@ -1,23 +1,51 @@
...
+    :nodeRegistry="nodeRegistry"
+    :moddle="moddle"
+    :collaboration="collaboration"
+    :process-node="processNode"
+    :plane-elements="planeElements"
+    :is-rendering="isRendering"
+    @remove-node="$emit('remove-node', $event)"
+    @add-node="$emit('add-node', $event)"
+    @save-state="$emit('save-state', $event)"
+  >
```
lisa-fehr: Pool has allowSetNodePosition which is in the crown component.
Does `element.component.allowSetNodePosition` in this file still work with the new crown component?

src/components/crownConfig.vue
```diff
@@ -204,7 +182,25 @@ export default {
 
       this.$emit('save-state');
     },
+    repositionCrown() {
+      const shapeView = this.shape.findView(this.paper);
+
+      if (!shapeView) {
```
lisa-fehr: Boundary events and intermediate events are now placing the crown at the end of the label on initial drop.

lisa-fehr: If you drag an escalation event from a call activity to a task, the crown moves to the label after error.

src/components/nodes/pool/pool.vue
chipit24: Yup, that still seems to work as expected. Did you notice any problems with it?

src/components/crownConfig.vue
chipit24: Looks like there's a `useModelGeometry` option that can be passed to `getBBox` that fixed this:
https://resources.jointjs.com/docs/jointjs/v3.1/joint.html#dia.ElementView.prototype.getBBox

That's super useful!

chipit24:
Adding WIP label as I'm going to pull out logic related to `allowSetNodePosition` to fix the interaction with the pool when an element is dragged outside of it.

The current issue is that the pool component reaches into the component state of an element that is being dragged off of it, and sets that component's `allowSetNodePosition` value to `false`.

The `allowSetNodePosition` value was initially mixed into all components, but now resides in the new crown-config component that is a child component. So fixing this directly would require passing down the `allowSetNodePosition` prop.

There are a few issues with this: 
* This should not be logic that is in the `crownConfig` file.
* Reaching into a component's internal state should be avoided.
* "Prop drilling" (components basically acting as middlemen for props and passing them down without using the prop themselves) should be avoided.

PR #831, opened 2019-11-19
Merge warning and error panels

### LINKED ISSUE #792: Merge Warning and Error Panel
Merge the content of the two panels into a single one. Use the Chrome console as reference.

tuantruong625:
**Acceptance Criteria**
- Warnings and Error panels merge into one.
- Button shows a green checkmark and `BPMN Valid` if there are no issues
image

- Button shows a number of issues if diagram contains issues
image

image

**Todo**

- [x] Modify `warnings` object

- [x] Merge warnings + validation error props

- [x] Add element type to `Element:` heading

- [x] Fix Two Failing test 
      ` Validates gateway direction - Modeler.spec.js`
      ` shows warning for unknown element during parsing - Modeler.spec.js`

Fixes #792

### DIFF
--- a/src/components/ValidationStatus.vue
+++ b/src/components/ValidationStatus.vue
@@ -11,9 +11,13 @@
 
     <span class="divider" />
 
-    <div v-if="toggleValidationPanel && numberOfProblems" class="validation-container position-absolute text-left" data-test="validation-list">
+    <div v-if="isProblemsPanelDisplayed && numberOfProblems" class="validation-container position-absolute text-left" data-test="validation-list">
       <div class="validation-container-list d-flex align-items-baseline" v-for="error in errorList" :key="`${error.id}_${error.errorKey}`">
-        <font-awesome-icon class="status-bar-container-status-icon ml-1 mr-2 mt-1" :style="{ color: iconColor }" :icon="validationIcon" />
+        <font-awesome-icon
+          class="status-bar-container-status-icon ml-1 mr-2 mt-1"
+          :style="{ color: isErrorCategory(error) ? errorColor : warningColor }"
+          :icon="isErrorCategory(error) ? faTimesCircle: faExclamationTriangle"
+        />
         <div class="validation-container-list-message">
           <h6 class="text-capitalize mb-0">{{ error.errorKey }}</h6>
           <p class="mb-0"><em>{{ error.message }}.</em></p>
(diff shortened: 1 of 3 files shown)


src/components/ValidationStatus.vue
```diff
@@ -118,93 +82,41 @@ export default {
...
-$button-color: #3BD7FF;
 
 .status-bar-container {
-  color: $seconadry-grey;
+  color: $secondary-grey;
   height: $status-bar-container-height;
-  cursor: pointer;
 
-  &__status {
+  &-status {
```
JCook21: I changed these styles to be dashes instead if underscores as I find the class names a little easier to read. Just a matter of personal preference and I'm happy to change them back if people feel strongly.

src/components/ValidationStatus.vue
```diff
@@ -215,6 +127,7 @@ $button-color: #3BD7FF;
 }
 
 .validation-container {
+  cursor: default;
```
JCook21: Previously the cursor implied that something in the validation panel could be clicked. This changes that and makes sure the default cursor is displayed.

src/components/ValidationStatus.vue
tuantruong625: I'm fine with either or. I usually follow the BEM methodology when naming my classes just personal preference. :)

http://getbem.com/

lisa-fehr: I'm not a css pro, but the existing style is one I see more often in other projects.

lisa-fehr:
@JCook21 Noticed it says it is `BPMN Valid` when I haven't validated the code. That is misleading.

It isn't addressed in the issue. But I think it needs another state. Even if the checkmark is just a question mark.

src/components/ValidationStatus.vue
```diff
@@ -1,113 +1,77 @@
...
-          <font-awesome-icon
-            class="status-bar-container__status-icon h-100"
-            :style="{ color: warnings.length > 0 ? warningColor : validColor }"
-            :icon="warnings.length > 0 ? faExclamationTriangle : faCheckCircle"
-          />
-          <font-awesome-icon class="status-bar-container__status-ellipsis" :icon="ellipsisIcon" />
+      <div class="validation-container-list d-flex align-items-baseline" v-for="(warning, index) in warnings" :key="index">
+        <font-awesome-icon class="status-bar-container-status-icon ml-1 mr-2 mt-1" :style="{ color: warningColor }" :icon="faExclamationTriangle" />
+        <div class="validation-container-list-message">
+          <h6 class="text-capitalize mb-0">{{ warning.title }}</h6>
```
chipit24: Is the `h6` here only used for the styling? I actually don't see any other heading levels used on the page, so skipping to level 6 seems like it's done for styling as oppose to creating some sort of semantic hierarchy/outline.

See https://developers.google.com/web/fundamentals/accessibility/how-to-review?utm_source=lighthouse&utm_medium=devtools#take_advantage_of_headings_and_landmarks:
> Make proper use of h1-h6 hierarchy. Think of headings as tools to create an outline for your page. Don’t rely on the built-in styling of headings; instead, consider all headings as if they were the same size and use the semantically appropriate level for primary, secondary, and tertiary content. Then use CSS to make the headings match your design.

You may not even need heading elements here, as there's no real sections that need a title. I think using a description/definition list here might work well: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl.

src/components/ValidationStatus.vue
```diff
@@ -1,113 +1,77 @@
...
+          <p class="mb-0"><em>{{ warning.text }}.</em></p>
```
chipit24: This structure may change if you switch to using a `dl` element, but I don't think the wrapper `p` element is required. Could you get away with using only the `em` element here?

Another note, following from the previous comment, is that `em` is used for emphasis, and if you're only concerned about applying an italicized style, that should be done with CSS instead.

src/components/ValidationStatus.vue
JCook21: OK, I'll switch it back if it's the convention.

src/components/ValidationStatus.vue
JCook21: I did think about a definition list, let me try that out.

PR #836, opened 2019-11-19
Fix pool undo bug

### LINKED ISSUE #833: Element appears outside of pool when pressing undo
## Description
Elements with the new crown appear outside of a pool when pressing "undo".

## Reproduction steps
1. Drag a pool onto the grid.
2. Drag the start event outside of the pool (the grid should turn red and the start event should return to the pool once released).
3. Press undo.

## Expected behaviour
The start event re-appears in its original position inside the pool.

## Actual behaviour
The start event remains outside of the pool.

## Software used
Latest modeler `develop` branch, Chrome 78 on macOS Mojave.

## Other information
pool-bug

chipit24:
Fixes #833.


src/components/crownConfig.vue
```diff
@@ -243,15 +237,24 @@ export default {
...
-      } else {
-        this.setUpCrownConfig();
-      }
-    });
+  async mounted() {
+    await this.$nextTick();
+    await this.paperDoneRendering();
+
+    this.setUpCrownConfig();
+    this.setUpPositionHandling();
```
JCook21: Nice change, this is much clearer to read.

src/components/nodes/boundaryEvent/boundaryEvent.vue
```diff
@@ -147,7 +146,7 @@ export default {
     },
     resetToInitialPosition() {
       this.shape.position(this.validPosition.x, this.validPosition.y);
-      this.allowSetNodePosition = true;
+      store.commit('allowSavingState');
```
JCook21: Could we add a method to the store that encapsulates this? As it stands this doesn't really clarify what's going on well.

src/components/nodes/boundaryEvent/boundaryEvent.vue
```diff
@@ -171,13 +170,14 @@ export default {
         return;
       }
 
-      this.allowSetNodePosition = false;
+      store.commit('preventSavingState');
```
JCook21: Same comment as above.

src/components/crownConfig.vue
```diff
@@ -207,20 +204,17 @@ export default {
         cursor: 'pointer',
       };
     },
+    paperDoneRendering() {
+      if (!this.isRendering) {
+        return;
+      }
+
+      new Promise(resolve => this.paper.once('render:done', resolve));
+    },
```
chipit24: I feel like this should belong to `paperManager`, but the reliance on `isRendering` makes this tricky, so I'll defer that change for now.

Knowing if rendering was triggered, or if it's actually in progress would help clean up our rendering related logic. I posted this question in the JointJS repo here: https://github.com/clientIO/joint/issues/1219.

src/components/nodes/pool/pool.vue
```diff
@@ -558,18 +559,19 @@ export default {
             invalidPool = null;
           }
 
-          this.paper.drawBackground({ color: invalidNodeColor });
-          element.component.allowSetNodePosition = false;
+          store.commit('preventSavingState');
+          this.paperManager.setInvalidGirdColor();
```
JCook21: Gird (https://www.merriam-webster.com/dictionary/gird) or Grid?

JCook21: Also, I wonder if this could be better named. Callers of this method shouldn't care about what's happening, all they should care about is that the paper manager is managing the invalid state for them. how about `setStateInvalid()` and `setStateValid()` instead?

lisa-fehr:
Works great on pools and pool lanes. 👍

PR #838, opened 2019-11-20
Cleanup interim code

### LINKED ISSUE #809: Clean up override when new crown is on all elements
When all elements have the new crown, get rid of the override on element type.

- In utils.js, remove `hasNewCrownConfig` check and 
```
return cy
    .get(`#${$element.attr('id')} ~ [data-test=${crownButton}]`)
    .then(crownButtons => crownButtons.filter((index, button) => Cypress.$(button).is(':visible')))
    .then(crownButtons => crownButtons[0]);
```
- Delete old crownConfig.js

- Update svg components (deleteButton, messageFlowButton, sequenceFlowButton) to import the svg instead of embedding. Update svg to white color.
Like: https://github.com/ProcessMaker/modeler/pull/816/files#diff-0a6bef07cb969f5b08e2333d3c82c4c0R11

JCook21:
Resolves #809. 

- Removes interim code added to support testing when some elements used the old crown and some the new.
- Convert the new crown components to use the SVG images in assets instead of inline SVG data.
- Edit the SVG image colours to be white.


src/components/nodes/textAnnotation/textAnnotation.vue
```diff
@@ -68,7 +67,7 @@ export default {
         newHeight = currentBoundsHeight;
       }
 
-      this.shape.resize(this.nodeWidth, newHeight - highlightPadding);
+      this.shape.resize(this.nodeWidth, newHeight - 3);
```
JCook21: This change was needed to allow the crown config mixing to be removed. I'll try to find a better place to declare this as a constant.

src/components/paperManager.js
```diff
@@ -22,7 +21,7 @@ export default class PaperManager {
       perpendicularLinks: true,
       interactive: interactiveFunc,
       highlighting: {
-        default: { options: { padding: highlightPadding } },
+        default: { options: { padding: 3 } },
```
JCook21: Same comment as above.

src/components/nodes/textAnnotation/textAnnotation.vue
```diff
@@ -88,7 +93,7 @@ export default {
       });
 
       this.paper.once('render:done', () => {
-        this.setElementHeight(height, bounds.height, text);
+        this.shape.resize(this.nodeWidth, this.calculateNewHeight(height, text, bounds.height));
```
JCook21: I decided to refactor this as I worked on extracting the padding into a variable. The problem with the previous method was that it combined a command and a query IMO.

src/components/messageFlowButton.vue
```diff
@@ -1,30 +1,23 @@
...
+    :src="messageFlow"
     @click="$emit('click')"
     id="message-flow-button"
     data-test="message-flow-button"
     width="19"
     height="20"
-    viewBox="0 0 19 20"
-    fill="none"
-    xmlns="http://www.w3.org/2000/svg"
+    alt="Create a message flow"
```
lisa-fehr: I tried to check but my emulators wouldn't load up. Do Microsoft browser still display the alt text on hover?

JCook21: Good question, I'll try to find out.

JCook21: I can't find many references to this online. One Reddit post from 2018 did link to a (dead) link on Microsoft saying that this behaviour was being removed from their browsers.

src/components/paperManager.js
```diff
@@ -1,5 +1,4 @@
 import { dia } from 'jointjs';
-import { highlightPadding } from '@/mixins/crownConfig';
```
lisa-fehr: I did move crownConfig stuff to mixins/highlightConfig. If you want to do something like this instead:
```
import { highlightPadding } from '@/mixins/highlightConfig';
```

And place this in highlightConfig:
`export const highlightPadding = 3;`

It is only used in a few places, so either way is probably fine.

JCook21: I think I'll probably leave this 'as is' for now. I'm not convinced that these two padding amounts really represent the same thing and warrant being extracted to a single constant in this case.

src/components/deleteButton.vue
```diff
@@ -8,10 +9,18 @@
     data-icon="trash-alt"
     class="svg-inline--fa fa-trash-alt fa-w-14"
     role="img"
```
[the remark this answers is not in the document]
JCook21: Sounds good. I created a new issue for converting these to buttons so that we don't lose sight of making this change.

PR #843, opened 2019-11-20
Fix lane overlap bug

### LINKED ISSUE #824: Pool Lanes can hide content
Create/edit a process, add multiple controls

Add a pool control

Add at least one extra pool lane

Change the height of one of the pool lanes

Result:

Controls get hidden by the movement of additional pool lanes, can no longer be selected while hidden.

https://www.dropbox.com/s/i88kj81t1npf26q/pools-lanes-overlapping.mp4?dl=0

chipit24:
Fixes #824.

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -137,7 +137,15 @@ import runningInCypressTest from '@/runningInCypressTest';
 import getValidationProperties from '@/targetValidationUtils';
 import MiniPaper from '@/components/MiniPaper';
 
-import { faCompress, faExpand, faMapMarked, faMinus, faPlus, faStepBackward, faStepForward } from '@fortawesome/free-solid-svg-icons';
+import {
+  faCompress,
+  faExpand,
+  faMapMarked,
+  faMinus,
+  faPlus,
+  faStepBackward,
+  faStepForward,
+} from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 
 import { id as poolId } from './nodes/pool';
@@ -771,6 +779,21 @@ export default {
     },
     bringPoolToFront(poolShape) {
       this.bringShapeToFront(poolShape);
+
+      const hasLanes = poolShape.component.node.definition.processRef.laneSets[0];
+      if (!hasLanes) {
+        return;
+      }
+
+      const poolLanes = poolShape
+        .getEmbeddedCells()
+        .filter(cell => cell.component && cell.component.node.type === laneId);
+      const poolLaneElements = poolShape
+        .getEmbeddedCells()
+        .filter(cell => cell.component && cell.component.node.type !== laneId);
+
+      poolLanes.forEach(this.bringShapeToFront);
+      poolLaneElements.forEach(this.bringShapeToFront);
     },
     bringShapeToFront(shape) {
       shape.toFront({ deep: true });
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -147,8 +147,6 @@ import {
   faStepForward,
 } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
-
-import { id as poolId } from './nodes/pool';
 import { id as laneId } from './nodes/poolLane';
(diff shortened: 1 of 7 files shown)


chipit24:
The single failing test in this branch is the one I added, "_should not cover child elements with lane_", which is expected.

src/components/Modeler.vue
```diff
@@ -160,6 +158,7 @@ import registerInspectorExtension from '@/components/InspectorExtensionManager';
 
 import initAnchor from '@/mixins/linkManager.js';
 import { addIdToNodeAndSetUpDiagramReference, addNodeToProcess, getTargetProcess } from '@/components/nodeManager';
+import ensureShapeIsNotCovered from '@/components/shapeStackUtils';
```
chipit24: This name could be improved, but I figured it's better than `setShapeStacking`, and avoids describing what's happening in an imperative way: it should be clear that we're not concerned about setting specific z-indices, but about making sure a shape isn't covered or overlapping another shape.

src/components/nodes/poolLane/index.js
```diff
@@ -3,8 +3,6 @@ import idConfigSettings from '@/components/inspectors/idConfigSettings';
 import nameConfigSettings from '@/components/inspectors/nameConfigSettings';
 
 export const id = 'processmaker-modeler-lane';
-export const minLaneHeight = 100;
-
```
chipit24: This export wasn't actually used anywhere, and the value for the minimum lane height is specified in `src/components/nodes/poolLane/laneSizes.js`.

tests/e2e/specs/Pools.spec.js
```diff
@@ -228,4 +230,29 @@ describe('Pools', () => {
...
+      .then($pool => {
+        getCrownButtonForElement($pool, 'lane-above-button').click({ force: true });
+      });
+
+    const startEventPosition = { x: 150, y: 150 };
+    const topLanePosition = { x: startEventPosition.x, y: startEventPosition.y - 50 };
+    getPositionInPaperCoords(startEventPosition).then(({ x, y }) => {
+      getElementAtPosition(topLanePosition, nodeTypes.poolLane).click();
+
+      cy.get('.main-paper [button-id="bottom-right-resize-button"]')
```
lisa-fehr: Nice. I do see that there can be two bottom-right-resize-buttons. But it seems to work regardless.
Screen Shot 2019-11-21 at 2 33 31 PM

src/components/shapeStackUtils.js
```diff
@@ -0,0 +1,54 @@
...
+}
+
+function getElementPool(shape) {
+  return shape.component.node.pool;
+}
+
+function isPool(shape) {
+  return shape.component.node.type === poolId;
+}
+
```
chipit24: I thought about adding unit tests for this, but I wasn't sure it would add any value. The heavy-lifting is accomplished by calls to `toFront`, and tests would essentially be verifying that `toFront` was called in the correct order, the correct number of times.

There might be better combinations of calling `toFront` that also fix overlapping issues, or maybe there's a different JointJS function we could use. I think unit tests could slow us down in those cases, and I think this functionality is sufficiently covered by our e2e tests. 

Maybe that's a hint that this isn't the best (or correct abstraction) right now, but I think it's still much better than having all this logic spread into `Modeler.vue`.

src/mixins/resizeConfig.js
```diff
@@ -464,12 +459,20 @@ export default {
...
-        button.attr('root/display', 'initial');
+        try {
+          this.shape.embed(button);
+        } catch (error) {
+          /* There is an error when re-adding points after removing/re-adding the pool.
+           * Ignore it for now. */
+        }
+
+        button.addTo(this.graph);
+        button.toFront();
```
chipit24: The left corner handles were hidden behind the pool, so I had to call this to ensure it was always visible.

tests/e2e/specs/Pools.spec.js
chipit24: How did you search for the 2nd bottom right resize button? It seems to not be visible in the screenshot above. I think it may be part of the mini-map paper, which I think duplicates the handles as well.

lisa-fehr: I highlighted the pool and then did a text search in the Elements console. The other bottom right resize was where I expected it to be.

You are correct. It is the mini-map!

PR #848, opened 2019-11-22
Moved crown images into buttons

### LINKED ISSUE #839: Convert crown images to buttons
img elements have an implicit role of "img" and so this shouldn't be set. See https://www.w3.org/TR/html-aria/#document-conformance-requirements-for-use-of-aria-attributes-in-html.

The best solution for this should be to use a button element with the `img` as its child. The button can then have an `aria-label`, and the img's alt text can be set to an empty string `""` and have a `role` of "presentation". The button would of course need to be slighly re-styled. This would make more the most accessible and semantically correct structure, in my opinion.

_Originally posted by @chipit24 in https://github.com/ProcessMaker/modeler/pull/838_

lisa-fehr:
Closes #839 

Adds more area to click around the crown icons.

### DIFF
--- a/src/components/crownButton.vue
+++ b/src/components/crownButton.vue
@@ -9,6 +9,8 @@
       <img
         :src="src"
         alt=""
+        width="width"
+        height="height"
       >
     </slot>
   </button>
@@ -16,7 +18,19 @@
 <script>
 
 export default {
-  props: ['id', 'ariaLabel', 'src'],
+  props: {
+    'id': String,
+    'ariaLabel': String,
+    'src': String,
+    'width': {
+      type: Number,
+      default: 19,
+    },
+    'height': {
+      type: Number,
+      default: 20,
+    },
+  },
 };
 </script>
 
@@ -29,7 +43,5 @@ export default {
 img {
   margin: 5px 10px;
   font-size: 1rem;
-  width: 19px;
-  height: 20px;
 }
 </style>
--- a/src/components/crownButton.vue
+++ b/src/components/crownButton.vue
@@ -1,21 +1,27 @@
 <template>
   <button
-    @click="$emit('click')"
+    v-on="$listeners"
     :id="id"
     :aria-label="ariaLabel"
+    class="btn"
   >
-    <slot/>
+    <slot>
+      <img
+        :src="src"
+        alt=""
+      >
+    </slot>
   </button>
 </template>
 <script>
 
 export default {
-  props: ['id', 'ariaLabel'],
+  props: ['id', 'ariaLabel', 'src'],
 };
 </script>
 
 <style scoped>
-button {
+.btn {
   border: none;
   padding: 0;
   margin-top: 0;
--- a/src/components/crownButton.vue
+++ b/src/components/crownButton.vue
@@ -1,0 +1,29 @@
+<template>
+  <button
+    @click="$emit('click')"
+    :id="id"
+    :aria-label="ariaLabel"
+  >
+    <slot/>
+  </button>
+</template>
+<script>
+
+export default {
+  props: ['id', 'ariaLabel'],
+};
+</script>
+
+<style scoped>
+button {
+  border: none;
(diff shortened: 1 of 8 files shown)


src/components/crownButton.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+  <button
+    @click="$emit('click')"
```
chipit24: According to the docs at https://vuejs.org/v2/api/#inheritAttrs, it seems like you can remove the entire script block and all props, and all appropriate attributes will be applied to this root `button` element as expected (you don't need to set `inheritAttrs` to `false`).

I'm not sure if that will happen with event listeners, so in that case, you can put a `v-on="$listeners"` on the button.

src/components/deleteButton.vue
```diff
@@ -1,22 +1,26 @@
...
     id="delete-button"
-    data-test="delete-button"
-    aria-hidden="true"
-    data-prefix="fas"
-    data-icon="trash-alt"
-    class="svg-inline--fa fa-trash-alt fa-w-14"
-    role="img"
-    alt="Delete this node"
+    aria-label="Delete this node"
+    @click="$emit('click')"
```
chipit24: Since this click listener is also just being passed down, I think you can replace it with `v-on="$listeners"`; `crown-button` is really a transparent wrapper.

src/components/messageFlowButton.vue
```diff
@@ -1,19 +1,21 @@
...
-    height="20"
-    alt="Create a message flow"
+    aria-label="Create a message flow"
+    @click="$emit('click')"
   >
-
+    <img
+      :src="messageFlow"
+      alt=""
+    >
```
chipit24: For the delete button you  had additional properties:
```
aria-hidden="true"
data-prefix="fas"
data-icon="trash-alt"
class="svg-inline--fa fa-trash-alt fa-w-14"
```

Are these not required here? I think this should be consistent. I think common properties can be applied to the slot in crown button instead of specifying them individually on each `img`.

src/components/sequenceFlowButton.vue
```diff
@@ -1,18 +1,22 @@
...
-    height="20px"
-    alt="Create a sequence flow"
+    aria-label="Create a sequence flow"
+    @click="$emit('click')"
   >
+    <img
+      :src="sequenceFlow"
+      alt=""
+      role="presentation"
+    >
```
chipit24: Same here, different properties as the other buttons.

src/components/crownButton.vue
lisa-fehr: Thanks, that's what I was looking for.

src/components/messageFlowButton.vue
lisa-fehr: The trash icon seems to use font-awesome. I wonder if it could be simplified.

lisa-fehr:
```
aria-hidden="true"
data-prefix="fas"
data-icon="trash-alt"
class="svg-inline--fa fa-trash-alt fa-w-14"
```

Those properties are just for the trash icon. I can't seem to remove them. They were originally in the svg file, but they aren't valid there.

Any thoughts on what to do with it?

JCook21: You could allow the crown button component to receive an optional array of key value pairs as a prop. These could then be iterated over to set any additional properties.

lisa-fehr: Not sure how you'd inject those into the image on crownButton. Seems like a lot of work for 1 out of 6 icons. Maybe create the image in javascript and use v-html on the crownButton?

JCook21: I'm not sure I understand: what is the problem with leaving this as is?

lisa-fehr: I'm fine with the slot to override the image for now. It would be nice if it was the same as the other svqs at some point.

PR #849, opened 2019-11-22
Hide validation button unless auto-validate is enabled

JCook21:
Changes to hide the validation button unless auto validate is enabled.

Screen Shot 2019-11-22 at 11 49 35 AM
Screen Shot 2019-11-22 at 11 49 46 AM

### DIFF
--- a/src/components/ValidationStatus.vue
+++ b/src/components/ValidationStatus.vue
@@ -1,5 +1,24 @@
 <template>
   <div class="status-bar-container d-flex align-items-center justify-content-end">
+
+    <template v-if="autoValidate">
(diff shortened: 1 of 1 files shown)


lisa-fehr:
This removes showing warnings right away. And shows them after clicking validate. Might just want to ask if this is ok behavior. Originally it was a bunch of popovers.

JCook21:
> This removes showing warnings right away. And shows them after clicking validate. Might just want to ask if this is ok behavior. Originally it was a bunch of popovers.

This is true, although that behaviour change is a result of restyling the error/warnings panel. This is something we should address with Alan.

PR #851, opened 2019-11-25
Add "element switch" dropdown to start event

### LINKED ISSUE #850: Convert Start Event to use element select dropdown
Screen Shot 2019-11-25 at 9 48 00 AM

**Acceptance criteria**
As a user:
 * I should only be able to drag on a single type of start event.
 * When I add a new start event to the diagram the crown displays an icon that toggles a dropdown when clicked.
    * The dropdown allows me to choose a specific start event type.
    * When I choose a specific start event type the generic start event is replaced with the start event that I choose.
    * The dropdown isn't available when editing the start event.

Tasks:
 * Remove the start timer and message start events from the controls list, keeping only "Start Event".
    * Set `control` to `false` in these element's config file.
 * Add the cog icon to the start event's crown, to the left of the trash icon.
 * Show a dropdown when the cog icon is pressed, that displays two options, one for start timer event, and another for message start event.
    * Create a dropdown button component that emits an event when one of the start event options is selected.
 * Selecting one of the options in the dropdown will replace the generic start event with the one selected in the dropdown.
    * TBD: Create a node manager class? Where should the emitted event be handled?
 * Only show the cog icon when the start event is initially dropped onto the diagram.
 * Fix broken tests for the start event.

chipit24:
Fixes #850.

### DIFF
--- a/src/components/nodes/startEvent/startEvent.vue
+++ b/src/components/nodes/startEvent/startEvent.vue
@@ -14,7 +14,20 @@
     @remove-node="$emit('remove-node', $event)"
     @add-node="$emit('add-node', $event)"
     @save-state="$emit('save-state', $event)"
-  />
+  >
+    <template v-slot:dropdown>
+      <div class="dropdown-container">
+        <button class="button" @click="dropdownOpen = !dropdownOpen">
+          <i class="fas fa-cog"/>
+        </button>
+
+        <select v-if="dropdownOpen" class="dropdown" @change="$emit('replace-node', node, $event.target.value)">
+          <option value=""/>
+          <option value="processmaker-modeler-start-timer-event">Start Timer Event</option>
+          <option value="processmaker-modeler-message-start-event">Message Start Event</option>
+        </select></div>
+    </template>
+  </crown-config>
 </template>
 
 <script>
@@ -48,6 +61,7 @@ export default {
     return {
       shape: null,
       definition: null,
+      dropdownOpen: true,
     };
   },
   watch: {
@@ -83,3 +97,24 @@ export default {
   },
 };
 </script>
+
+<style scoped lang="scss">
+  .dropdown-container {
+    position: relative;
+    display: flex;
+    margin: 0 10px;
+
+    .button {
+      background: none;
+      border: none;
+      color: white;
+      padding: 0;
+    }
+
+    .dropdown {
+      position: absolute;
+      left: 0;
+      top: 2rem;
+    }
+  }
+</style>
--- a/src/components/nodes/startEvent/startEvent.vue
+++ b/src/components/nodes/startEvent/startEvent.vue
@@ -11,6 +11,7 @@
     :process-node="processNode"
     :plane-elements="planeElements"
     :is-rendering="isRendering"
+    :showDropdown="showDropdown"
     @remove-node="$emit('remove-node', $event)"
     @add-node="$emit('add-node', $event)"
     @save-state="$emit('save-state', $event)"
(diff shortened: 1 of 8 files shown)


src/components/nodes/startEvent/startEvent.vue
```diff
@@ -14,7 +14,20 @@
...
+        <button class="button" @click="dropdownOpen = !dropdownOpen">
+          <i class="fas fa-cog"/>
+        </button>
+
+        <select v-if="dropdownOpen" class="dropdown" @change="$emit('replace-node', node, $event.target.value)">
+          <option value=""/>
+          <option value="processmaker-modeler-start-timer-event">Start Timer Event</option>
+          <option value="processmaker-modeler-message-start-event">Message Start Event</option>
+        </select></div>
+    </template>
```
chipit24: I added this as a named slot just to get the initial implementation added quickly. It may be better to add this to the crown, especially since the other start events extend this component and so it will appear for all start events that way it is now.

The one advantage of this is that is eliminates the need for the crown config to pass down props and emit up events that it doesn't need to know about (in this case, it's just the `replace-node` event, though that could of course change).

src/components/nodes/startEvent/startEvent.vue
```diff
@@ -48,13 +50,35 @@ export default {
...
+    },
+  },
+  methods: {
+    removeDropdownOnUnhighlight() {
+      const unwatch = this.$watch('highlighted', highlighted => {
+        if (!highlighted) {
+          this.showDropdown = false;
+          unwatch();
+        }
+      });
```
JCook21: I'm not sure I understand this syntax. Why is there a variable 'unpatch' here? It doesn't appear to be used...

src/components/nodes/startEvent/startEvent.vue
```diff
@@ -48,13 +50,35 @@ export default {
...
+  },
   watch: {
     'node.definition.name'(name) {
       this.shape.attr('label/text', name);
     },
   },
+  created() {
+    if (this.isBaseStartEvent && this.highlighted) {
+      this.showDropdown = true;
+      this.removeDropdownOnUnhighlight();
```
JCook21: I'm not sure if having a separate method here really makes things much clearer. What doesn't anyone else think? Would this be clearer if it was inlined?

src/components/nodes/startEvent/startEvent.vue
chipit24: From https://vuejs.org/v2/api/#vm-watch:
> `vm.$watch` returns an unwatch function that stops firing the callback:
> ```js
> var unwatch = vm.$watch('a', cb)
> // later, teardown the watcher
> unwatch()
> ```

Except in our case I need to use the unwatch function within the watcher's own callback. It's like a "watch only once" method. Once the dropdown is hidden I didn't want to keep running this watcher as it should only have to set `showDropdown` to `false` once.

I could also keep it set up as a regular watcher and add an early return like:
```js
if (!this.showDropdown) {
  return;
}
```

chipit24: I think I would prefer using a regular watcher with an early return; the way I set it up with unwatch here isn't the greatest.

JCook21: I think a regular watcher would be a little clearer. The current code does have a high 'wtf per minute' count.... ;P

JCook21: With the greatest respect, it seems to be over clever. :)

PR #853, opened 2019-11-26
Control error highlighting with validation toggle

JCook21:
This PR ensures that invalid nodes are only highlighted when the auto-validate toggle is enabled.

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -104,6 +104,7 @@
       :root-elements="definitions.get('rootElements')"
       :isRendering="isRendering"
       :paperManager="paperManager"
+      :auto-validate="autoValidate"
       @add-node="addNode"
       @remove-node="removeNode"
       @set-cursor="cursor = $event"
@@ -138,15 +139,7 @@ import runningInCypressTest from '@/runningInCypressTest';
 import getValidationProperties from '@/targetValidationUtils';
 import MiniPaper from '@/components/MiniPaper';
 
-import {
-  faCompress,
-  faExpand,
-  faMapMarked,
-  faMinus,
-  faPlus,
-  faStepBackward,
-  faStepForward,
-} from '@fortawesome/free-solid-svg-icons';
+import { faCompress, faExpand, faMapMarked, faMinus, faPlus, faStepBackward, faStepForward } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 import { id as laneId } from './nodes/poolLane';
 import { id as sequenceFlowId } from './nodes/sequenceFlow';
@@ -267,10 +260,8 @@ export default {
     },
     highlightedNode: () => store.getters.highlightedNode,
     invalidNodes() {
-      return Object.entries(this.validationErrors).reduce((invalidIds, [, errors]) => {
-        invalidIds.push(...errors.map(error => error.id));
-        return invalidIds;
-      }, []);
+      return Object.entries(this.validationErrors)
+        .flatMap(([, errors]) => errors.map(error => error.id));
     },
   },
   methods: {
--- a/src/components/paperManager.js
+++ b/src/components/paperManager.js
@@ -1,4 +1,4 @@
-import { dia } from 'jointjs';
+import { dia, util } from 'jointjs';
 import { defaultNodeColor, invalidNodeColor } from '@/components/nodeColors';
 
 export default class PaperManager {
@@ -92,5 +92,9 @@ export default class PaperManager {
 
   setStateValid() {
     this.paper.drawBackground({ color: defaultNodeColor });
+  }
+
+  hasScheduledUpdates() {
+    return this.#paper._updates.priorities.some(updates => !util.isEmpty(updates));
   }
 }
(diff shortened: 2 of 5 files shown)


src/components/Modeler.vue
```diff
@@ -267,10 +260,8 @@ export default {
     },
     highlightedNode: () => store.getters.highlightedNode,
     invalidNodes() {
-      return Object.entries(this.validationErrors).reduce((invalidIds, [, errors]) => {
-        invalidIds.push(...errors.map(error => error.id));
-        return invalidIds;
-      }, []);
+      return Object.entries(this.validationErrors)
+        .flatMap(([, errors]) => errors.map(error => error.id));
```
JCook21: I felt this was slightly simpler to read than the reduce, which is functionally equivalent.

src/components/paperManager.js
```diff
@@ -93,4 +93,8 @@ export default class PaperManager {
   setStateValid() {
     this.paper.drawBackground({ color: defaultNodeColor });
   }
+
+  hasScheduledUpdates() {
+    return this.#paper._updates.priorities.some(updates => !util.isEmpty(updates));
```
JCook21: @chipit24 found this solution online. Some browsers (Safari...) were not properly detecting if updates needed to be rendered from the 'isRendering' property. This code works properly and gets around that issue. We should maybe look at using this more consistently in the project.

src/main.js
```diff
@@ -5,17 +5,16 @@ import ModelerApp from './ModelerApp.vue';
...
-
 import 'bootstrap/dist/css/bootstrap.css';
 import 'bootstrap-vue/dist/bootstrap-vue.css';
 
 import '@fortawesome/fontawesome-free/css/all.min.css';
 
+Vue.use(BootstrapVue);
+
+Vue.use(VueDeepSet);
+
```
JCook21: I made some edits to this file earlier but thought I'd leave this auto-rearranging in.

src/mixins/highlightConfig.js
```diff
@@ -55,49 +58,42 @@ export default {
...
-      if (this.hasError) {
+      if (this.hasError && this.autoValidate) {
         this.shapeView.highlight(null, { highlighter: errorHighlighter });
-      } else {
-        this.shapeView.unhighlight(null, { highlighter: errorHighlighter });
+        return;
       }
+      this.shapeView.unhighlight(null, { highlighter: errorHighlighter });
     },
-    configureCrown() {
```
JCook21: This method was really just being called from one place so I inlined it.

src/mixins/highlightConfig.js
```diff
@@ -55,49 +58,42 @@ export default {
...
-      this.paper.once('render:done', () => this.setHighlight());
```
JCook21: This callback was unnecessary since the mounted method checks if rendering is still happening before calling this method.

src/mixins/highlightConfig.js
```diff
@@ -55,49 +58,42 @@ export default {
...
-
-      const shapeView = this.shape.findView(this.paper);
+    setUpCrownConfig() {
```
lisa-fehr: I guess this could be `setUpHighlightConfig` now. Instead of the method name it was originally under in the crownConfig.

JCook21: Good point!

src/components/paperManager.js
chipit24: I didn't really find it, I asked for it since I created that issue in the jointjs repo :p I find they are quick to reply if you post an issue at https://github.com/clientIO/joint/issues.

JCook21: In either case, you take the credit Rob. :)

src/mixins/highlightConfig.js
lisa-fehr: I'm seeing that the highlight is now a square. Any ideas which change caused it?

src/mixins/highlightConfig.js
```diff
@@ -55,52 +58,45 @@ export default {
...
     this.$nextTick(() => {
       /* Use nextTick to ensure this code runs after the component it is mixed into mounts.
        * This will ensure this.shape is defined. */
 
-      if (this.isRendering) {
-        this.paper.once('render:done', this.setUpCrownConfig);
-      } else {
-        this.setUpCrownConfig();
+      if (this.paperManager.hasScheduledUpdates()) {
+        this.paperManager.addOnceHandler('render:done', this.setUpHighlightConfig());
```
chipit24: The 2nd argument for `addOnceHandler` needs to be a function reference, you are currently calling the method there instead.
```suggestion
        this.paperManager.addOnceHandler('render:done', this.setUpHighlightConfig);
```

src/mixins/highlightConfig.js
```diff
@@ -14,7 +14,7 @@ const defaultHighlighter = {
   name: 'stroke',
   options: {
     attrs: {
-      stroke: '#feb663',
+      stroke: '#5096db',
```
JCook21: This is the same colour as the crown background.

lisa-fehr:
Not huge. But when you drag on a new item while validating, the highlight is behind the red error highlight. Makes it seem purple at first.
Screen Shot 2019-11-26 at 1 21 23 PM

Screen Shot 2019-11-26 at 1 21 31 PM

JCook21:
> Not huge. But when you drag on a new item while validating, the highlight is behind the red error highlight. Makes it seem purple at first.
>

I think I've found why this was happening. It was setting the selected highlight before the error highlight, meaning that the red was on top of the blue making it look purple as the red is semi-transparent. I just reversed this ordering and I think it takes care of it.

PR #857, opened 2019-11-28
Fix issues with error highlighting after resizing

### LINKED ISSUE #470: Validation border render issue with lanes
**Issue**

When there is an error with lanes and the user attempts to resize pool/lane the red error borders will not update with the pool/lane and becomes fixed to the last size of the pool/lane.

**How to reproduce**
- Create a pool with a lane that has an error (missing label error)
- Resize the pool/lane

**Expected**
- Red error border adjust to the resize of the pool/lane

**Screenshot**
Screen Shot 2019-07-08 at 4 32 23 PM

**Video of bug**
https://drive.google.com/open?id=1esMRws8fJjKIEb8KG8dNLhJxAUx94Yki

JCook21:
Closes #470.

Resolves an issue where error highlights were not changing after an element is resized. After this I decided to refactor the code to simplify things.

### DIFF
--- a/src/components/paperManager.js
+++ b/src/components/paperManager.js
@@ -94,7 +94,11 @@ export default class PaperManager {
     this.paper.drawBackground({ color: defaultNodeColor });
   }
 
-  hasScheduledUpdates() {
-    return this.#paper._updates.priorities.some(updates => !util.isEmpty(updates));
+  executeAfterScheduledUpdates(callback) {
+    if (this.#paper._updates.priorities.some(updates => !util.isEmpty(updates))) {
+      this.addOnceHandler('render:done', callback);
+      return;
+    }
+    callback();
   }
 }
--- a/src/mixins/highlightConfig.js
+++ b/src/mixins/highlightConfig.js
@@ -1,21 +1,25 @@
 const errorHighlighter = {
-  name: 'stroke',
-  options: {
-    padding: 10,
-    attrs: {
-      stroke: 'red',
-      'stroke-width': 10,
-      opacity: 0.3,
+  highlighter: {
+    name: 'stroke',
+    options: {
+      padding: 10,
+      attrs: {
+        stroke: 'red',
+        'stroke-width': 10,
+        opacity: 0.3,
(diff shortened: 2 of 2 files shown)


src/mixins/highlightConfig.js
```diff
@@ -56,53 +56,28 @@ export default {
...
-    setHighlight() {
-      if (this.highlighted) {
-        this.shapeView.highlight(this.shapeBody, { highlighter: defaultHighlighter });
-        return;
-      }
-
-      this.shapeView.unhighlight(this.shapeBody, { highlighter: defaultHighlighter });
-    },
-    setErrorHighlight() {
+    setShapeHighlight() {
```
JCook21: I decided to merge the highlighting code into a single method for simplicity.

src/mixins/highlightConfig.js
```diff
@@ -56,53 +56,28 @@ export default {
...
-      this.shape.on('change:size', () => {
-        if (this.paperManager.hasScheduledUpdates()) {
-          this.paperManager.addOnceHandler('render:done', this.reHighlightExpandedElement);
-          return;
-        }
-        this.reHighlightExpandedElement();
-      });
-    },
-    reHighlightExpandedElement() {
+      this.shapeView.unhighlight(this.shapeBody, defaultHighlighter);
```
JCook21: Unhighlighting and then rehighlighting doesn't seem to cause any issues with flicker, etc. It seemed to be simpler and more consistent to bring this all into a single method.

src/components/paperManager.js
```diff
@@ -94,7 +94,11 @@ export default class PaperManager {
     this.paper.drawBackground({ color: defaultNodeColor });
   }
 
-  hasScheduledUpdates() {
-    return this.#paper._updates.priorities.some(updates => !util.isEmpty(updates));
+  executeAfterScheduledUpdates(callback) {
```
JCook21: This gets rid of some duplication and better encapsulates the logic of how this callback is handled.

PR #861, opened 2019-11-28
Fix text annotation highlighting.

JCook21:
Previously highlighting was not resizing when adding more text to a text annotation. This PR fixes that.


JCook21:
> Still a little off when slowly typing a few characters over. Or deleting characters slowly. But it is an improvement.
> 
> **Deleting:**
> Screen Shot 2019-11-28 at 4 08 22 PM
> **Typing slow:**
> Screen Shot 2019-11-28 at 4 07 12 PM

Yeah, it's weird. It's like the first character on a newline doesn't trigger the re-highlight but the second does. I tried a few things but couldn't improve this.

PR #866, opened 2019-12-02
Refactor paper parse function

chipit24:
This PR is for a slight refactor of the `renderPaper` method.

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -612,14 +612,6 @@ export default {
 
       this.isRendering = false;
       this.$emit('parsed');
-      this.removeLoaderIfProcessIsEmpty();
-    },
-    removeLoaderIfProcessIsEmpty() {
-      const emptyProcess = store.getters.nodes.length === 0;
-      if (!emptyProcess) {
-        return;
-      }
-      this.isRendering = false;
     },
     loadXML(xml = this.currentXML) {
       this.moddle.fromXML(xml, (err, definitions) => {
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -604,8 +604,10 @@ export default {
     async renderPaper() {
       this.isRendering = true;
 
-      await this.waitForCursorToChange();
-      await this.paperManager.performAtomicAction(() => this.parse());
+      await this.paperManager.performAtomicAction(async() => {
+        await this.waitForCursorToChange();
+        this.parse();
+      });
       await this.paperManager.awaitScheduledUpdates();
 
       this.isRendering = false;
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -606,8 +606,9 @@ export default {
       await this.paperManager.performAtomicAction(async() => {
         this.isRendering = true;
         await this.waitForCursorToChange();
-        this.paperManager.awaitScheduledUpdates().then(() => this.isRendering = false);
         this.parse();
+        await this.paperManager.awaitScheduledUpdates();
+        this.isRendering = false;
       });
       this.$emit('parsed');
       this.removeLoaderIfProcessIsEmpty();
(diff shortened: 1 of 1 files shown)


src/components/Modeler.vue
```diff
@@ -602,22 +610,16 @@ export default {
...
-      this.removeLoaderIfProcessIsEmpty();
-    },
-    removeLoaderIfProcessIsEmpty() {
-      const emptyProcess = store.getters.nodes.length === 0;
-      if (!emptyProcess) {
-        return;
-      }
+      await this.paperManager.awaitScheduledUpdates();
+
       this.isRendering = false;
```
chipit24: Since we're now always setting `isRendering` to `false`, it seems the `removeLoaderIfProcessIsEmpty` method is no longer required, as that method conditionally set  the flag to false.

src/components/Modeler.vue
```diff
@@ -602,22 +610,16 @@ export default {
       return new Promise(resolve => setTimeout(resolve, cursorWaitTime));
     },
     async renderPaper() {
-      await this.$nextTick();
+      this.isRendering = true;
+
       await this.paperManager.performAtomicAction(async() => {
-        this.isRendering = true;
         await this.waitForCursorToChange();
-        this.paperManager.awaitScheduledUpdates().then(() => this.isRendering = false);
```
chipit24: Since we're in an `async` function, and to keep things consistent, I switched to use `await`. This made it easier for me to read.

src/components/Modeler.vue
```diff
@@ -602,22 +610,16 @@ export default {
       return new Promise(resolve => setTimeout(resolve, cursorWaitTime));
     },
     async renderPaper() {
-      await this.$nextTick();
```
chipit24: A call to `$nextTick` was already being made in `waitForCursorToChange`, so calling it here seemed redundant.

src/components/Modeler.vue
```diff
@@ -602,22 +610,16 @@ export default {
...
       });
-      this.$emit('parsed');
-      this.removeLoaderIfProcessIsEmpty();
-    },
-    removeLoaderIfProcessIsEmpty() {
-      const emptyProcess = store.getters.nodes.length === 0;
-      if (!emptyProcess) {
-        return;
-      }
+      await this.paperManager.awaitScheduledUpdates();
```
JCook21: To be honest I find this style harder to read, although that's a matter of personal preference. The naming is making this clunkier too: the method name `awaitScheduledUpdates()` works well when using chaining but it reads terribly when using await. Every other call to this method is using chaining too so I'm not sure if we need to address that for consistency.

chipit24: The name can definitely be improved, `await awaitScheduledUpdates` reads a little funny, but naming is hard so I left that for now.

> Every other call to this method is using chaining too so I'm not sure if we need to address that for consistency.

I'm not sure I understand you on this point; other calls are using `await` and not chaining `.then` blocks. Having a sole `.then` block in an `async` function surrounded by `await` calls looks odd and it's easy to think that execution will pause on the `.then` when it doesn't.

One of the benefits of async/await is that it allows functions like this to be composed in a way that is flat and _reads_ like synchronous code. I found it made reasoning about how `isRendering` changes more straightforward. However, readability suffers when you have to wrap it all in a try/catch block (as oppose to a `.catch`), or when you're not actually working with resolved values, for example.

That's not to say that changing `.then` blocks looks ugly; it can read very nicely too if there's no nesting of `.then` blocks within one another. Take for example this block:

```js
await this.paperManager.performAtomicAction(async() => {
  await this.waitForCursorToChange();
  this.parse();
});
```

This still looks a little odd to me too and is akin to nesting `.then` blocks. In any case, the point of this is to improve readability and if it's not doing that then let's discuss in person.

PR #872, opened 2019-12-02
Only show dropdown on initial render

### LINKED ISSUE #871: Hide dropdown selection on start event
The new dropdown in the crown config should only be available when a start event is initially dropped onto the diagram. As soon as the user clicks away or interacts with the start event in some other way the dropdown should not appear. NOTE: this should cover the use case of the dropdown icon not appearing on anything other than a generic start event.

chipit24:
Fixes #871.

Todo:
- [x] Only show the dropdown for the start event.
- [x] Remove the dropdown when the element is unhighlighted.
- [x] Do not show the dropdown when the element is swapped (only show it for the base start event).
- [x] Add tests.

### DIFF
--- a/src/components/nodes/startEvent/startEvent.vue
+++ b/src/components/nodes/startEvent/startEvent.vue
@@ -58,6 +58,19 @@ export default {
       this.shape.attr('label/text', name);
     },
   },
+  methods: {
+    removeDropdownOnUnhighlight() {
+      const unwatch = this.$watch('highlighted', highlighted => {
+        if (!highlighted) {
+          this.showDropdown = false;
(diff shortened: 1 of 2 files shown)


chipit24:
> Looks alright. But if I upload a xml where the start event is connected to something, I can switch it. Could be another issue, or fixed here?
 
I can fix it as part of this PR. The correct solution would be that the start event can only be switched when initially dragged onto the paper. So even the default start event that renders cannot be switched.

That brings up another point: When creating a new process in core, it's blank, but in standalone we render a start event. I think we should keep this consistent and not render a default start event in standalone.

PR #876, opened 2019-12-02
Fix skipped tests

### LINKED ISSUE #864: Skipped tests
Fix skipped tests or remove if no longer relevant.

JCook21:
Closes #864 

- Fix test that is skipped
- Remove test that depends on ProcessMaker core
- Update test that depends on ProcessMaker core to run independently
- Add test case to assert that a text annotation can be added to a pool

### DIFF
--- a/src/components/inspectors/DateTimeExpression.vue
+++ b/src/components/inspectors/DateTimeExpression.vue
@@ -8,6 +8,7 @@
       class="p-0"
       :value="convertFromUTC(value)"
       @input="emitValue"
+      data-test="date-picker"
     />
   </div>
 </template>
--- a/tests/e2e/specs/IntermediateCatchEvent.spec.js
+++ b/tests/e2e/specs/IntermediateCatchEvent.spec.js
@@ -33,7 +33,7 @@ describe('Intermediate Catch Event', () => {
       .should('have', validIntermediateCatchEventXML.trim());
   });
 
-  it.skip('Update date/time field on Intermediate Catch Event', () => {
+  it('Update date/time field on Intermediate Catch Event', () => {
     const intermediateCatchEventPosition = { x: 250, y: 250 };
     dragFromSourceToDest(nodeTypes.intermediateCatchEvent, intermediateCatchEventPosition);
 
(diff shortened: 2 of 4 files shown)


src/components/inspectors/DateTimeExpression.vue
```diff
@@ -8,6 +8,7 @@
       class="p-0"
       :value="convertFromUTC(value)"
       @input="emitValue"
+      data-test="date-picker"
```
JCook21: This seemed to be the easiest solution to selecting this element in a test. Let me know if there's a better way.

tests/e2e/specs/Tasks.spec.js
```diff
@@ -118,25 +118,4 @@ describe('Tasks', () => {
...
-    cy.contains('Task Assignment').next().select('User');
-    cy.contains('Assigned User')
-      .should('exist')
-      .next().find('.multiselect').should('exist');
-
-    cy.contains('Task Assignment').next().select('Group');
-    cy.contains('Assigned Group')
-      .should('exist')
-      .next().find('.multiselect').should('exist');
-  });
```
JCook21: This test seems to rely completely on ProcessMaker core so I thought deleting it was the best option.

tests/e2e/specs/TextAnnotation.spec.js
```diff
@@ -44,18 +40,54 @@ describe('Text Annotation', () => {
...
+    <bpmn:association id="node_8" associationDirection="None" sourceRef="node_6" targetRef="node_7" />`;
+
+    cy.get('[data-test=downloadXMLBtn]').click();
+    cy.window()
+      .its('xml')
+      .then(xml => xml.trim())
+      .then(xml => {
+        expect(xml).to.contain(expectedXML.trim());
+      });
+  });
```
JCook21: This change adds something we can assert about outside of core.

tests/e2e/specs/IntermediateCatchEvent.spec.js
```diff
@@ -42,14 +42,11 @@ describe('Intermediate Catch Event', () => {
...
     typeIntoTextInput(nameInput, testString);
-
     cy.contains('Timing Control').click();
     cy.get('[data-test=intermediateTypeSelect]').select('Date/Time');
-    const startDateValue = '2019-02-27';
-    typeIntoTextInput('.start-date', startDateValue);
-
-    const startTimeValue = '00:30';
-    cy.get('[data-test=startTime]').select(startTimeValue, { force: true });
+    const startDateTime = '02/27/2019 12:30 AM';
```
lisa-fehr: When I looked at this test, I tried to click the time in the new date/time picker. It just didn't seem easy to do. So I'm ok with this.

tests/e2e/specs/TextAnnotation.spec.js
chipit24: We have a function called `assertDownloadedXmlContainsExpected` in `utils.js`, could that be used here?

tests/e2e/specs/TextAnnotation.spec.js
```diff
@@ -44,18 +40,54 @@ describe('Text Annotation', () => {
...
+    </bpmn:textAnnotation>
+    <bpmn:association id="node_4" associationDirection="None" sourceRef="node_3" targetRef="node_2" />
+  </bpmn:process>`;
+
+    cy.get('[data-test=downloadXMLBtn]').click();
+    cy.window()
+      .its('xml')
+      .then(xml => xml.trim())
+      .then(xml => {
+        expect(xml).to.contain(expectedXML.trim());
```
chipit24: Same comment here regarding `assertDownloadedXmlContainsExpected`.

PR #877, opened 2019-12-02
Add tooltips for hide/show menu button

### LINKED ISSUE #867: Hide Menus Tooltip
Add a tooltip to the icon for hiding menus.

Hide Menus
Show Menus 

Make sure they are translated.

chipit24:
Fixes #867.

Screen Shot 2019-12-02 at 4 32 43 PM
Screen Shot 2019-12-02 at 4 32 31 PM

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -27,24 +27,31 @@
     >
       <div class="toolbar d-inline-block mt-3 position-relative" role="toolbar" aria-label="Toolbar" :class="{ 'ignore-pointer': canvasDragPosition }">
         <div class="btn-group btn-group-sm mr-2" role="group" aria-label="Undo/redo controls">
-          <span id="undo-wrapper" role="button" class="d-inline-block" tabindex="0" @click="undo">
-            <button type="button" class="btn btn-sm btn-secondary btn-undo" :disabled="!canUndo" data-test="undo">
-              <font-awesome-icon :icon="undoIcon" />
-            </button>
-          </span>
-          <b-tooltip target="undo-wrapper">{{ $t('Undo') }}</b-tooltip>
+          <b-button
+            class="btn btn-sm btn-secondary btn-undo"
+            :disabled="!canUndo"
+            data-test="undo"
+            v-b-tooltip.hover
+            :title="$t('Undo')"
+            @click="undo"
+          >
+            <font-awesome-icon :icon="undoIcon" />
+          </b-button>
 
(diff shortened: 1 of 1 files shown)


src/components/Modeler.vue
```diff
@@ -139,7 +145,15 @@ import runningInCypressTest from '@/runningInCypressTest';
...
-import { faCompress, faExpand, faMapMarked, faMinus, faPlus, faStepBackward, faStepForward } from '@fortawesome/free-solid-svg-icons';
+import {
+  faCompress,
+  faExpand,
+  faMapMarked,
+  faMinus,
+  faPlus,
+  faStepBackward,
+  faStepForward,
+} from '@fortawesome/free-solid-svg-icons';
```
JCook21: Do we have a preferred code style for this? Both seem valid for eslint and it would be good to have a consistent standard.

chipit24: The related rules I could find are https://eslint.org/docs/rules/sort-imports and https://eslint.org/docs/rules/object-curly-newline. However, there doesn't appear to be a rule to split imports into multi lines when they reach a certain max length (related: https://github.com/airbnb/javascript/issues/1051). So I'm fine letting PHPStorm handle this for now.

JCook21: > The related rules I could find are https://eslint.org/docs/rules/sort-imports and https://eslint.org/docs/rules/object-curly-newline. However, there doesn't appear to be a rule to split imports into multi lines when they reach a certain max length (related: [airbnb/javascript#1051](https://github.com/airbnb/javascript/issues/1051)). So I'm fine letting PHPStorm handle this for now.

To be clear, I don't mind what I do but it should be consistent between the four of us. Can we arrive at some sort of standard? There must be a setting in PHPStorm that's breaking these onto multiple lines...

JCook21: Just to further this, I've just come across a case where having your PHPStorm apply this formatting is causing a merge conflict. I don't think the rest of us have the setting enabled that's causing this formatting: can you look into it @chipit24?

chipit24:
The single test failure is a little odd; it seems unrelated to the changes made in this PR, and the test passes for me locally. I'm going to re-run it on CI.

src/components/Modeler.vue
chipit24: I don't remember changing this setting specifically in PHPStorm so I assumed it was the same for all of us; I'll update my settings!

Issue #869, opened 2019-12-02
Fix test failures for new start event menu


JCook21:
When the new start event is fully converted to the new selection system fix test failures.

lisa-fehr:
Solved in #874

PR #878, opened 2019-12-02
Style event list

### LINKED ISSUE #870: Style event selection dropdown
Add styling for the event selection dropdown to the start events. Currently this is implemented as a select input but it should be converted to a list of buttons.

image

tuantruong625:
The purpose of this PR is to style the element list.

image

 Note: Tests will be added once the other #869  and #871  are merged into #850.

Fixes #870

### DIFF
--- a/src/components/crownConfig.vue
+++ b/src/components/crownConfig.vue
@@ -51,8 +51,6 @@
       </ul>
     </div>
 
-
-
     <delete-button
       @click="removeShape"
       class="crown-config__icon"
@@ -314,6 +312,30 @@ export default {
 
 <style lang="scss">
   $primary-color: #5096db;
+  $primary-light: #fff;
+
+  $element-list-top: 2.75rem;
+  $element-list-left: -0.75rem;
+  $element-list-top-chevron: -0.3rem;
+  $element-list-left-chevron: 0.6rem;
+  $crown-top-chevron: 0.75rem;
+  $crown-left-chevron: 0.45rem;
+
+  $chevron-width: 1.25rem;
+  $chevron-height: 1.25rem;
+
+  @mixin chevron($top, $left) {
+    content: '';
+    background-color: $primary-color;
+    width: $chevron-width;
+    height: $chevron-height;
+    position: absolute;
+    top: $top;
+    left: $left;
+    z-index: -1;
+    transform: rotate(45deg);
+    border-radius: 1px;
+  }
 
   .crown-config {
     background-color: $primary-color;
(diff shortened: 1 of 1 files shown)


src/components/crownConfig.vue
```diff
@@ -23,24 +23,26 @@
...
-    <div class="dropdown-container">
-      <button class="button" @click="dropdownOpen = !dropdownOpen">
+    <div class="cog-container" role="menubar">
+      <button class="cog-container--button" @click="dropdownOpen = !dropdownOpen">
         <i class="fas fa-cog"/>
       </button>
 
-      <ul class="dropdown" v-if="dropdownOpen">
-        <li>
+      <ul class="element-list" v-if="dropdownOpen" role="list">
```
chipit24: From https://www.w3.org/TR/html-aria:
> Note that some user agents suppress a list's implicit ARIA semantics if list markers are removed. Authors can use role=list to reinstate the role, if necessary.

So good call on specifying the role attribute here.

src/components/crownConfig.vue
```diff
@@ -23,24 +23,26 @@
...
+        <li class="element-list--item" role="listitem">
```
chipit24: The `li` element already has a default role of `listitem`, but considering my comment above, it may be best to explifictly define the role anyway.

I dug into it a little more more and came across this article: https://inclusive-components.design/menus-menu-buttons/ (the other articles on that website are also great). That page gives the following example:

```html
<button aria-haspopup="true" aria-expanded="false">
  Difficulty
  <span aria-hidden="true">&#x25be;</span>
</button>
<div role="menu">
  <button role="menuitem" tabindex="-1">Easy</button>
  <button role="menuitem" tabindex="-1">Medium</button>
  <button role="menuitem" tabindex="-1">Incredibly Hard</button>
</div>
```

https://www.w3.org/TR/wai-aria-1.1/#menu
https://www.w3.org/TR/wai-aria-1.1/#menuitem

src/components/crownConfig.vue
```diff
@@ -23,24 +23,26 @@
...
             data-test="switch-to-start-timer-event"
+            class="element-list--item__button"
             type="button"
             @click="$emit('replace-node', { node, typeToReplaceWith: 'processmaker-modeler-start-timer-event' })"
           >{{ $t('Start Timer Event') }}
           </button>
         </li>
 
-        <li>
+        <li class="element-list--item" role="listitems">
```
chipit24: Typo: listitem**s** -> listitem

PR #881, opened 2019-12-03
Removed default start event from standalone modeler.

### LINKED ISSUE #880: Remove default start event from standalone modeler.

JCook21:
Fixes #880 

- Stop standalone modeler rendering a default start event.
- Add a start event to `beforeEach()` for every test to ensure backwards compatibility.
- Edits to make a test keep passing.

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -168,7 +168,6 @@ import BpmnModdle from 'bpmn-moddle';
 import controls from './controls';
 import pull from 'lodash/pull';
 import remove from 'lodash/remove';
-import { startEvent } from '@/components/nodes';
 import store from '@/store';
 import InspectorPanel from '@/components/inspectors/InspectorPanel';
 import undoRedoStore from '@/undoRedoStore';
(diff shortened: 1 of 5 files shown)


tests/e2e/specs/UndoRedo.spec.js
```diff
@@ -105,6 +105,13 @@ describe('Undo/redo', () => {
     const startEventMoveToPosition = { x: 300, y: 300 };
 
     cy.get('[data-test=undo]')
+      .click()
+      .should('be.disabled');
+
+    waitToRenderAllShapes();
+
+    cy.get('[data-test=redo]')
+      .click()
```
JCook21: These changes are needed as the undo button is no longer initially disabled. This is because we've added a start event, which can be undone by clicking the undo button.

tests/e2e/support/index.js
```diff
@@ -29,4 +31,6 @@ Cypress.Cookies.defaults({
 
 beforeEach(() => {
   cy.loadModeler();
+  dragFromSourceToDest(nodeTypes.startEvent, { x: 150, y: 150 });
+  cy.get('.paper-container').click();
```
JCook21: This removes focus from the start event, which puts the paper in the same state that it was before these changes were made.

src/components/Modeler.vue
```diff
@@ -789,26 +788,6 @@ export default {
...
-      diagram.bounds.x = 150;
-      diagram.bounds.y = 150;
-
-      this.addNode({
-        definition,
-        diagram,
-        type: startEvent.id,
-      });
-
-      setTimeout(() => undoRedoStore.dispatch('resetHistory'));
```
chipit24: Since it seems it was only used here, the `resetHistory` action could be removed from the undo/redo store.

PR #885, opened 2019-12-04
Add dropdown to end events

### LINKED ISSUE #882: Convert End Event to use element select dropdown
Screen Shot 2019-11-25 at 9 48 00 AM

**Acceptance criteria**
As a user:
 * I should only be able to drag on a single type of end event.
 * When I add a new end event to the diagram the crown displays an icon that toggles a dropdown when clicked.
    * The dropdown allows me to choose a specific end event type.
    * When I choose a specific end event type the generic end event is replaced with the end event that I choose.
    * The dropdown isn't available when editing the end event.

Tasks:
 * Remove the message end event and the error end event from the controls list, keeping only "End Event".
    * Set `control` to `false` in these element's config file.
 * Add the cog icon to the end event's crown, to the left of the trash icon.
 * Show a dropdown when the cog icon is pressed, that displays two options, one for message end event, and another for error end event.
    * Create a dropdown button component that emits an event when one of the end event options is selected.
 * Selecting one of the options in the dropdown will replace the generic end event with the one selected in the dropdown.
 * Only show the cog icon when the end event is initially dropped onto the diagram.
 * Fix broken tests for the end event.

JCook21:
Closes #882 

Allow end events to use the new dropdown selection mechanism and refactoring.

### DIFF
--- a/src/components/nodes/startEvent/startEvent.vue
+++ b/src/components/nodes/startEvent/startEvent.vue
@@ -47,7 +47,6 @@ export default {
     return {
       shape: null,
       definition: null,
-      showDropdown: false,
       dropdownData: [
         {
           label: 'Start Timer Event',
--- a/src/components/nodes/startEvent/startEvent.vue
+++ b/src/components/nodes/startEvent/startEvent.vue
@@ -12,10 +12,7 @@
     :plane-elements="planeElements"
     :is-rendering="isRendering"
     :dropdown-data="dropdownData"
-    @remove-node="$emit('remove-node', $event)"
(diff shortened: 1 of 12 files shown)


src/components/nodes/startEvent/startEvent.vue
```diff
@@ -51,34 +48,25 @@ export default {
       shape: null,
       definition: null,
       showDropdown: false,
```
JCook21: I think this `showDropdown` prop can be removed. I'll look into it.

PR #891, opened 2019-12-05
Renamed all references of 'call activity' to 'sub process'

### LINKED ISSUE #890: Change name of 'Call Activity' element to 'Sub Process'.

JCook21:
Closes #890.


src/components/nodes/index.js
```diff
@@ -15,7 +15,7 @@ export { default as intermediateTimerEvent } from './intermediateTimerEvent';
 export { default as intermediateMessageCatchEvent } from './intermediateMessageCatchEvent';
 export { default as intermediateMessageThrowEvent } from './intermediateMessageThrowEvent';
 export { default as task } from './userTask';
-export { default as callActivity } from './callActivity';
```
chipit24: This will break imports in processmaker and in any connector that expects a "callActivity" named export. 

We could mark the callActivity export as deprecated for the next release, and keep both a callActivity and a subProcess export available (both pointing to the same file) to prevent a lot of problems and to give us time to fix the imports in process maker and possibly other repos (at least the ones we have control over).

JCook21: Good point, I'll do that.

PR #902, opened 2019-12-05
Accordion group

### LINKED ISSUE #898: Configuration Accordion
There seems to be an inconsistency in how the "Configuration" accordion behaves when a BPMN element is initially placed into the Process model: sometimes the "Configuration" panel is expanded by default, other times not.

Also when one selects an accordion, that accordion expands while any expanded accordion collapses.  Only one accordion can be open at a time.

chipit24:
Fixes #898.

This PR requires the changes in https://github.com/ProcessMaker/vue-form-elements/pull/129 to be merged and published for the feature to work completely. Merging this PR before https://github.com/ProcessMaker/vue-form-elements/pull/129 will not cause any braking changes, but until vue-form-elements is updated, accordions will still collapse and expand individually.

Video:
https://www.dropbox.com/s/ntte5irq018b8y4/collapsing-accordions.mov?dl=0

### DIFF
--- a/src/components/inspectors/InspectorPanel.vue
+++ b/src/components/inspectors/InspectorPanel.vue
@@ -1,6 +1,7 @@
 <template>
   <b-card no-body class="inspector-container border-top-0 border-bottom-0 border-right-0 rounded-0" data-test="inspector-container">
     <vue-form-renderer
+      :key="config[0].name"
       v-if="highlightedNode"
       :data="data"
       @update="updateDefinition"
--- a/src/components/inspectors/process.js
+++ b/src/components/inspectors/process.js
@@ -16,7 +16,7 @@ export default {
             initiallyOpen: true,
             label: 'Configuration',
             icon: 'cog',
-            name: 'configuration',
+            name: 'inspector-accordion',
           },
           items: [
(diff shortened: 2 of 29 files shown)


src/components/inspectors/InspectorPanel.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <b-card no-body class="inspector-container border-top-0 border-bottom-0 border-right-0 rounded-0" data-test="inspector-container">
     <vue-form-renderer
+      :key="config[0].name"
```
chipit24: This ensures the form-renderer components are re-rendered when a different config is loaded. This means that at least one default accordion will be opened when switching between different node types.

src/components/inspectors/process.js
```diff
@@ -16,7 +16,7 @@ export default {
             initiallyOpen: true,
             label: 'Configuration',
             icon: 'cog',
-            name: 'configuration',
```
chipit24: This name isn't displayed visually anywhere, but is required to be the same to group all accordions so they act in tandem (i.e. as one opens, the other closes).

PR #903, opened 2019-12-06
Add ordering for the controls panel

### LINKED ISSUE #895: Change order of components in the controls panel
Screen Shot 2019-12-05 at 11 43 04 AM

Modify the order of the elements so that the "grouped" elements (those with the new element select dropdown) are at the top. The order should be (from top to bottom):
 * Start Event
 * Intermediate Event
 * End Event
 * Task
 * Gateway
 * All other components ...

JCook21:
Closes #895 

This PR sorts controls in the control panel by a rank property. If a control has no rank it will be sorted lower than those that do.

### DIFF
--- a/src/components/Modeler.vue
+++ b/src/components/Modeler.vue
@@ -451,6 +451,22 @@ export default {
     registerBpmnExtension(namespace, extension) {
       this.extensions[namespace] = extension;
     },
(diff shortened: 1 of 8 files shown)


JCook21:
> As long as that is expected. This addresses the issue.

Completely my mistake, should be fixed now.

PR #905, opened 2019-12-06
Fix accordion sorting

### LINKED ISSUE #709: Advanced Accordion always last

chipit24:
Fixes #709.

### DIFF
--- a/tests/unit/inspectorExtensionManager.spec.js
+++ b/tests/unit/inspectorExtensionManager.spec.js
@@ -17,11 +17,14 @@ describe('Inspector Extension Manager', function() {
         ],
       }],
     };
+
     config = {
       id: 5,
       container: true,
       name: 'inspector-accordion',
-      label: 'Advanced',
+      config: {
+        label: 'Advanced',
+      },
     };
   });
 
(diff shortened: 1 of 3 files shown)


tests/unit/inspectorExtensionManager.spec.js
```diff
@@ -17,32 +17,31 @@ describe('Inspector Extension Manager', function() {
...
-    delete config.container;
-    const items = node.inspectorConfig[0].items;
-    node.inspectorConfig[0].items = [{ items }];
-
-    registerInspectorExtension(node, config);
-
-    const [lastItem] = node.inspectorConfig[0].items[0].items.slice(-1);
-    expect(lastItem).toBe(config);
-  });
-
```
chipit24: I removed this test as it wasn't adding any value. Containers in this context are the accordion containers, and we currently don't have any "Advanced" form input, and even if we did, there's no sorting requirement for inputs inside an accordion. So I instead modified the spec above to add another accordion after the Advanced one to confirm the correct behaviour.

lisa-fehr:
One curious thing:

Task doesn't seem to have a users list on processmaker? I might be out of sync.
Screen Shot 2019-12-06 at 4 11 45 PM

chipit24:
I don't think the Task in processmaker should have a user list; that was added as an inspector extension in standalone for testing. User's seem to be assigned through the Assignment Rules accordion on processmaker.

PR #909, opened 2019-12-06
Add transitions to modeler

### LINKED ISSUE #859: Investigate transitions
Add transition states for
- Collapsing menus, controls, the inspector panel
- Issues panel
- `BPMN Issues` button
- Filter search

JCook21:
Closes #859

### DIFF
--- a/src/components/MiniPaper.vue
+++ b/src/components/MiniPaper.vue
@@ -75,13 +75,14 @@ $transition: 0.3s;
     pointer-events: none;
   }
 }
-.mini-paper-expanded {
+
+.mini-paper-container.expanded {
   right: $mini-paper-container-right-expanded-position;
 }
 
 .mini-paper-container.closed {
   opacity: 0;
-  transition: opacity $transition linear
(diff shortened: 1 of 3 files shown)


src/components/MiniPaper.vue
```diff
@@ -26,9 +26,6 @@ export default {
     },
   },
   computed: {
-    visibilityClass() {
-      return this.isOpen ? 'visible' : 'invisible';
```
JCook21: The use of 'invisible' was stopping the transition from working in fading the mini-map out. I experimented and it seemed unnecessary so I removed it. I also though that having the method here didn't add much so I inlined it.

tuantruong625:
The only missing transition would the chevrons on the accordions and should follow the same pattern as the issues chevron.

Screen Shot 2019-12-09 at 1 22 50 PM

JCook21:
> The only missing transition would the chevrons on the accordions and should follow the same pattern as the issues chevron.
> 

Good point. I'll create a new issue to add those.

PR #910, opened 2019-12-09
Fix console error when adding lane

### LINKED ISSUE #908: Error when adding lane to pool containing boundary event
##  Description
A console error is thrown when adding a lane to a pool which contains a boundary event.

## Reproduction steps
* Drag a generic task onto the grid.
* Drag a boundary event onto the task.
* Drag a pool onto the grid.
* Add a lane to the pool.

Note: The order above is important. The error doesn't occur when the pool is added fist.

## Expected behaviour
No console errors are thrown.

## Actual behaviour
The following console error is thrown:
```
[Vue warn]: Error in nextTick: "TypeError: Cannot read property 'component' of null"
```

## Software used
Standalone modeler develop branch, Chrome 78, macOS Catalina.

chipit24:
Fixes #908.

### DIFF
--- a/src/components/nodes/pool/pool.vue
+++ b/src/components/nodes/pool/pool.vue
@@ -212,6 +212,58 @@ export default {
         this.updateLaneChildren();
       }
     },
+    positionLaneRelativeToPool(element, pool) {
+      const { width, height } = pool.getBBox();
+      const elementBounds = element.component.node.diagram.bounds;
+
+      if (elementBounds.x && elementBounds.y) {
+        /* If lane already has a position, don't re-position or re-size it. */
+        return;
+      }
+
+      const isFirstLane = pool.getEmbeddedCells().filter(cell => {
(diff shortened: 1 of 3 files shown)


src/components/nodes/pool/pool.vue
```diff
@@ -250,6 +251,7 @@ export default {
 
         this.getElementsUnderArea(element, this.graph).filter(laneElement => {
           return laneElement.component &&
```
JCook21: This expression is getting quite long and complicated. Could it be split into multiple calls to `filter()`? Another option would be to move it to a method and give it a descriptive name.

chipit24: I was going to move this whole block into another function as it's using an early return but it's hard to tell. I agree that this is a huge function that needs to be broken up though.

JCook21: Multiple calls to `.filter()` would make this more readable IMO.

PR #928, opened 2019-12-09
Add Boundary Timer Event and Error Event to Script Task

### LINKED ISSUE #921: Add boundary event dropdown for script task
Supported boundary event types:
 * Boundary Error Event
 * Boundary Timer Event

lisa-fehr:
Closes #921

### DIFF
--- a/src/components/nodes/task/task.vue
+++ b/src/components/nodes/task/task.vue
@@ -12,9 +12,7 @@
     :plane-elements="planeElements"
     :is-rendering="isRendering"
     :boundary-event-dropdown-data="boundaryEventDropdownData"
-    @remove-node="$emit('remove-node', $event)"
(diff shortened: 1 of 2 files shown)


src/components/nodes/task/task.vue
```diff
@@ -11,6 +11,7 @@
     :process-node="processNode"
     :plane-elements="planeElements"
     :is-rendering="isRendering"
+    :boundary-event-dropdown-data="boundaryEventDropdownData"
     @remove-node="$emit('remove-node', $event)"
     @add-node="$emit('add-node', $event)"
     @save-state="$emit('save-state', $event)"
```
chipit24: Something we've been doing is replacing all the individual event listeners to a single `v-on="$listeners"` binding. This may help clean up the component since these event are simply being emitted back up as-is.

PR #931, opened 2019-12-10
Updates to text in Modeler

JCook21:
Updates to label and helper text in Modeler.

### DIFF
--- a/src/components/nodes/boundaryEvent/index.js
+++ b/src/components/nodes/boundaryEvent/index.js
@@ -42,9 +42,7 @@ export default {
           items: [
             {
               component: 'FormInput',
-              config: {
-                ...nameConfigSettings,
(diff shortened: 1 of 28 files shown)


src/components/nodes/boundaryEvent/index.js
```diff
@@ -44,7 +44,6 @@ export default {
               component: 'FormInput',
               config: {
                 ...nameConfigSettings,
```
chipit24: I wonder if destructuring is still needed in this situation. Could you assign the config directly?
```js
config: nameConfigSettings,
```

This could apply to a few other places as well.

JCook21: Great point, thanks. Updated to do this and all seems good.

PR #948, opened 2019-12-12
Add boundary validation tests

### LINKED ISSUE #934: Add test for boundary event positioning and validation
Add tests to support the work done in https://github.com/ProcessMaker/modeler/pull/932. The tests should verify the following:

 * The validation correctly works when adding boundary events from the crown dropdown menu (the button to add the boundary event should be disabled if it's invalid).
   * All buttons should be disabled when all 12 ports are occupied by a boundary event.
   * The error boundary event button should  be disabled when there is already a boundary event on the target.
 * Clicking to add a boundary event should fill an empty port and not overlap a boundary event already on the target.

chipit24:
Fixes #934.

This PR adds tests to verify:
 * Boundary events are added to different positions around the crown when using the _add boundary event_ dropdown.
 * You cannot add more boundary events to the target shape using the boundary event dropdown if all 12 ports are occupied.
 * Boundary event buttons are disabled when you cannot add them to the target shape.

I also removed `boundaryEventSwitcher.js` as it was unused.


tests/e2e/specs/BoundaryEventValidatiion.spec.js
```diff
@@ -0,0 +1,44 @@
...
+import isEqual from 'lodash/isEqual';
+
+describe('Boundary event validation', function() {
+  it('should add boundary events to empty ports around boundary event target, and not allow adding any more', function() {
+    const taskPosition = { x: 250, y: 200 };
+    dragFromSourceToDest(nodeTypes.task, taskPosition);
+
+    const numberOfPortsAroundTask = 12;
+    for (let i = 0; i < numberOfPortsAroundTask; i++) {
+      setBoundaryEvent(nodeTypes.boundaryMessageEvent, taskPosition, nodeTypes.task);
```
lisa-fehr: nodeTypes.task is the default if you want to shorten these.

chipit24: I'm not sure I understand. Why would I probably want `nodeTypes.subProcess`?
For ports, the same validation should apply regardless of the specific target task type or boundary event type. The decision to use the base task type and the message boundary event was arbitrary.

lisa-fehr: This was before we merged in my ticket to add Message Events to all Task types. You can simplify it to:
`setBoundaryEvent(nodeTypes.boundaryMessageEvent, taskPosition);`

PR #949, opened 2019-12-12
Add tooltips to disabled Boundary Events

### LINKED ISSUE #935: Add tooltips to disabled boundary event buttons
Buttons in the boundary event dropdown will become disabled if they are invalid, as part of https://github.com/ProcessMaker/modeler/pull/932. Add a tooltip to these disabled buttons to let the user know why they are disabled.

Question:
 * What wording should be used for the tooltips? One is needed to tell the user there are no more empty ports available, and the other to let the user know an error event already exists on the target.

lisa-fehr:
Closes #935 
Positioned tooltip to the right as it can make clicking other Boundary Events difficult.

Proposed wording:
Screen Shot 2019-12-12 at 1 43 51 PM
Screen Shot 2019-12-12 at 1 42 39 PM

Warning variant (orange background) isn't quite like text-warning (orange text), but it may be easier to read.


lisa-fehr:
Add warning style to the text

src/components/crownBoundaryEventDropdown.vue
```diff
@@ -77,6 +81,16 @@ export default {
...
     },
+    tooltip(nodeType) {
+      if (getEmptyBoundaryEventPositionsForShape(this.shape).length === 0) {
+        return 'No available space';
+      }
+      if (!canAddBoundaryEventToTarget(nodeType, this.shape)
+        && nodeType === 'processmaker-modeler-boundary-error-event') {
+        return 'Can only add one per Task';
+      }
+      return null;
```
JCook21: This feels like a polymorphic structure trying to escape. :) Could this be extracted using the 'replace conditional with polymorphism' refactoring?

lisa-fehr: I might need help with that.

src/components/crownBoundaryEventDropdown.vue
```diff
@@ -24,6 +25,9 @@
           @click="addBoundaryEvent(nodeType)"
```
chipit24: I wonder if it would be cleaner to add the invalid text to `dropdownData`, and re-use the logic for the `:disabled` binding to display this text. This would eliminate the need for the added conditional logic in this component.

So the only addition to this file would be changinig the `b-tooltip` component to:
```html
<b-tooltip
  v-if="!canAddBoundaryEventToTarget(dropdownData.nodeType, shape)"
  :target="nodeType"
  variant="warning"
  placement="right"
>
  {{ dropdownData.disabledLabel }}
</b-tooltip>
```

**Note:** `dropdownData` is destructured in the `v-for` on the parent `li` but I referenced it directly in my example above.

This would work well with the changes just merged in https://github.com/ProcessMaker/modeler/pull/953, as the text would only have to be added to the objects in `boundaryEventDropdownData` in `task.vue`.

JCook21: @chipit24 I think this is a great solution and is better than reaching for polymorphism right now. :)

lisa-fehr: Only issue is, I was trying not to display the "Can only add one per Task" for Error Message, when the real warning is "No available space"

lisa-fehr: I pushed up changes based on this recommendation. Hopefully they show up. Github is acting weird.

src/components/crownBoundaryEventDropdown.vue
```diff
@@ -53,10 +53,12 @@ export default {
     return {
       dropdownOpen: false,
       boundaryEventIcon,
+      noAvailableSpaceLabel: 'No available space',
```
chipit24: Could this be inlined instead?

lisa-fehr: It made it trickier to read. But I can inline it.

PR #954, opened 2019-12-12
Extract components from Modeler.vue

JCook21:
This PR attempts to extract components from Modeler.vue to help reduce the size of that component.


src/components/MiniMap.vue
```diff
@@ -1,22 +1,39 @@
...
+    data-test="mini-map-btn"
+    @click="isOpen = !isOpen"
+    v-b-tooltip.hover
+    :title="isOpen ? $t('Hide Mini-Map') : $t('Show Mini-Map')"
+  >
+    <font-awesome-icon :icon="miniMapOpen ? minusIcon : mapIcon" />
+    <div class="mini-paper-container" @click="movePaper" :class="{ 'opened' : isOpen,  'closed': !isOpen, 'expanded' : expanded }">
+      <div ref="miniPaper" class="mini-paper" />
+    </div>
+  </b-button>
```
JCook21: These changes combine the mini-map and the buttons/icons that control it. This works as the mini-map is absolutely positioned.

src/components/inspectors/InspectorPanel.vue
```diff
@@ -57,14 +70,6 @@ export default {
...
       return store.getters.highlightedNode;
     },
-    inspectorStyles() {
-      this.$nextTick(() => {
-        const inspectorHeader = this.$refs.formRenderer.$children[0].$el;
-        inspectorHeader.classList.add('card-header', 'text-sm');
-      });
-
-      return 'card-body p-0';
-    },
```
JCook21: This method was unused so I deleted it.

src/components/MiniMap.vue
```diff
@@ -1,22 +1,39 @@
...
-  </div>
+  <b-button
+    class="btn btn-sm btn-secondary mini-map-btn ml-auto"
+    data-test="mini-map-btn"
+    @click="isOpen = !isOpen"
+    v-b-tooltip.hover
+    :title="isOpen ? $t('Hide Mini-Map') : $t('Show Mini-Map')"
+  >
+    <font-awesome-icon :icon="miniMapOpen ? minusIcon : mapIcon" />
+    <div class="mini-paper-container" @click="movePaper" :class="{ 'opened' : isOpen,  'closed': !isOpen, 'expanded' : expanded }">
```
lisa-fehr: Should the mini-map close every time you click on it? What if someone is navigating a large diagram?

JCook21: Thanks for noticing, that was due to something I tried due to a styling issue. I found another solution instead and have fixed that.

src/components/MiniMap.vue
```diff
@@ -1,22 +1,45 @@
...
-    <div ref="miniPaper" class="mini-paper" />
+  <div class="mini-map-btn">
+    <b-button
+      class="btn btn-sm ml-auto"
+      data-test="mini-map-btn"
+      @click="isOpen = !isOpen"
+      v-b-tooltip.hover
+      :title="isOpen ? $t('Hide Mini-Map') : $t('Show Mini-Map')"
+    >
+      <font-awesome-icon :icon="miniMapOpen ? minusIcon : mapIcon" />
```
lisa-fehr: I'm seeing an error that miniMapOpen is not defined.

JCook21: Thanks, looks like I missed that reference in the refactor.

src/components/MiniMap.vue
```diff
@@ -90,4 +112,11 @@ $transition: 0.3s;
     transition-timing-function: ease-in, ease-in, linear;
   }
 }
+
+.mini-map-btn {
+  .btn-sm {
+    border-bottom-left-radius: 0;
+    border-top-left-radius: 0;
+  }
+}
```
chipit24: Since styles are only applied to one element, the scss nesting here isn't necessary and the selector could be `.mini-map-btn .btn-sm {`.

chipit24:
Since the mini-map is positioned relative to it's parent, and because clicking the mini-map button slightly changes the width of the button, it causes a very slight jump when opening and closing the mini-map:

mini-map-jump

src/components/Modeler.vue
```diff
@@ -7,127 +7,56 @@
...
-      </div>
+      <tool-bar
+        :canvas-drag-position="canvasDragPosition"
+        :cursor="cursor"
+        :is-rendering="isRendering"
+        :panels-changed="panelsChanged"
+        :paper-manager="paperManager"
+        @load-xml="loadXML"
+        @panels-changed="panelsChanged"
+        @mini-map-changed="miniMapState"
```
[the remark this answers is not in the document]
JCook21: Yes, that's better. I didn't know Vue supported that; thanks!

PR #957, opened 2019-12-13
Refactor Css

tuantruong625:
The purpose of this PR is to implement a scalable and maintainable css architecture (ITCSS).

### Settings

Used with preprocessors and contain font, colors definitions, etc.

---

### Tools

Globally used mixins and functions. It’s important not to output any CSS in the first 2 layers.

---

### Generic

Reset and/or normalize styles, box-sizing definition, etc. This is the first layer that generates actual CSS.

---

### Elements

Styling for bare HTML elements (like H1, A, etc.). These come with default styling from the browser so we can redefine them here.

---

### Objects

Class-based selectors which define undecorated design patterns, for example media object known from OOCSS

---
Reference: https://www.xfive.co/blog/itcss-scalable-maintainable-css-architecture/

Fixes #957


src/assets/scss/settings/_settings.variables.scss
```diff
@@ -0,0 +1,7 @@
+$cursors: default, not-allowed, wait;
+$inspector-column-max-width: 265px;
+$controls-column-max-width: 265px;
+$controls-column-compressed-max-width: 95px;
+$toolbar-height: 2rem;
+$vertex-error-color: #ED4757;
+$controls-transition: 0.3s;
```
JCook21: In principle I think this is good but I think there are a few issues in practice. A lot of these variables are only used in a single component and I think there's value in those cases in having them declared close to where they're used. I think the controls and inspector variables should be moved into the components that refer to them, and this should be easier with the refactoring that I've just merged.

I think you could go further with the `$controls-transition` though. The transition duration is being used globally and is currently duplicated. You could rename this variable `$transition-duration` and refactor to use it everywhere.

src/components/inspectors/InspectorPanel.vue
```diff
@@ -133,7 +133,7 @@ export default {
       }
 
       if (this.hasCustomInspectorHandler) {
-        return this.customInspectorHandler;
+        return this.customInspectorHanEdler;
```
JCook21: Typo in this line. :)

JCook21:
@tuantruong625 changes look good. What do you feel that the advantages of removing all of the SCSS from Vue components is though? I'm honestly curious to know. :)

src/assets/scss/index.scss
```diff
@@ -0,0 +1,24 @@
...
+
+// Components – specific UI components. This is where the majority of our work takes place and our UI components are often composed of Objects and Components
+@import './components/components.modeler';
+@import './components/components.toolbar';
+@import './components/components.inspector';
+@import './components/components.controls';
+@import './components/components.sequenceFlowFormSelect';
+@import './components/components.statusBar';
+@import './components/components.miniPaper';
+@import './components/components.validationStatus';
```
JCook21: Is there any way of importing all files matching a name pattern or in a directory? The risk here is that people will forget to add new SCSS files to this list which could be a pain point in the future.

tuantruong625:
> @tuantruong625 changes look good. What do you feel that the advantages of removing all of the SCSS from Vue components is though? I'm honestly curious to know. :)

It makes the cascade of the scss more explicit. In `index.scss`, the components use any variables, mixins, functions etc. defined at the top of the files.

tuantruong625:
> I like the idea of breaking out the common CSS, like settings and tools, but disagree with the rest. With these changes, we lose the added benefit of being able to specify attributes like `lang` and `scoped` on the style tag in our Vue single file components.
> 
> The styles for components live in a completely different directly, and now when working on a component, we'd have to open two files from two different nested directory structures to get a complete picture of the component. I'm a fan of grouping things by feature, see: https://vuejs.org/v2/guide/single-file-components.html#What-About-Separation-of-Concerns.
> 
> I think if we want to break out the styles, we should keep them close to where they're used, like in the example in the docs I linked above:
> 
> ```
> 
> <template>
>   <div>This will be pre-compiled</div>
> </template>
> <script src="./my-component.js"></script>
> <style src="./my-component.css"></style>
> ```
> 
> @tuantruong625 @JCook21

Yeah agreed, especially keeping it scope will prevent any conflicts with the styles in the processmaker app.

src/components/crown/crownBoundaryEventDropdown.vue
```diff
@@ -94,3 +94,73 @@ export default {
...
+        font-size: 0.85rem;
+
+        &:disabled {
+          opacity: 0.6;
+          cursor: not-allowed;
+        }
+      }
+    }
+  }
+</style>
```
chipit24: Why didn't you break these styles into a separate file?

src/components/modeler/modeler.scss
```diff
@@ -1,3 +1,5 @@
+@import '../../../node_modules/jointjs/dist/joint.min.css';
```
chipit24: I think in scss you can use ~ to refer to node modules:
```scss
@import '~jointjs/dist/joint.min.css';
```

See https://github.com/webpack-contrib/sass-loader#resolving-import-at-rules.

src/components/controls/controls.vue
```diff
@@ -118,70 +118,4 @@ export default {
...
-    color: white;
-  }
-}
-
-.tool-icon {
-  width: 1.5rem;
-  pointer-events: none;
-}
-</style>
+<style lang="scss" src="./controls.scss" scoped />
```
chipit24: The previous styles weren't scoped so I would make sure this doesn't remove styling from other areas.

PR #959, opened 2019-12-16
Small cleanup

JCook21:
- Remove mistakenly committed file
- Remove unused code
- Move method to component it's related to and simplify code


src/components/controls.vue
```diff
@@ -158,7 +165,7 @@ $controls-transition: 0.3s;
   color: white;
   position: absolute;
   z-index: 10;
-  box-shadow: 5px 5px 8px 0px #0000004a;
+  box-shadow: 5px 5px 8px 0 #0000004a;
```
JCook21: This change is because PHPStorm pointed out the px is unneeded with a 0 measurement.

src/components/controls.vue
```diff
@@ -60,9 +59,17 @@ export default {
     };
   },
   computed: {
-    controlItems() {
-      return flatten(Object.values(this.controls))
-        .filter(control => control.label.toLowerCase().includes(this.filterQuery.toLowerCase()))
+    controls() {
+      return this.nodeTypes
+        .filter(nodeType => nodeType.control)
+        .filter(nodeType => nodeType.label.toLowerCase().includes(this.filterQuery.toLowerCase()))
```
JCook21: I'm not sure why this is needed, especially as the filterQuery is an empty string. I tried removing this check and everything was broken though...

chipit24: `filterQuery` isn't always an empty string; it's updated using the `b-form-input` in the template.

PR #961, opened 2019-12-16
Refactor Crown

JCook21:
- Refactored crown buttons to better encapsulate logic associated with clicking each button.
- Moved functionality associated with deleting pools into delete button and removed unneeded pool crown component.


src/components/crownBoundaryEventDropdown.vue
```diff
@@ -59,11 +55,18 @@ export default {
       boundaryEventIcon,
     };
   },
+  computed: {},
```
tuantruong625: This can be removed.

src/components/crownConfig.vue
```diff
@@ -89,7 +91,7 @@ export default {
     },
     boundaryEventDropdownData: {
       type: Array,
-      default: null,
+      default: () => [],
```
JCook21: I changed this for consistency since the cog dropdown data is defined in this way.

PR #962, opened 2019-12-17
Message flows stacking

### LINKED ISSUE #958: Lanes Cover Message Flows
image

tuantruong625:
The purpose of this PR is to fix messages flows that are hidden when interacting with pools/lanes.

How to test:
- Connect a message flow from two different pools
- Click on both pools
- Add lanes to either pool

**Expected**
- Messages flows are visible on all actions

Fixes #958


src/components/nodes/pool/pool.vue
```diff
@@ -174,7 +174,7 @@ export default {
       lanes.push(this.pushNewLane());
 
       await Promise.all(lanes);
-      this.$emit('set-shape-stacking', this.shape);
+      this.$emit('set-shape-stacking', this.shape, this.graph);
```
chipit24: The use of `this.graph` is also acting like a flag because of the `if` statement in `bringFlowsToFront`. Do you think it would be cleaner to leave the `ensureShapeIsNotCovered` function the same, and loop through sequence flows here? Something like:

```js
this.$emit('set-shape-stacking', this.shape);
this.graph.getLinks().forEach(link => {
  this.$emit('set-shape-stacking', link);
});
```

src/components/shapeStackUtils.js
```diff
@@ -44,6 +46,12 @@ function bringShapeToFront(shape) {
   shape.toFront({ deep: true });
 }
 
+function bringFlowsToFront(graph) {
+  if (graph) {
```
JCook21: Could it ever be possible that graph is falsey here? If not let's remove the if check.

src/components/shapeStackUtils.js
```diff
@@ -1,15 +1,17 @@
...
   if (isPool(shape)) {
     bringPoolToFront(shape);
+    bringFlowsToFront(graph);
   }
 
   const parentPool = getElementPool(shape);
 
   if (parentPool) {
     bringPoolToFront(parentPool);
+    bringFlowsToFront(graph);
```
JCook21: Instead of the duplication here how would you feel above recursively calling this method to achieve this?

src/components/nodes/pool/pool.vue
```diff
@@ -173,6 +173,9 @@ export default {
 
       await Promise.all(lanes);
       this.$emit('set-shape-stacking', this.shape);
+      this.graph.getLinks().forEach(link => {
+        this.$emit('set-shape-stacking', link);
+      });
```
chipit24: Based on my previous comment, and our discussion, since this has been added here, the `bringFlowsToFront` function and the `graph` parameter of `ensureShapeIsNotCovered` should be removed. This address @JCook21's comments as well.

PR #980, opened 2019-12-19
Fix start timer end date bug

### LINKED ISSUE #979: Start Timer Event end date reset
## Description
The end date for the start timer event is reset to today's date when unselected and re-selected.

## Reproduction steps
1. Drag a Start Event onto the grid and select Start Timer Event from the element select dropdown that opens to render a start timer event.
2. Open the Timing Control inspector accordion, select a day to repeat on, select the "On" radio button under "Ends", and choose a date other than today's date (the default).
3. De-select the start timer event (click on an empty space on the grid), then re-select the start timer event and open the Timing Control accordion again.
4. Notice that the date you set in step 2 has been reset to the default date.

## Expected behaviour
The end date should not reset to the default date.

## Actual behaviour
The end date resets to the default date.

## Software used
Chrome 79 on macOS Catalina, running latest modeler develop branch.

chipit24:
Fixes #979.

### DIFF
--- a/src/components/inspectors/TimerExpression.vue
+++ b/src/components/inspectors/TimerExpression.vue
@@ -212,7 +212,7 @@ export default {
       this.update();
     },
   },
-  mounted() {
+  created() {
     this.parseTimerConfig(this.value);
   },
   methods: {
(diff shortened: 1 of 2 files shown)


src/components/inspectors/TimerExpression.vue
```diff
@@ -212,7 +212,7 @@ export default {
       this.update();
     },
   },
-  mounted() {
```
chipit24: The reason this works is because `created` is run before this component mounts, and hence before the date picker child component mounts. What was happening before is the date picker component would mount, then emit an event right away to set the default date, since at that point `parseTimerConfig` hasn't run and so the date picker did not yet receive a value. This would overwrite the end date set during the parsing of the timer config.

Running `parseTimerConfig` before the date picker mounts ensures that the end date is set, and so the date picker would mount with an actual value and not emit its default value.

PR #982, opened 2019-12-19
Fix animation styles

chipit24:


src/components/modeler/modeler.scss
```diff
@@ -1,5 +1,6 @@
 @import '~jointjs/dist/joint.min.css';
 @import '../../assets/scss/utilities/utilities.cursor';
+@import '../../assets/scss/tools/tools.animations';
```
chipit24: The `@` alias doesn't seem to work in the path here. I wonder if there's some way to make it work 🤔

chipit24: Seems it is: https://github.com/vuejs-templates/webpack/issues/860. I'll give it a try.

PR #984, opened 2020-01-02
Require arrow functions in tests

JCook21:
This PR adds an eslint rule to require arrow functions in tests where possible and it includes changes to tests to satisfy this rule.

### DIFF
--- a/tests/e2e/.eslintrc.js
+++ b/tests/e2e/.eslintrc.js
@@ -22,6 +22,7 @@ module.exports = {
     'object-shorthand': 'error',
     'space-before-function-paren': ['error', 'never'],
     'keyword-spacing': 'error',
+    'prefer-arrow-callback': 'error',
   },
   parserOptions: {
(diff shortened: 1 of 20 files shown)


tests/e2e/.eslintrc.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
     'object-shorthand': 'error',
     'space-before-function-paren': ['error', 'never'],
     'keyword-spacing': 'error',
+    'prefer-arrow-callback': 'error',
```
chipit24: During standup you mentioned this would be a warning. Did you mean to make this an error? I'm good with keeping it as an error.

JCook21: @chipit24 I think I said the wrong thing in standup. Are you ok with this being an error? IMO it makes it more likely that this rule will definitely be observed.

chipit24: Yup, I prefer error.

PR #988, opened 2020-01-02
Change mixin to function

### LINKED ISSUE #985: Extract intermediateMailIcon mixin into function
A mixin, `src/mixins/intermediateMailIcon.vue`, defines properties of the shape on the Vue component. To improve testing and readability, this mixin can instead be turned into a function.

The problem with mixins is that they introduce side effects that are hard to work with. This mixin specifically modifies the component state (changes properties on `this`), but it's not obvious from the component. And if more mixins are added in the future, the state they modify and introduce may interfere each other, and tracking down the mixin which introduced the problem could be difficult.

The other issue with this mixin is that in one of the places where it's included, `intermediateMessageCatchEvent.vue`, the properties that are added by the mixin are modified in the component. These properties of the mixin should instead be arguments to a function.

chipit24:
Fixes #985.


chipit24:
This is the test that is failing on CI:
> 1) Common behaviour test for boundary event type Boundary Error Event retains outgoing sequence flows on Boundary Events:
     TypeError: Cannot read property 'attr' of null

This test is passing for me locally, and all other boundary event tests are passing both on CI and locally, so I'm not sure what's going on. I'm going to re-run the workflow on CI.

PR #989, opened 2020-01-02
Remove filter from controls

### LINKED ISSUE #987: Remove controls filter
Now that the new crown UI has been implemented the controls filter no longer works as it did before. This is because the selection of specific controls happens in the crown dropdown now. The easiest way to deal with this will be to remove the filter, unless @velkymx has any other suggestions or ideas for this.

JCook21:
Closes #987 

This PR removes the filter options from the controls panel. It also adds some classes to elements in the app as I was seeing some extra rounding of elements and double borders being displayed in some places where there weren't any before. If anyone knows why these have started to appear I'd be interested to know.


lisa-fehr:
@JCook21 I have the rounded corners introduction pinned down to this pr, maybe you can confirm: https://github.com/ProcessMaker/modeler/pull/954/files

JCook21:
> @JCook21 I have the rounded corners introduction pinned down to this pr, maybe you can confirm: https://github.com/ProcessMaker/modeler/pull/954/files

That's weird given that there's nothing obvious in that PR that would cause this. I'm not too worried about diagnosing this further since we have the fix for it.

PR #990, opened 2020-01-02
Open one crown dropdown at a time to prevent overlap

### LINKED ISSUE #986: Prevent overlapping of crown dropdowns
Screen Shot 2020-01-02 at 10 51 30 AM

Close task dropdown when opening boundary event dropdown and vice versa.

lisa-fehr:
Closes #986


src/components/crown/crownConfig/crownConfig.vue
```diff
@@ -134,6 +138,12 @@ export default {
     },
   },
   methods: {
+    closeBoundaryEventDropDown() {
+      this.$refs.boundaryEventDropdown.dropdownOpen = false;
+    },
+    closeTaskDropDown() {
+      this.$refs.taskDropdown.dropdownOpen = false;
```
chipit24: This is my opinion, but I want to discourage the use of `$refs` as much as possible. They are an "escape hatch" that should really only be used for accessing native DOM attributes.

I consider it an escape hatch because a component has a well-defined API: props and events. When using a ref, this API is sidestepped and it'll make refactoring and reasoning about the component more difficult.

`dropdownOpen` is part of the component's internal state, and should really only be modified through props. If we want to refactor either of the dropdown components, we now have to consider all the places where the component might be used and determine if a ref is relying on any of the components internal details.

My suggestion is to use props to share the dropdown open status between the two components. Let me know what you think.

lisa-fehr: Makes sense. If the data prop was renamed or removed, you can't directly see the association.

src/components/crown/crownConfig/crownConfig.vue
```diff
@@ -30,6 +30,8 @@
       :dropdown-data="dropdownData"
       :dropdown-initially-open="dropdownInitiallyOpen"
       :node="node"
+      :isOpen="taskDropDownWasClicked"
+      @click="() => { this.taskDropDownWasClicked = true; this.boundaryEventDropDownWasClicked = false; }"
```
JCook21: I would prefer if the component emitted an event that was caught here to be processed. As it is I feel that the crown config has too much coupling to the dropdown component by knowing that interactions with it happen through clicks.

Also, how would you feel about renaming the variables? I feel that including `clicked` in the variable name is a little too specific, particularly when the dropdown is initially open without it having been clicked. How about `taskDropDownOpen` and `boundaryEventDropDownOpen`?

src/components/crown/crownConfig/crownConfig.vue
```diff
@@ -39,6 +41,8 @@
       :moddle="moddle"
       :node="node"
       :shape="shape"
+      :isOpen="boundaryEventDropDownWasClicked"
+      @click="() => { this.taskDropDownWasClicked = false; this.boundaryEventDropDownWasClicked = true; }"
```
JCook21: Same comment as above.

src/components/crown/crownConfig/crownConfig.vue
lisa-fehr: It is getting a little tricky. There is a test that expects the Boundary Event dropdown to hide after adding a Boundary Event, but it is staying open when reselecting the task.

chipit24: This is an interesting problem. Generally, when two sibling components need to share some state, I pull that state up into some common parent component and pass it down as props (and then the sibling components no longer need to hold that state as internal data at all).

However, we don't want to bloat the crown config component with state that really only belongs to these two dropdowns. How would you feel about creating another component that wraps these two to allow sharing this state?

lisa-fehr: Moving them to a shared component works great, except for the logic where the Task Dropdown is open initially. The crownConfig needs to retain that information.

PR #993, opened 2020-01-03
Boundary message event icon colour

### LINKED ISSUE #929: Change envelope style on Boundary Message Event
Envelope became Yellow. Convert the boundary event envelope back to black.
Screen Shot 2019-12-09 at 4 58 46 PM

tuantruong625:
The purpose of this PR to change the colour of the mail icon to black.

Screen Shot 2020-01-03 at 9 42 25 AM

Fixes #929

### DIFF
--- a/src/components/nodes/boundaryMessageEvent/boundaryMessageEvent.vue
+++ b/src/components/nodes/boundaryMessageEvent/boundaryMessageEvent.vue
@@ -12,7 +12,7 @@ export default {
     },
   },
   mounted() {
-    const shapeAttributes = getIntermediateMailIconShapeAttributes('#000', 1, '#fff', boundaryMessageMailSymbol);
+    const shapeAttributes = getIntermediateMailIconShapeAttributes(boundaryMessageMailSymbol);
     this.shape.attr(shapeAttributes);
   },
 };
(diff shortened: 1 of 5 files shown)


src/mixins/intermediateMailIcon.js
```diff
@@ -1,9 +1,7 @@
-import intermediateMailSymbol from '@/assets/intermediate-mail.svg';
-
-export default function getIntermediateMailIconShapeAttributes(stroke = '#000', strokeWidth = 1, fill = '#fff') {
+export default function getIntermediateMailIconShapeAttributes(stroke = '#000', strokeWidth = 1, fill = '#fff', icon) {
```
JCook21: I think this is problematic. You've added a new argument that doesn't have a default value after arguments that do. This forces all callers of this function to supply arguments for all values. You should either move the `icon` argument to be first (PHP/WebStorm will do this refactoring for you) or remove the default values from this function.

src/components/nodes/boundaryMessageEvent/boundaryMessageEvent.vue
```diff
@@ -11,7 +12,7 @@ export default {
     },
   },
   mounted() {
-    const shapeAttributes = getIntermediateMailIconShapeAttributes();
+    const shapeAttributes = getIntermediateMailIconShapeAttributes('#000', 1, '#fff', boundaryMessageMailSymbol);
```
JCook21: Because you've added an argument that requires a value after ones that have defaults this call is now required to pass all arguments.

PR #996, opened 2020-01-06
Disable user select in validation footer

tuantruong625:
The purpose of this PR is to prevent selecting the `Auto Validation` text in the validation footer.


tuantruong625:
> Looks good! Can you explain why this change is needed though?

In some cases, when you are toggling the validation button the `Auto Validation` text becomes highlighted.

PR #1002, opened 2020-01-07
Fix multiple accordions open bug

### LINKED ISSUE #999: Multiple accordions open for task component
## Description
Only a single accordion should be open at one time in the inspector config, but it seems possible to have more than one open for task components.

## Reproduction steps
* Drag a Task component onto the grid, and select the base Task type.
* Expand the "Assignment Rules" accordion in the inspector panel.

## Expected behaviour
When the "Assignment Rules" accordion opens, the "Configuration" accordion closes.

## Actual behaviour
Both the Configuration and the Assignment Rules accordions remain open.

## Software used
Latest modeler develop branch on Chrome 79 running on macOS Catalina.

tuantruong625:
The purpose of this PR is to fix the bug where multiple accordions would be open. The expected behavior for the accordions is that only one can be open at a time.

**Note**
Changes are also required in processmaker-core repo.
https://github.com/ProcessMaker/processmaker/pull/2712

Fixes #999

### DIFF
--- a/src/setup/extensions/testTaskInspectorExtension.js
+++ b/src/setup/extensions/testTaskInspectorExtension.js
@@ -39,7 +39,7 @@ window.ProcessMaker.EventBus.$on('modeler-init', ({ registerInspectorExtension }
       initiallyOpen: false,
       label: 'Assignment Rules',
       icon: 'users',
-      name: 'assignmentRules',
+      name: 'inspector-accordion',
     },
     items: [
       {


tuantruong625:
Having the behavior of one accordion open at a time requires them to be part of a group. 
https://bootstrap-vue.js.org/docs/components/collapse/

In `FormAccordion` component, accordions are binding together by `accordion-${config.name}`. We use the config name to group accordion together. If accordions did not have matching `config.name` they will not work together like an accordion group.

Issue #997, opened 2020-01-06
Change Task Warning


velkymx:
After placing a configured task, and then trying to change it to another type of task, show a warning:

"Changing this task type will replace your current configuration"

[Cancel] [Confirm]

chipit24:
Acceptance criteria:
 * A confirmation modal should open when the user attempts to change the element type:
   * This modal should only be displayed after the element select dropdown has been closed at least once.
   * Clicking confirm should switch to the new element, and clicking cancel should close the modal and retain the same element.

PR #1003, opened 2020-01-07
Added confirmation modal to switch type

lisa-fehr:
Related #997 

Covers all type switching. Not just on Task.

It will not ask to confirm when type dropdown is initially open (When an element is first added to the diagram). 

It does not ask to confirm switching type after initially uploading a xml. Needs alternative solution to set taskDropdownInitiallyOpen to false after uploading xml.

### DIFF
--- a/src/components/crown/crownButtons/crownTaskDropdown.vue
+++ b/src/components/crown/crownButtons/crownTaskDropdown.vue
@@ -1,6 +1,7 @@
 <template>
   <div class="cog-container" v-if="dropdownData.length > 0" role="menuitem">
     <crown-button
+      data-test="select-type-dropdown"
       id="dropdown-button"
       aria-label="Select a type"
       v-on="$listeners"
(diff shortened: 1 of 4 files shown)


src/components/crown/crownButtons/crownTaskDropdown.vue
```diff
@@ -42,7 +43,7 @@ export default {
         this.$emit('toggle-dropdown-state', false);
         return;
       }
-      this.$emit('replace-node', data);
+      this.$emit('confirm-replace', data);
```
JCook21: I think the event name should still be 'replace-node'. The fact that we're confirming the replacement is an implementation detail, the action is still replacing the node.

lisa-fehr: I can't seem to get it to work that way. It seems to emit up the `replace-node` regardless of my interception and destroys the node before the confirmReplace method ever fires.

lisa-fehr: I could put the confirmReplace on Modeler.vue

JCook21: OK, that's fine. My main problem is more the name of the event: `confirm-replace` seems to me to be too closely tied to the implementation. Can you think of something that's more generic. describing the intent rather than the specifics?

PR #1004, opened 2020-01-07
Remove rule disallowing multiple blanks start events

JCook21:
This has been requested by @velkymx in Slack.


JCook21:
> Should the package upgrades be in another branch or are they required for this change?

That must have just come through an NPM install... I don't think there's any harm in leaving them in though. The lock file seems to have been updated for a patch release.

PR #1008, opened 2020-01-08
Remove flows when switching type

### LINKED ISSUE #1007: Remove attached flows when switching type
# Bug report issue
## Description
Flows remain attached to previous element after switching an element to a different type.
## Reproduction steps
- Attach Sequence Flows to a Task
- Switch the Task to another type
- Download the xml and see that the flows remain
## Expected behaviour
When the previous element is replaced, the flows are removed.
## Actual behaviour
Flows remain attached to the deleted/replaced element.

lisa-fehr:
Closes #1007

### DIFF
--- a/src/components/crown/utils.js
+++ b/src/components/crown/utils.js
@@ -1,10 +1,11 @@
 export function removeFlows(graph, shape) {
-  graph.getConnectedLinks(this.shape).forEach(shape => this.$emit('remove-node', shape.component.node));
-  shape.getEmbeddedCells({ deep: true }).forEach(cell => {
-    if (cell.component) {
-      graph.getConnectedLinks(cell).forEach(shape => this.$emit('remove-node', shape.component.node));
+  graph.getConnectedLinks(shape).forEach(shape => this.$emit('remove-node', shape.component.node));
+  shape.getEmbeddedCells({ deep: true })
+    .filter(cell => cell.component)
+    .flatMap(cell => graph.getConnectedLinks(cell))
+    .forEach(cell => {
+      this.$emit('remove-node', shape.component.node);
       shape.unembed(cell);
       this.$emit('remove-node', cell.component.node);
-    }
-  });
+    });
 }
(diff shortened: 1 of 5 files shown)


src/components/crown/utils.js
```diff
@@ -0,0 +1,10 @@
+export function removeFlows(graph, shape) {
+  graph.getConnectedLinks(this.shape).forEach(shape => this.$emit('remove-node', shape.component.node));
+  shape.getEmbeddedCells({ deep: true }).forEach(cell => {
+    if (cell.component) {
+      graph.getConnectedLinks(cell).forEach(shape => this.$emit('remove-node', shape.component.node));
+      shape.unembed(cell);
+      this.$emit('remove-node', cell.component.node);
+    }
+  });
+}
```
JCook21: I think this could probably be rewritten as a single pipeline, which would remove some of the nesting and make this more readable IMO. Something like this:
`shape.getEmbeddedCells({ deep: true })
    .filter(cell => cell.component)
    .flatMap(cell => graph.getConnectedLinks(cell))
    .foreach(shape => {
        this.$emit('remove-node', shape.component.node));
        shape.unembed(cell);
        this.$emit('remove-node', cell.component.node);
    });
`

src/components/crown/utils.js
```diff
@@ -0,0 +1,10 @@
+export function removeFlows(graph, shape) {
+  graph.getConnectedLinks(this.shape).forEach(shape => this.$emit('remove-node', shape.component.node));
+  shape.getEmbeddedCells({ deep: true }).forEach(cell => {
+    if (cell.component) {
+      graph.getConnectedLinks(cell).forEach(shape => this.$emit('remove-node', shape.component.node));
+      shape.unembed(cell);
+      this.$emit('remove-node', cell.component.node);
+    }
+  });
```
JCook21: `I think this could probably be rewritten as a single pipeline, which would remove some of the nesting and make this more readable IMO. Something like this:
`shape.getEmbeddedCells({ deep: true })
    .filter(cell => cell.component)
    .flatMap(cell => graph.getConnectedLinks(cell))
    .foreach(shape => {
        this.$emit('remove-node', shape.component.node));
        shape.unembed(cell);
        this.$emit('remove-node', cell.component.node);
    });
`

JCook21: I just realized that this is just a refactoring in moving this code to this place. I still think it's a good time to make an improvement though. :)

PR #1009, opened 2020-01-08
Extract node class and responsibilities

JCook21:
This PR introduces a node class and moves code into that class. There is much more that can be done with this but I wanted to open this PR before continuing as there's a danger that there wold be just too much code to review.


src/components/nodeManager.js
```diff
@@ -1,46 +1,23 @@
...
-
-export function getTargetProcess(node, processes, processNode) {
-  return node.pool
-    ? processes.find(({id}) => id === node.pool.component.node.definition.get('processRef').id)
-    : processNode.definition;
-}
-
-export function isBoundaryEvent(node) {
-  return node.definition.$type === 'bpmn:BoundaryEvent';
+  targetProcess.get(target).push(node.definition);
```
JCook21: I really wanted to make this function part of the node class too but kept running into a really weird problem. Simply moving the code into the node class without even calling it caused a very hard to diagnose crash. I spent a number of hours on this and sought help but still couldn't resolve it. I decided to leave this for now and to move on.

src/setup/registerNodes.js
```diff
@@ -73,7 +74,6 @@ const customParserFactory = (nodeType, primaryIdentifier, secondaryIdentifier) =
 };
 
 export default function registerNodes({ registerNode, registerBpmnExtension }) {
-  registerNode(startEvent);
```
JCook21: Adding the start event first is no longer necessary since we explicitly sort the controls in the left hand controls column.

src/components/nodes/boundaryMessageEvent/boundaryMessageEvent.vue
```diff
@@ -8,7 +8,7 @@ export default {
   methods: {
     isValidBoundaryEventTarget(model) {
       const component = model.component;
-      return component && component.node.definition.$type === 'bpmn:CallActivity';
+      return component && component.isBpmnType('bpmn:CallActivity');
```
lisa-fehr: Should it be this?
```suggestion
      return component && component.node.isBpmnType('bpmn:CallActivity');
```

tests/unit/nodeManager.spec.js
```diff
@@ -123,30 +124,39 @@ describe('nodeManager', () => {
...
 
-      const targetProcess = getTargetProcess(node, processes, {});
+      const targetProcess = node.getTargetProcess(processes, {});
 
       expect(targetProcess).toBe(process);
     });
   });
-});
+})
+;
```
lisa-fehr: Minor suggestion to move this to 161

PR #1010, opened 2020-01-09
Merge breadcrumb nav and toolbar

### LINKED ISSUE #788: Move Secondary Nav Buttons
image

tuantruong625:
The purpose of this PR is to merge the breadcrumb navigation with the toolbar into a single bar.

Associated PR in processmaker repo
https://github.com/ProcessMaker/processmaker/pull/2722

Fixes #788


src/components/toolbar/ToolBar.vue
```diff
@@ -163,7 +184,10 @@ export default {
...
         .then(() => window.ProcessMaker.EventBus.$emit('modeler-change'));
     },
+    saveBpmn() {
+      this.$emit('saveBpmn');
+    },
   },
 };
 </script>
-<style lang="scss" src="./toolbar.scss" scoped />
+<style lang="scss"  src="./toolbar.scss" />
```
lisa-fehr: Does scoped need to be removed?

src/components/modeler/Modeler.vue
```diff
@@ -1,5 +1,19 @@
 <template>
   <b-card-body class="overflow-hidden position-relative p-0 vh-100" data-test="body-container">
+    <b-row class="w-100 m-0">
```
JCook21: Could we make the row wrapper part of the tool-bar component? What do you think?

src/components/modeler/Modeler.vue
```diff
@@ -121,12 +126,7 @@ import PaperManager from '../paperManager';
 import registerInspectorExtension from '@/components/InspectorExtensionManager';
 
 import initAnchor from '@/mixins/linkManager.js';
-import {
-  addIdToNodeAndSetUpDiagramReference,
-  addNodeToProcess,
-  getTargetProcess,
-  isBoundaryEvent,
-} from '@/components/nodeManager';
+import { addIdToNodeAndSetUpDiagramReference, addNodeToProcess, getTargetProcess, isBoundaryEvent } from '@/components/nodeManager';
```
JCook21: You might want to rebase from develop: some of these have been deleted and brought into the new Node class.

src/components/modeler/Modeler.vue
```diff
@@ -220,6 +221,9 @@ export default {
     },
   },
   methods: {
+    saveBpmn() {
+      this.$emit('saveBpmn');
```
JCook21: Since this just re-emits the event would it be possible to do that inline in the handler listed in the template?

src/components/toolbar/ToolBar.vue
```diff
@@ -1,99 +1,119 @@
...
+          <font-awesome-icon :icon="miniMapOpen ? minusIcon : mapIcon" />
+        </b-button>
+      </div>
 
       <b-button
+        v-if="breadcrumbData.length > 0"
         class="btn btn-sm btn-secondary mini-map-btn ml-auto"
         data-test="mini-map-btn"
-        @click="miniMapOpen = !miniMapOpen"
+        @click="saveBpmn"
```
JCook21: Same comment as in the modeler: why bind this to a (mostly empty) method where we can emit the event inline here in the handler?

src/components/toolbar/breadcrumb/Breadcrumb.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+  <nav id="breadcrumbs">
+    <ol v-if="breadcrumbData.length > 0" class="breadcrumb">
+      <li class="breadcrumb-item">
+        <a href="/"><i class="fas fa-home" /></a>
+      </li>
+      <li class="breadcrumb-item" v-for="(breadcrumb, index) in breadcrumbData[0]" :key="index">
+        <a :href="breadcrumb.url" v-if="breadcrumb.url">{{ breadcrumb.text }}</a>
+        <span v-if="!breadcrumb.url">{{ breadcrumb.text }}</span>
```
JCook21: Could this v-if be a v-else instead? That might save the negation of the property. Not sure if that makes it any clearer though.

jest.config.js
```diff
@@ -28,5 +28,5 @@ module.exports = {
   ],
   collectCoverage: true,
   coverageDirectory: 'jest-coverage',
-  testEnvironment: 'node',
+  testEnvironment: 'jsdom',
```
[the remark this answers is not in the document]
tuantruong625: Good call.

src/components/toolbar/ToolBar.vue
```diff
@@ -1,99 +1,119 @@
...
+          @click="miniMapOpen = !miniMapOpen"
+          v-b-tooltip.hover
+          :title="miniMapOpen ? $t('Hide Mini-Map') : $t('Show Mini-Map')"
+        >
+          <font-awesome-icon :icon="miniMapOpen ? minusIcon : mapIcon" />
+        </b-button>
+      </div>
 
       <b-button
+        v-if="breadcrumbData.length > 0"
```
chipit24: I wonder if this `v-if` is necessary. The save button doesn't inherently have anything to do with breadcrumbs. The fact that the breadcrumbs exist means this is being loaded from the core app, which is where the save button would actually have an effect. So in essence, this `v-if` is a flag that's saying "if loaded in processmaker core".

Since clicking this button only emits an event, I wonder if it's safe to remove the `v-if` and always render it.

tuantruong625: It would be just a visual thing. Clicking the save button in stand-alone does nothing. Removing would be safe.

src/components/toolbar/ToolBar.vue
```diff
@@ -166,4 +188,4 @@ export default {
   },
 };
 </script>
-<style lang="scss" src="./toolbar.scss" scoped />
+<style lang="scss"  src="./toolbar.scss" />
```
lisa-fehr: Remove the space added:
```suggestion
<style lang="scss" src="./toolbar.scss" />
```

tests/e2e/specs/SequenceFlows.spec.js
```diff
@@ -52,7 +52,12 @@ describe('Sequence Flows', () => {
     cy.get('[name=conditionExpression]').should('have.value', testString);
   });
 
-  it('Allows modifying anchor points', () => {
+  /**
```
lisa-fehr: Should this comment get moved into an issue?

PR #1016, opened 2020-01-13
Fix duplicate messages and errors

### LINKED ISSUE #1015: Message element duplicated
## Description
The message element is duplicated when a process with a message is uploaded, or undo/redo is pressed.

## Reproduction steps
1. Add an Intermediate Message Throw Event onto the paper.
2. Press undo/redo or download and re-upload the XML for this process.
3. Download the XML and examine it.

## Expected behaviour
The XML contains a single `bpmn:message` element.

## Actual behaviour
The XML contains two `bpmn:message` elements (`<bpmn:message id="node_2_message" name="node_2_message" />` has been duplicated).

## Software used
Latest modeler develop branch, running on Chrome 79 on macOS Catalina.

## Other information
Screen Shot 2020-01-10 at 4 19 39 PM

The same issue also occurs wth the Message End Event and Error End Event.

chipit24:
Fixes #1015.

The issue this PR fixes didn't seem to cause any major usability issues, but since duplicate message elements were being added on reload of the BPMN process, two messages would show up in the dropdown of the associated catch event.

Here is a GIF of that issue:
duplicate_messages

### DIFF
--- a/src/components/nodes/intermediateMessageThrowEvent/intermediateMessageThrowEvent.vue
+++ b/src/components/nodes/intermediateMessageThrowEvent/intermediateMessageThrowEvent.vue
@@ -9,10 +9,25 @@ export default {
   data() {
     return {
       message: this.moddle.create('bpmn:Message', {
-        id: `${ this.id }_message`,
-        name: `${ this.id }_message`,
+        id: `${this.id}_message`,
+        name: `${this.id}_message`,
       }),
     };
+  },
+  methods: {
+    addMessageRef() {
+      if (this.node.definition.get('eventDefinitions')[0].messageRef) {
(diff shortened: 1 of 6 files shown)


JCook21:
@chipit24 the changes look good but it's hard for me to review this without a little more context. Can you explain how this was broken before as well as how your changes fix this?

src/components/nodes/intermediateMessageThrowEvent/intermediateMessageThrowEvent.vue
```diff
@@ -24,11 +39,8 @@ export default {
         x: 2,
       },
     });
-    this.rootElements.push(this.message);
```
chipit24: This was the crux of the issue. Event if we don't end up using `this.message` it was being added to the BPMN process.

Issue #1011, opened 2020-01-10
Message Flow Names Not Saving


velkymx:
https://www.dropbox.com/s/cpayv2t3uslkpvh/message-name-not-saving.mp4?dl=0

tuantruong625:
Fixed in the latest develop branch.
https://github.com/ProcessMaker/modeler/commit/b8a2df3cfd632ef5336929d06bba4a4d5bba85b9

chipit24:
🔥 **_hot_** 🔥 **_fix_** 🔥 was pushed to `develop` that should resolve this: https://github.com/ProcessMaker/modeler/commit/b8a2df3cfd632ef5336929d06bba4a4d5bba85b9.

PR #1020, opened 2020-01-13
Add test for message name

chipit24:
The PR adds a test for the following issue: https://github.com/ProcessMaker/modeler/issues/1011.

### DIFF
--- a/tests/e2e/specs/IntermediateMessageThrowEvent.spec.js
+++ b/tests/e2e/specs/IntermediateMessageThrowEvent.spec.js
@@ -120,6 +120,9 @@ describe('Intermediate Message Throw Event', () => {
     addNodeTypeToPaper(intermediateMessageThrowEventPosition, nodeTypes.intermediateCatchEvent, 'switch-to-intermediate-message-throw-event');
 
     typeIntoTextInput('[name=messageName]', messageName);
+
+    /* Something outside of the inspector panel has to be selected to trigger the focusout event;
+     * otherwise it creates a race condition with the Undo button click event. */
     cy.get('[aria-label="Toolbar"]').click();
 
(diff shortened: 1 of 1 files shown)


tests/e2e/specs/IntermediateMessageThrowEvent.spec.js
```diff
@@ -115,4 +115,20 @@ describe('Intermediate Message Throw Event', () => {
       expect(numberOfMessages).to.equal(1, 'More than 1 message element was found');
     });
   });
+
+  it('should retain message name after loading XML', () => {
+    addNodeTypeToPaper(intermediateMessageThrowEventPosition, nodeTypes.intermediateCatchEvent, 'switch-to-intermediate-message-throw-event');
+
+    typeIntoTextInput('[name=messageName]', messageName);
+    cy.get('[aria-label="Toolbar"]').click();
```
chipit24: I had to select something outside of the inspector panel to trigger the `focusout` event; otherwise it created a race condition with the Undo button click event.

JCook21: I think this should be added as a comment to the code. The risk is that no-one will remember this the next time this test is edited and this will become a recurring problem. A comment seems appropriate here.

lisa-fehr: Could this be re-useable in utils? This could be handy.

chipit24: I don't know if I want to make it a util since it's a hack. So far this is the only place I've needed this; it can be pulled out into a util when there's 1-2 other places we're having to use it, but hopefully by then we'll figure out a way to handle the weird async nature of these event handlers.

Issue #1019, opened 2020-01-13
Keyboard Navigation Bug


tuantruong625:
# Bug report issue
## Description
When a user has an input focused and uses the arrow keys to move the caret around any text. The highlighted node will move its position.

## Reproduction steps
- Drag any element onto the paper
- Focus input in the inspector,
- Use the arrow keys
## Expected behaviour
Focusing the input and navigating the input with arrow keys does not move the element.
## Actual behaviour
Focusing the input and navigating the input with arrow keys changes the position of the highlighted element.
## Software used
- Latest Develop Branch. January 13, 2020.
- MacOS Catalina
- Chrome 79.0

PR #1029, opened 2020-01-14
Fix for arrow key events incorrectly triggering element movement

johnblackspear:
Addresses [Keyboard Navigation Bug (1019)](https://github.com/ProcessMaker/modeler/issues/1019).

This checks to see whether a model element has focus before moving it when pressing arrow keys.

This PR adds an early return to the `keydown` arrow key handler originally defined in `src/components/modeler/Modeler.vue`. I've extracted the event listener logic to `src/components/modeler/moveWithArrowKeys.js`  - and we could consider moving all the event listeners out of the Modeler source to improve the readability of `mounted()`. What do you think?

I've attempted to add some tests for this but the effort of that seems to outweigh the benefit (considering the small size of the change) : Cypress has issues around triggering the `keydown` and then retargeting the now moved element; jest has issues around mocking out `moveShapeByKeypress` that means we'll need to refactor quite a few things to test this. The pragmatic solution seemed to be a manual test.

It's been manually tested in Firefox, Chrome, and Safari.

### DIFF
--- a/src/components/modeler/moveWithArrowKeys.js
+++ b/src/components/modeler/moveWithArrowKeys.js
@@ -1,4 +1,5 @@
 import PaperManager from '@/components/paperManager';
+import store from '@/store';
 
 export const moveAmount = PaperManager.gridSize / 2;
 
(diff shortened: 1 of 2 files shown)


src/components/modeler/moveWithArrowKeys.js
```diff
@@ -47,3 +48,16 @@ function expandPoolToContainElement(shape) {
 
   pool.component.updateLaneChildren();
 }
+
+export function keydownListener(event) {
+  const formElementHasFocus = event.target.toString().toLowerCase().indexOf('body') === -1;
```
chipit24: I think using [`includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) would be more readable than `indexOf` with a comparison to `-1`.

I wonder if we can also use `event.target === document.body`?

johnblackspear: Yeah that would be a lot more readable, thank you

PR #1031, opened 2020-01-15
Remove FormDatePicker dependency from vue-form-elements

### LINKED ISSUE #1030: Copy datetime picker into modeler
Copy datetime picker into modeler to reduce dependancy on `vue-form-elements`

lisa-fehr:
Closes #1030

### DIFF
--- a/src/components/dateUtils.js
+++ b/src/components/dateUtils.js
@@ -1,0 +1,32 @@
+/* global ProcessMaker*/
+import moment from 'moment-timezone';
+
+function getProcessMakerUserValue(key) {
+  if (typeof ProcessMaker !== 'undefined' && ProcessMaker.user) {
+    return ProcessMaker.user[key];
+  }
+}
+
+export function getTimezone() {
+  return getProcessMakerUserValue('timezone') || moment.tz.guess();
+}
+
+export function getLang() {
+  return getProcessMakerUserValue('lang') || 'en';
+}
+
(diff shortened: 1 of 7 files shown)


src/components/dateUtils.js
```diff
@@ -0,0 +1,32 @@
+/* global ProcessMaker*/
+import moment from 'moment-timezone';
```
chipit24: It looks like you'll have to install `moment-timezone`, I don't see it under dependencies.

src/components/inspectors/mixins/DataFormat.js
```diff
@@ -0,0 +1,112 @@
+import Validator from 'validatorjs';
```
chipit24: This file, along with `validation.js`, were made to handle all form renderer components, but since we're only dealing with the date picker in this case, I think they can be significantly reduced and possibly inlined into the datepicker component.

src/components/inspectors/FormDatePicker.vue
```diff
@@ -0,0 +1,116 @@
+<template>
+  <div class="form-group position-relative">
+    <label v-uni-for="name">{{ label }}</label>
```
chipit24: The unique ID mixin requires a matching `v-uni-id` property, which I don't see here, so I think `v-uni-for="name"` can be removed. You could wrap the `label` around `date-picker` to create the same association.

PR #1040, opened 2020-01-20
Show name and expression labels

### LINKED ISSUE #1039: Show name and expression labels on sequence flows
## Description
Show name and expression labels on sequence flows

tuantruong625:
The purpose of this PR is to show the name and expression labels on sequences flows.

Fixes #1039

### DIFF
--- a/src/components/nodes/sequenceFlow/sequenceFlow.vue
+++ b/src/components/nodes/sequenceFlow/sequenceFlow.vue
@@ -92,14 +92,14 @@ export default {
     'node.definition': {
       handler({ startEvent: startEventId, conditionExpression }) {
         const startEvent = store.getters.globalProcessEvents.find(event => event.id == startEventId);
+        const newExpressionLabel = get(conditionExpression, 'body');
+        const newNameLabel = this.shapeName || get(startEvent, 'name');
 
         if (this.targetIsCallActivity) {
           this.nameLabel = get(startEvent, 'name');
+          this.expressionLabel = newExpressionLabel;
           return;
         }
-
-        const newExpressionLabel = get(conditionExpression, 'body') || get(startEvent, 'name');
-        const newNameLabel = this.shapeName || get(startEvent, 'name');
 
         if (newNameLabel !== this.nameLabel || newExpressionLabel !== this.expressionLabel) {
           this.nameLabel = newNameLabel;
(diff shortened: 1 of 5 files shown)


src/components/nodes/sequenceFlow/sequenceFlow.vue
```diff
@@ -79,15 +94,17 @@ export default {
...
     'node.definition': {
       handler({ startEvent: startEventId, conditionExpression }) {
-        if (!this.sourceIsGateway && !this.targetIsCallActivity) {
+        if (this.targetIsCallActivity) {
           return;
         }
 
         const startEvent = store.getters.globalProcessEvents.find(event => event.id == startEventId);
-        const newLabel = get(conditionExpression, 'body') || get(startEvent, 'name');
+        const newExpressionLabel = get(conditionExpression, 'body') || get(startEvent, 'name');
```
lisa-fehr: Wonder if we can drop the `get(startEvent, 'name')`. In what scenario would this appear?

chipit24:
The labels don't appear to be in the correct positions for me. The name is not quite in the centre, and the expression is right on the line, not under it.

Screen Shot 2020-01-23 at 2 27 37 PM

Should both labels be centred between the shapes, and be above and below the flow?

src/components/nodes/sequenceFlow/sequenceFlowConfig.js
```diff
@@ -1 +1,2 @@
-export const expressionPosition = { offset: -25 };
+export const namePosition = { offset: { y: -20 } };
+export const expressionPosition =  { offset: { y: 40 } };
```
johnblackspear: I suppose some fiddling with these values could be the answer to Rob's comment?

tuantruong625: When testing with Rob, you are able to change the position of the label which caused the label to be off-centered.

src/components/nodes/sequenceFlow/sequenceFlow.vue
tuantruong625: Connecting a start-event to a sub-process will trigger this scenario.

tuantruong625:
**Note**
When connecting a start event to a subprocess the association between are set in the XML but not visually represented on the sequence flow.

chipit24:
When a gateway is connected to a sub-process, the expression doesn't show, only the name.

src/components/nodes/sequenceFlow/sequenceFlow.vue
```diff
@@ -95,9 +98,12 @@ export default {
         const newExpressionLabel = get(conditionExpression, 'body');
         const newNameLabel = this.shapeName || get(startEvent, 'name');
 
-        if (this.targetIsCallActivity) {
+        if (this.targetIsCallActivity ) {
```
chipit24:
```suggestion
        if (this.targetIsCallActivity) {
```

velkymx:
@tuantruong625 let's remove the expression and just use the label.

PR #1041, opened 2020-01-20
Remove messageRef after deleting message

### LINKED ISSUE #1014: Remove messageRef when message is deleted
If you assign the 'Listen for Message' and then delete the message, the association remains.

Example: https://www.dropbox.com/s/e6k4cqpjcfq87sn/messageRefRemains.mov?dl=0

lisa-fehr:
Closes #1014

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -85,6 +85,16 @@ export default new Vuex.Store({
         state.nodes.splice(index, 1);
       }
     },
+    removeMessageRef(state, message) {
+      state.nodes.filter(({ definition }) => (
+        definition.$type === 'bpmn:IntermediateCatchEvent' ||
+        definition.$type === 'bpmn:StartEvent'
+      )
+        && definition.eventDefinitions && definition.eventDefinitions.find(({ messageRef }) => messageRef === message))
+        .forEach(({ definition }) => {
+          definition.eventDefinitions[0].messageRef = null;
+        });
+    },
     setGraph(state, graph) {
       state.graph = graph;
     },
--- a/src/store.js
+++ b/src/store.js
@@ -90,7 +90,7 @@ export default new Vuex.Store({
         definition.$type === 'bpmn:IntermediateCatchEvent' ||
         definition.$type === 'bpmn:StartEvent'
       )
-        && definition.eventDefinitions && definition.eventDefinitions.find(({ messageRef }) => messageRef === message))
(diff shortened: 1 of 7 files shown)


tests/e2e/specs/IntermediateCatchEvent.spec.js
```diff
@@ -95,4 +97,35 @@ describe('Intermediate Catch Event', () => {
...
+    getElementAtPosition(intermediateThrowEventPosition).click().then($throwEvent => {
+      getCrownButtonForElement($throwEvent, 'delete-button').click({ force: true });
+    });
+
+    waitToRenderAllShapes();
+
+    cy.get('[data-test=downloadXMLBtn]').click();
+    cy.window().its('xml').then(removeIndentationAndLinebreaks).should(xml => {
+      expect(xml).to.not.contain(messageRef);
+    });
```
chipit24: We have a helper in `utils.js` that you could use here:
```js
assertDownloadedXmlDoesNotContainExpected(messageRef);
```

tests/e2e/specs/StartMessageEvent.spec.js
```diff
@@ -0,0 +1,42 @@
...
+    getElementAtPosition(endMessageEventPosition).click().then($throwEvent => {
+      getCrownButtonForElement($throwEvent, 'delete-button').click({ force: true });
+    });
+
+    waitToRenderAllShapes();
+
+    cy.get('[data-test=downloadXMLBtn]').click();
+    cy.window().its('xml').then(removeIndentationAndLinebreaks).should(xml => {
+      expect(xml).to.not.contain(messageRef);
+    });
```
chipit24: Same here, regarding the helper function.

src/store.js
```diff
@@ -85,6 +85,16 @@ export default new Vuex.Store({
         state.nodes.splice(index, 1);
       }
     },
+    removeMessageRef(state, message) {
+      state.nodes.filter(({ definition }) => (
+        definition.$type === 'bpmn:IntermediateCatchEvent' ||
+        definition.$type === 'bpmn:StartEvent'
+      )
+        && definition.eventDefinitions && definition.eventDefinitions.find(({ messageRef }) => messageRef === message))
```
chipit24: I think using `includes` instead of `find` here would be more "intent revealing", since we only care about checking that the message ref matches, we don't actually need to retrieve a reference to it.

lisa-fehr: `includes` didn't work on the associated array. So I tried `some`

Issue #1028, opened 2020-01-14
Process large diagram issue


lisa-fehr:
# Bug report issue
## Description
Uploading a large bpmn hangs and can use up a lot of memory.
## Reproduction steps
Upload the following bpmn: https://www.dropbox.com/s/3pk72nvqo0izkqf/SSC%20POLAND%20IT%20and%20IS%20Resources%20Management.txt?dl=0 (rename extension to xml or bpmn)
## Expected behaviour
Upload should not hang
## Actual behaviour
Can crash the browser. In a good scenario it takes 7 minutes to load. After loading, alterations to the diagram can still take a long time.

![Screen Shot 2020-01-14 at 1 51 38 PM](https://user-images.githubusercontent.com/6653340/72377998-87f76b00-36de-11ea-9e2c-187618b919e4.png)

PR #1042, opened 2020-01-21
Improve performance for large processes

chipit24:
Fixes https://github.com/ProcessMaker/modeler/issues/1028.

improved_loading

The process loads and is interactive much sooner (within 7 seconds), and an alert is now shown along with the loading cursor.

### DIFF
--- a/src/components/modeler/Modeler.vue
+++ b/src/components/modeler/Modeler.vue
@@ -665,6 +665,10 @@ export default {
       return shape.component != null;
     },
     setShapeStacking(shape) {
+      if (this.isRendering) {
+        return;
+      }
+
       this.paperManager.performAtomicAction(() => ensureShapeIsNotCovered(shape, this.graph));
     },
   },
--- a/src/components/modeler/Modeler.vue
+++ b/src/components/modeler/Modeler.vue
@@ -40,6 +40,7 @@
       </b-col>
 
       <mini-paper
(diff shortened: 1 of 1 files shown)


src/components/modeler/Modeler.vue
```diff
@@ -40,6 +40,7 @@
       </b-col>
 
       <mini-paper
+        v-if="miniMapOpen"
```
chipit24: Not rendering the mini paper (as it's closed by default on load anyway) reduced the load time by 2-3x.

src/components/modeler/Modeler.vue
```diff
@@ -665,6 +666,10 @@ export default {
       return shape.component != null;
     },
     setShapeStacking(shape) {
+      if (this.isRendering) {
+        return;
+      }
+
```
chipit24: Multiple calls to `toFront` (which is what `ensureShapeIsNotCovered` does) seemed to be called many times when loading the large process, and appeared to create a significant amount of CPU/rendering work.

While this didn't seem to cause any overlapping issues on initial load, it is still required to prevent overlapping after the process has loaded.

lisa-fehr:
Oops. Just noticed that the mini-map looks odd @chipit24 
Screen Shot 2020-01-22 at 2 10 49 PM

PR #1047, opened 2020-01-22
Allow closer link connections on shapes

### LINKED ISSUE #1037: Closer Element Support
Allow closer items threshold for sequence flows before they try to redraw/reroute themselves.

chipit24:
Fixes #1037.

Looks like padding was added as an option to the orthogonal router in JointJS v2.2.0. Setting this to a small value (1 in this case) allows shapes to get much closer together without looping. Setting it to 0 seems to have the same result as not setting it at all (defaults to 20). Reference: https://resources.jointjs.com/docs/jointjs/v3.1/joint.html#routers.orthogonal.

Screen Shot 2020-01-22 at 10 56 36 AM

### DIFF
--- a/src/components/nodes/messageFlow/messageFlow.vue
+++ b/src/components/nodes/messageFlow/messageFlow.vue
@@ -60,7 +60,7 @@ export default {
   },
   methods: {
     updateRouter() {
-      this.shape.router('orthogonal');
+      this.shape.router('orthogonal', { padding: 1 });
     },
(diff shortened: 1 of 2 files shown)


chipit24:
Yeah, that does look odd, but even in the "before" screen the flow should be adjusted to enter from the side, not go up and then down into the corner. Unless that's how it got connected by default without adjusting vertices or the anchor point.

PR #1050, opened 2020-01-22
Fix start timer loop

### LINKED ISSUE #1044: Message Start Event Crash
When you drag a new start event on to the canvas and choose the Message Start, the whole browser freezes up.

image

chipit24:
Fixes #1044.

### DIFF
--- a/src/components/inspectors/TimerExpression.vue
+++ b/src/components/inspectors/TimerExpression.vue
@@ -61,6 +61,7 @@
             :class="{'date-disabled' : ends !== 'ondate'}"
             :disabled="ends !== 'ondate'"
             :placeholder="$t('End date')"
+            data-format="datetime"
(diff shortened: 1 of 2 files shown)


src/components/inspectors/TimerExpression.vue
```diff
@@ -61,6 +61,7 @@
             :class="{'date-disabled' : ends !== 'ondate'}"
             :disabled="ends !== 'ondate'"
             :placeholder="$t('End date')"
+            data-format="datetime"
```
chipit24: When the date-picker only deals with date formats, it sets the time portion to the start of the day, but the timer expression component sets the time portion to the same time as the start date. This seems to create a loop.

There's no such time resettnig when the date format is set to "datetime".

Issue #1051, opened 2020-01-23
Select sub process start event in the sub process inspector


nolanpro:
Re: https://github.com/ProcessMaker/processmaker/issues/2764

- Remove the pm:subProcess atribute from sequence flows
- Save the selected start event in the bpmn:callActivity's pm:config
- Increase the limit on the api call process count

PR #1052, opened 2020-01-23
Add start event to sub process pm:config

nolanpro:
Re: https://github.com/ProcessMaker/processmaker/issues/2764
Relies on: https://github.com/ProcessMaker/processmaker-bpmn-moddle/pull/33
Fixes: #1051 

Modify sub process node add start event setting

image


velkymx:
Add the label "Start Event" to the select for start events.

src/components/nodes/subProcess/index.js
```diff
@@ -29,18 +30,27 @@ export default {
...
-      setNodeProp(node, key, value[key]);
+    if (currentConfig.name != value.name) {
+      if (oldConfig.name === value.name || !oldConfig.name) {
+        // SubProcessFormSelect automatically updated the name so set the new name here
+        setNodeProp(node, 'name', currentConfig.name);
+      } else {
+        // The user is editing the name field manually, so update the config object name
+        currentConfig.name = value.name;
+      }
     }
```
chipit24: @nolanpro or @velkymx can one of you please shed some light on how this is suppose to work.

`calledElement` and `name` are already stored on the element (in the BPMN definition); duplicating that information in `config` seems redundant and introduces the extra logic above. If we stick to storing only `processId` and `startEvent` in `config`, and remove the convoluted logic around setting the node's name, then this whole `inspectorHandler` can be removed.

chipit24:
How it looks with updated label:
Screen Shot 2020-01-30 at 10 07 32 AM

src/components/nodes/subProcess/index.js
velkymx: @chipit24 the sub process could have more than one start event. We need to ensure you can select the start event you want to run.

nolanpro: @chipit24 yes name and calledElement are duplicated in the pm:config attribute. I only did it that way because I couldn't figure out a way to update any attributes other than 'pm:config' from within SubProcessFormSelect.vue, so it set it in there then set the other attributes in the code above. If you know how to set other attributes from within the inspector component please refactor it. Thanks

src/components/nodes/subProcess/index.js
```diff
@@ -29,18 +30,27 @@ export default {
...
-      }
 
-      if (key === 'callActivityExpression') {
-        this.inspectorHandler(value[key], node, setNodeProp);
-        continue;
-      }
+    setNodeProp(node, 'id', value.id);
+    setNodeProp(node, 'name', value.name);
+
+    const oldConfig     = JSON.parse(node.definition.config);
```
johnblackspear: Might cause some white-space only diffs in future here :)

Issue #1043, opened 2020-01-21
Shape movement by arrow key is not undo-able


johnblackspear:
# Bug report issue
## Description
Moving a shape with the arrow keys cannot be undone using the Undo button.

## Reproduction steps

1. Open modeler
2. Add a shape to the diagram
3. Move the shape using the arrow keys
4. Press Undo

## Expected behaviour
The shape should move to a previous position.

## Actual behaviour
The shape is removed from the paper.

## Software used
Firefox 72.0.1 x64 on macOs 10.14.6 Mojave

## Other information
n/a

PR #1054, opened 2020-01-23
Allow undo of keyboard arrow-key moves

johnblackspear:
Addresses [issue 1043](https://github.com/ProcessMaker/modeler/issues/1043) and also cleanup comments made in PR #1029.

Diagram elements moved with arrow keys can now be moved back via undo.

I've added a new test spec to safeguard the undo behaviour, and a new unit test to describe the movement behaviour when we don't have an actively selected shape.

I've also extracted some functions and renamed some parameters to highlight intent.

### DIFF
--- a/src/components/modeler/Modeler.vue
+++ b/src/components/modeler/Modeler.vue
@@ -125,9 +125,9 @@ import registerInspectorExtension from '@/components/InspectorExtensionManager';
 import initAnchor from '@/mixins/linkManager.js';
 import ensureShapeIsNotCovered from '@/components/shapeStackUtils';
 import ToolBar from '@/components/toolbar/ToolBar';
-import { keydownListener } from '@/components/modeler/moveWithArrowKeys';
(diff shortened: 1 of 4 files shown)


src/components/modeler/Modeler.vue
```diff
@@ -662,6 +662,18 @@ export default {
 
       this.paperManager.setPaperDimensions(clientWidth, clientHeight);
     },
+    keydownListener(event) {
+      const formElementHasFocus = event.target.toString().toLowerCase().indexOf('body') === -1;
```
chipit24: Do you think this would be more readable using [`includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes)?
```suggestion
      const formElementHasFocus = !event.target.toString().toLowerCase().includes('body');
```

johnblackspear: Definitely better yes

Issue #1056, opened 2020-01-24
Change focus when selecting input and then a shape


lisa-fehr:
https://github.com/ProcessMaker/modeler/pull/1054#pullrequestreview-348053243

> One thing I noticed is that if you select a form input in the inspector, and then click on the shape, the form input still has focus (tested in Chrome, FF and Safari). This could be confusing for users, and defientely frustrating, as I have to click somewhere outside of the grid to re-focus the body.

> I wonder if a click event listener can ba attached to .paper-container to handle this

PR #1066, opened 2020-01-28
Fix/1056 blur form on diagram focus

johnblackspear:
Defocus input elements in the right-hand inputs on element click.

Addresses [Issue 1056](https://github.com/ProcessMaker/modeler/issues/1056)

### DIFF
--- a/src/components/modeler/Modeler.vue
+++ b/src/components/modeler/Modeler.vue
@@ -321,9 +321,9 @@ export default {
       store.commit('highlightNode', node);
     },
     blurFocusedScreenBuilderElement() {
-      const elementsToBlur = ['input', 'select'];
+      const elementsToBlur = ['INPUT', 'SELECT'];
       const shouldBlurActiveElement = elementsToBlur.some((element) => {
-        return document.activeElement && document.activeElement.toString().toLowerCase().includes(element);
+        return document.activeElement && document.activeElement.tagName.includes(element);
       });
 
       if (shouldBlurActiveElement) {
--- a/src/components/modeler/Modeler.vue
+++ b/src/components/modeler/Modeler.vue
@@ -754,9 +754,7 @@ export default {
       store.commit('highlightNode', this.processNode);
     }, this);
 
-    this.paperManager.addEventHandler('element:pointerclick', () => {
-      this.blurFocusedScreenBuilderElement();
-    }, this);
+    this.paperManager.addEventHandler('element:pointerclick', this.blurFocusedScreenBuilderElement, this);
 
     this.paperManager.addEventHandler('blank:pointerdown', (event, x, y) => {
       const scale = this.paperManager.scale;
--- a/src/components/modeler/Modeler.vue
+++ b/src/components/modeler/Modeler.vue
@@ -320,6 +320,11 @@ export default {
     highlightNode(node) {
       store.commit('highlightNode', node);
     },
+    blurActiveElement() {
+      if (document.activeElement.type) {
+        document.activeElement.blur();
+      }
+    },
     /**
      * Register a mixin into a node component.
      * Used during "modeler-before-init"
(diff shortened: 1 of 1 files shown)


src/components/modeler/Modeler.vue
```diff
@@ -744,6 +754,10 @@ export default {
       store.commit('highlightNode', this.processNode);
     }, this);
 
+    this.paperManager.addEventHandler('element:pointerclick', () => {
+      this.blurFocusedScreenBuilderElement();
+    }, this);
```
chipit24: Since you're explicitly passing in `this` as the callback scope, this could be simplified using an arrow function and it should work the same:
```js
this.paperManager.addEventHandler('element:pointerclick', this.blurFocusedScreenBuilderElement, this);
```

src/components/modeler/Modeler.vue
```diff
@@ -320,6 +320,16 @@ export default {
     highlightNode(node) {
       store.commit('highlightNode', node);
     },
+    blurFocusedScreenBuilderElement() {
+      const elementsToBlur = ['input', 'select'];
+      const shouldBlurActiveElement = elementsToBlur.some((element) => {
+        return document.activeElement && document.activeElement.toString().toLowerCase().includes(element);
```
chipit24: Apparently there is a [`tagName`](https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName) property on elements:
```js
document.activeElement.toString() // "[object HTMLInputElement]"
document.activeElement.tagName // "INPUT"
```
That might simplify the code (not that it's not simple as-is, just a suggestion):
```js
const shouldBlurActiveElement = elementsToBlur.some((element) => {
  return document.activeElement && document.activeElement.tagName.toLowerCase() === element
});
```

If you capitalize the element names in the array, then it could also be:
```js
const shouldBlurActiveElement = elementsToBlur.includes(document.activeElement && document.activeElement.tagName);
```

johnblackspear: Aah brilliant @ `tagName`

src/components/modeler/Modeler.vue
```diff
@@ -320,6 +320,16 @@ export default {
     highlightNode(node) {
       store.commit('highlightNode', node);
     },
+    blurFocusedScreenBuilderElement() {
+      const elementsToBlur = ['INPUT', 'SELECT'];
+      const shouldBlurActiveElement = elementsToBlur.some((element) => {
+        return document.activeElement && document.activeElement.tagName.includes(element);
```
lisa-fehr: I was wondering if `document.activeElement.tagName` would ever be lowercase, but it seems like it isn't supposed to be.

src/components/modeler/Modeler.vue
johnblackspear: Updated :)

src/components/modeler/Modeler.vue
johnblackspear: Yeah seems not to, but you never know with, say, Microsoft... should we convert the case to be safe you think?

src/components/modeler/Modeler.vue
johnblackspear: Good catch, improved and updated.

src/components/modeler/Modeler.vue
chipit24: I don't think we need to worry about it; we're really only supporting Chromium (Chrome, new Edge), Safari and FF.

I just tested on latest Chrome, Safari, and FF on my mac, and also on Edge 18 and Edge 79 beta on Windows 10 in BrowserStack and it's returning uppercase strings in all cases.

The WHATWG DOM standard [mentions](https://dom.spec.whatwg.org/#dom-element-tagname):
> The tagName attribute’s getter must return this’s HTML-uppercased qualified name.

And the W3C DOM 4.1 spec [says](https://www.w3.org/TR/dom41/#dom-element-tagname):
> If the context object is in the HTML namespace and its node document is an HTML document, let qualifiedName be converted to ASCII uppercase.

johnblackspear: Good, let's keep to the standard

PR #1071, opened 2020-01-29
Add signal start event

### LINKED ISSUE #1061: Add Signal Start Event
Screen Shot 2020-01-27 at 2 44 56 PM

Acceptance criteria:
 * The signal start event should have a solid border with a triangle like in the screenshot above.
   * The event should have a green stroke colour (`#00BF9C`) like the other start events.
 * The signal start event should be accessible from the "cog" dropdown under the base start event.
 * It should produce the following BPMN XML:
   ```
   <startEvent id="signalStart" >
     <signalEventDefinition />
   </startEvent>
   ```
 * The inspector for the signal start event should have Name input under the Configuration accordion and a Node Identifier input under the Advanced accordion.

chipit24:
Fixes #1061.

Screen Shot 2020-01-29 at 10 19 31 AM


src/components/nodes/startEvent/startEvent.vue
```diff
@@ -63,6 +63,11 @@ export default {
           nodeType: 'processmaker-modeler-message-start-event',
           dataTest: 'switch-to-message-start-event',
         },
+        {
```
lisa-fehr: I put Boundary Signal Event above Message Event - since it can be disabled. Should we bump this one above too?

chipit24: Yeah, I'll move it up!

lisa-fehr:
Just a thought: should we be grouping these files in a format like:
StartSignalEvent vs SignalStartEvent

So It groups with:
StartMessageEvent
StartTimerEvent

chipit24:
That's a really good point. Start with the most generic part of the name first, then add the more specific parts. The official Vue style guid also mentions this: https://vuejs.org/v2/style-guide/#Order-of-words-in-component-names-strongly-recommended.

johnblackspear:
Just checking my understanding according to the style guide @chipit24 linked - shall we standardise on `<Node><SubType><Action>`; so for example `EventSignalStart` etc?

chipit24:
> Just checking my understanding according to the style guide @chipit24 linked - shall we standardise on `<Node><SubType><Action>`; so for example `EventSignalStart` etc?

@johnblackspear I think `<Action>` in your example is more generic than `<SubType>`.

We have one big group: events. Under events we have 3 smaller groups: start, intermediate, and end. And under each of those we have our specific event types (error, signal, message, etc.). So I think the correct naming would be something like:

`[Event][Start|Intermediate|End][Timer|Message|Signal|Error|etc.]`

E.g.:
`EventStartSignal`.

@lisa-fehr What do you think?

lisa-fehr:
@chipit24 I was thinking [Start|Intermediate|End][Timer|Message|Signal|Error|etc.][Event]

Another ticket could move Event to the front of all nodes?

chipit24:
Yeah that would make sense too. Since our start events all extend the base start event so grouping them is logical from that perspective. I'll keep the naming the way it is in this PR and another PR/issue could be created to normalize the naming.

Issue #1062, opened 2020-01-27
Add Signal End Event


chipit24:
<img width="78" alt="Screen Shot 2020-01-27 at 2 57 56 PM" src="https://user-images.githubusercontent.com/7561061/73209021-6e5e1680-4115-11ea-8263-43913a4b81c2.png">

Acceptance criteria:
 * The signal end event should have a solid border with a triangle like in the screenshot above.
   * The event should have a red stroke colour like the other end events.
 * The signal end event should be accessible from the "cog" dropdown under the base end event.
 * It should produce the following BPMN XML:
   ```xml
   <endEvent id="signal">
     <signalEventDefinition />
   </endEvent>
   ```

johnblackspear:
Red colour as per the other elements is #ED4757

PR #1072, opened 2020-01-29
Add Signal End Event

johnblackspear:
Addresses [ticket #1062](https://github.com/ProcessMaker/modeler/issues/1062).


src/assets/signal-end-event.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="18" height="15" viewBox="0 0 18 15" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M9 0L18 15H0L9 0Z" fill="#ED4757"/>
+</svg>
```
chipit24: @lisa-fehr and I have been using a rounded version of this symbol 😬maybe once this is merged we can go back and adopt your superior pointy triangle.

src/components/nodes/signalEndEvent/index.js
```diff
@@ -0,0 +1,21 @@
+import component from './signalEndEvent.vue';
+import endEventConfig from '../endEvent/index';
```
chipit24: When you point an import to a folder and don't specify a filename, it will assume `index.js` by default; this is what we've been doing in all the other "index" node files. It might be worth switching this to keep it consistent, but at the same time, I kind of like explicitly adding the filename.

src/components/nodes/signalEndEvent/signalEndEvent.vue
```diff
@@ -0,0 +1,21 @@
...
+    this.shape.attr({
+      image: {
+        'ref-y': 2,
+        'xlink:href': signalEndEventIcon,
+      },
+    });
+  },
+  destroyed() {
+    pull(this.rootElements, this.error);
+  },
```
chipit24: Might have missed this when pairing with you–sorry. But since we're not creating an error element, we don't have to worry about removing it. So this whole `destroyed` block (along with the single prop it was accepting) can be removed.

tests/e2e/specs/SignalEndEvent.spec.js
```diff
@@ -0,0 +1,15 @@
+import { addNodeTypeToPaper, assertDownloadedXmlContainsExpected } from '../support/utils';
+import { nodeTypes } from '../support/constants';
+
+describe('Signal Start Event', () => {
+  it('Can create signal start event', () => {
```
chipit24: Missed this: Should re-name "start" to "end".

src/components/nodes/signalEndEvent/signalEndEvent.vue
johnblackspear: Super, will do

tests/e2e/specs/SignalEndEvent.spec.js
johnblackspear: Uh, yep

src/components/nodes/signalEndEvent/index.js
johnblackspear: I'll switch to the standard here :)

src/assets/signal-end-event.svg
johnblackspear: Aargh not the ~lime juice~ rounded corners

src/components/nodes/signalEndEvent/signalEndEvent.vue
```diff
@@ -1,21 +1,17 @@
 <script>
 import EndEvent from '@/components/nodes/endEvent/endEvent';
 import signalEndEventIcon from '@/assets/signal-end-event.svg';
-import pull from 'lodash/pull';
 
 export default {
   extends: EndEvent,
   props: ['rootElements'],
```
chipit24: Looks like you forgot this. The `rootElements` props isn't used anymore now and can be removed.
```suggestion
```

PR #1090, opened 2020-02-05
Add multiselect

### LINKED ISSUE #1100: Implement multiselect
Allow selecting and dragging multiple shapes.

Acceptance criteria:
 * While holding down shift and clicking on other elements, the other elements should become highlighted as well.
 * While holding down shift and dragging:
    * Panning of the paper should be disabled.
    * A selection box should be drawn from the point of click to where the cursor is being dragged.
    * Once the mouse is released, all elements within the selection box should become highlighted.
 * Even when multiple elements are selected, the inspector panel should show the inspector config for the 1st selected element.
 * When pressing and dragging on an element, or moving an element with the arrow keys, all other highlighted shapes should also move by the same amount.
 * When multiple elements in a pool are highlighted, or a pool is included in the highlighted elements, clicking and dragging should work as normal (only the selected element should be moved, not all highlighted elements)
    * This constraint is only added for now to limit the scope of this feature, as pools add many more edge cases.

chipit24:
Fixes #1100.

Todo:
- [x] ~Fix issue with dragging multiple shapes outside of a pool.~ Prevent dragging multiple pool elements
- [x] Allow dragging selection box to the left / top.
- [x] Fix tests.
- [x] Add tests for multiselect (selecting with shift+click, shift+drag, and moving multiple shapes).


chipit24:
The same test is failing for `BoundaryErrorEvent.spec.js`, `BoundaryMessageEvent.spec.js` and `BoundaryTimerEvent.spec.js`:
```
should turn pool red when hovered over and then back to default colour when no longer over pool
```
However, these tests are passing for me locally.

chipit24:
The tests that are failing on CI are all passing for me locally 😭.
The other test is from `MessageFlows.spec.js`:
```
Cannot connect to itself
```

tests/e2e/support/utils.js
```diff
@@ -133,7 +133,9 @@ export function connectNodesWithFlow(flowType, startPosition, endPosition, click
           cy.wrap($endElement)
             .trigger('mousemove', { clientX: x, clientY: y, force: true })
             .wait(renderTime)
-            .click(clickPosition, { force: true });
+            .trigger('mousedown', { force: true })
+            .click(clickPosition, { force: true })
```
lisa-fehr: Clicking while moused down? Is the `.trigger('mousedown', { force: true })` required?

src/components/inspectors/InspectorPanel.vue
```diff
@@ -76,7 +76,7 @@ export default {
   },
   computed: {
     highlightedNode() {
-      return store.getters.highlightedNode;
+      return store.getters.highlightedNodes[0];
```
johnblackspear: I like this - but once it's out, we can gather feedback on whether we actually want a default inspector panel when multiple shapes are selected.

src/components/modeler/setUpSelectionBox.js
```diff
@@ -0,0 +1,136 @@
...
+import store from '@/store';
+
+const shapesToNotTranslate = [
+  'processmaker.modeler.bpmn.pool',
+  'PoolLane',
+  'processmaker.components.nodes.boundaryEvent.Shape',
+];
+
+export default function setUpSelectionBox(setCursor, resetCursor, paperManager, graph) {
+  let initialPositionByGerrie;
```
johnblackspear: Probs a good idea to rename this guy, haha :D

tests/e2e/support/printLogsToTerminal.js
```diff
@@ -0,0 +1,33 @@
+/* This file allows console.log's to be displayed in the terminal to help with debugging
```
johnblackspear: This is cool, was looking for something like this, good find

tests/e2e/specs/MessageFlows.spec.js
```diff
@@ -99,17 +125,14 @@ describe('Message Flows', () => {
 
     getNumberOfLinks().should('equal', 0);
 
-    connectNodesWithFlow('message-flow-button', poolPosition, poolPosition);
```
lisa-fehr: Connecting to itself isn't an invalid target?

src/components/inspectors/InspectorPanel.vue
chipit24: I did this mostly because it was the easiest way to prevent errors with the inspector with multiple shapes. Getting real user feedback on things like this would be very nice.

Another option would be to show some sort of common inspector or collection of things that can be applied to all selected shapes here.

tests/e2e/support/utils.js
chipit24: Yeah, clicking while moused down doesn't really make sense. I'll remove it and see if it changes the tests.

tests/e2e/specs/MessageFlows.spec.js
chipit24: It is, but leaving it here for some reason caused this test to fail on CI. So I've moved it into the test above, "Cannot connect to itself", where I think it made more sense anyway:

```js
[poolPosition, startEventPosition, taskPosition].forEach(position => {
  connectNodesWithFlow('message-flow-button', position, position);
});
```

src/components/inspectors/InspectorPanel.vue
johnblackspear: Yeah things we can discuss with Alan :)

PR #1091, opened 2020-02-05
Console error validating parallel gateway

### LINKED ISSUE #1086: Error thrown when validating Parallel Gateway
## Description
A console error is thrown when validation is toggled for a process containing a parallel gateway.

## Reproduction steps
1. Drag a parallel gateway onto the grid.
2. Drag 3 tasks onto the grid.
3. Connect 2 of the tasks to the gateway.
4. Connect the gateway to the 3rd task that is not yet connected to anything.
5. Toggle Auto validate.

## Expected behaviour
No console errors are thrown.

## Actual behaviour
The following console error is thrown:
Screen Shot 2020-02-03 at 2 21 18 PM
```
[Vue warn]: Duplicate keys detected
```

## Software used
Latest modeler `develop` branch on Chrome 79 on macOS Catalina.

lisa-fehr:
Closes #1086 

Wasn't sure how to assert that the console log didn't contain 'Duplicate keys'.
Was able to assert that there were no errors. But this could catch other site errors (which are bad anyway).


src/components/validationStatus/errorListUtil.js
```diff
@@ -0,0 +1,13 @@
+import uniqueId from 'lodash/uniqueId';
+
+export default function errorList(validationErrors) {
+  return Object.entries(validationErrors).flatMap(([errorKey, errors]) => {
+    return errors.flatMap(error => {
+      return {
+        ...error,
+        errorKey,
+        ...{ 'errorId': `${ uniqueId('id') }_${ error.id }` },
```
chipit24: I don't think this needs to be descructured like this. Can it be set up as a key/value on the returned object?
```js
return {
  ...error,
  errorKey,
  errorId: `${ uniqueId('id') }_${ error.id }`
}
```

However, `uniqueId` is simply returning an auto-incrementing value, and at that point you may as well use the index as the key in the `v-for` in the template. I think what would work better as a unique key is the error ID (which I guess is the node ID in our case) and the error message.

The way it's set up now with `uniqueId`, if there is an actual bug and the exact same error gets duplicated in the list, no error will be thrown since it'll be assign a unique ID anyway.

The ID and message (node ID and error message) should be unique, and if it's not, we should expect to see a duplicate key warning.

lisa-fehr: Yeah. I wasn't sure if I should try to turn the message into an id.

src/components/validationStatus/ValidationStatus.vue
```diff
@@ -108,12 +109,7 @@ export default {
       },
     },
     errorList() {
-      return Object.entries(this.validationErrors)
-        .flatMap(([errorKey, errors]) => {
-          return errors.flatMap(error => {
-            return { ...error, errorKey };
-          });
-        });
+      return validationErrorList(this.validationErrors);
```
johnblackspear: Like the extracted method here

PR #1094, opened 2020-02-06
Restored original negation for boundary events in pools

### LINKED ISSUE #1093: Boundary events moves around when dragging pool
## Description
When you drag a pool that contains a boundary event, the boundary event moves around but it should stay attached to the task.

## Reproduction steps
1. Drag a task onto the grid.
2. Add a boundary event onto the task.
3. Drag a pool onto the grid.
4. Drag the pool around. 

## Expected behaviour
The boundary event should stay attached to the pool and not move in relation to the task it is attached to.

## Actual behaviour
The boundary event flies around:
boundary-event-pool-bug

## Software used
Latest modeler develop branch running on Chrome 79 on macOS Catalina.

lisa-fehr:
Closes #1093 

Small miss during a refactor.
Screen Shot 2020-02-06 at 4 59 53 PM

### DIFF
--- a/tests/e2e/specs/Pools.spec.js
+++ b/tests/e2e/specs/Pools.spec.js
@@ -14,25 +14,7 @@ import {
 } from '../support/utils';
 
 import { nodeTypes } from '../support/constants';
-
-function boundaryEventIsCloseToTask() {
-  const positionErrorMargin = 30;
-
-  const taskSelector = '.main-paper ' +
-    '[data-type="processmaker.components.nodes.task.Shape"]';
-
-  const boundaryTimerEventSelector = taskSelector +
-    ' ~ [data-type="processmaker.components.nodes.boundaryEvent.Shape"]';
-
-  cy.get(taskSelector).then($task => {
-    const { left: taskLeft, top: taskTop } = $task.position();
-    cy.get(boundaryTimerEventSelector).then($boundaryEvent => {
-      const { left, top } = $boundaryEvent.position();
-      expect(left).to.be.closeTo(taskLeft, positionErrorMargin);
-      expect(top).to.be.closeTo(taskTop, positionErrorMargin);
-    });
-  });
-}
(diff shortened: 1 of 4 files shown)


lisa-fehr:
Thought we already had a test for something like this. Maybe it was too tricky to do.

tests/e2e/specs/Pools.spec.js
```diff
@@ -15,6 +15,25 @@ import {
...
 import { nodeTypes } from '../support/constants';
 
+function boundaryEventIsCloseToTask() {
+  const positionErrorMargin = 30;
+
+  const taskSelector = '.main-paper ' +
+    '[data-type="processmaker.components.nodes.task.Shape"]';
+
+  const boundaryTimerEventSelector = taskSelector +
+    ' ~ [data-type="processmaker.components.nodes.boundaryEvent.Shape"]';
```
chipit24: This line, without the `~`, is exported in `tests/e2e/support/constants.js`:
https://github.com/ProcessMaker/modeler/blob/81ab6700ae8ae9baa5f137380d8f7f0996e13610/tests/e2e/support/constants.js#L42-L43

Maybe you can update that and use that export here as well?

tests/e2e/specs/Pools.spec.js
```diff
@@ -15,6 +15,25 @@ import {
 
 import { nodeTypes } from '../support/constants';
 
+function boundaryEventIsCloseToTask() {
```
chipit24: What do you think about renaming this to something like `assertBoundaryEventIsCloseToTask`? Similar to what I've done with the `assertDownloadedXmlContainsExpected` helpers. Currently, for me, this reads like it should return a boolean.

lisa-fehr: Sure. I thought about making it share-able. But it relies on there only being one task with a boundary event.

PR #1097, opened 2020-02-07
WIP: Spike for align and distribute

johnblackspear:
This is pretty scrappy - but it does spike out some basic alignment / and distribution. Horizontal distribution is skipped until we discuss this a little more :)

It does not yet take any subtleties into account and will move stuff pretty randomly since I don't filter for non-translatable or certain child shapes. 

Relates to #1098


src/components/nodes/utilities/AlignShapes.js
```diff
@@ -0,0 +1,122 @@
...
+// TODO: this needs to be associated with the multi-selection box logic. We can
+// also pre-emptively recalc this everytime we add / remove items from a
+// multiselect
+export function getBoundingBox(shapes) {
+  const left = (shape) => shape.position().x;
+  const right = (shape) => shape.position().x + shape.size().width;
+  const top = (shape) => shape.position().y;
+  const bottom = (shape) => shape.position().y + shape.size().height;
+
+  const minX = minimum(shapes, left);
```
chipit24: This is a cool helper. I wonder if using https://lodash.com/docs/4.17.15#minBy or `Math.min(shapes.map(left))` would simplify things.

src/components/modeler/setUpSelectionBox.js
```diff
@@ -35,6 +44,32 @@ export default function setUpSelectionBox(setCursor, resetCursor, paperManager,
   let selectionBox;
 
   function shiftKeyDownListener(event) {
+
+    const highlightedShapes = store.getters.highlightedShapes;
```
chipit24: This is great for testing but I think in final-form they'd work nicely under icons in the toolbar.

johnblackspear: Yeah this needs to never go near actual humans :D

src/components/nodes/utilities/AlignShapes.js
johnblackspear: Super, thanks, that's better :)

johnblackspear:
No longer needed as we've done the exploration for basic functionality that we wanted to

PR #1099, opened 2020-02-10
Copy and Paste Element

### LINKED ISSUE #1092: Copy and paste elements
Copy and paste some elements. Should copy inspector information as well.

Acceptance criteria:

 - [x]  A "Copy" button should appear in the crown for certain elements (for now, only: start events, intermediate events, end events, tasks, text annotations, and gateways).
    - [x]  Copy should not be available for pools, lanes, boundary events, or flows.
 - [x]  When the "copy" button is clicked, a copy of the element will be added to the paper, offset from the position of the original copied element so that it doesn't overlap / cover it.
    - [x]  Pressing copy multiples times should offset the new copy from the previous copy, so copied elements don't overlap / disappears behind one another.
 - [x]  The copied element should not include any flows or boundary events.
 - [x]  All inspector config properties should be copied over to the new element (all configuration, timing control, advanced properties, except for the node ID).
    - [x]  The copied element's ID should be auto-generated (as if it were manually dragged from the controls list).

tuantruong625:
Fixes #1092.

node_copy


tests/e2e/specs/EventBasedGateway.spec.js
```diff
@@ -59,7 +59,7 @@ describe('Event-Based Gateway', () => {
     getElementAtPosition(taskPosition)
       .click()
       .then($task => {
-        getCrownButtonForElement($task, 'delete-button').click();
+        getCrownButtonForElement($task, 'delete-button').click({ force: true });
```
chipit24: This was required since the copy crown button pushed the delete crown button off screen.

chipit24:
The branch name is `spike/copy-paste-element`, but this is a legit implementation and didn't end up being purely exploratory, just FYI. Not going to change the branch name at this point.

src/components/crown/crownButtons/copyButton.vue
```diff
@@ -0,0 +1,55 @@
...
+        'processmaker-modeler-intermediate-message-throw-event',
+        'processmaker-modeler-call-activity',
+        'processmaker-custom-marker-task-test',
+      ],
+    };
+  },
+  methods: {
+    copyElement() {
+      this.copyCount++;
+      this.$emit('copy-element', this.node, this.copyCount);
```
lisa-fehr: I wonder if you could do this instead:
```suggestion
      this.$emit('copy-element', this.node, this.copyCount++);
```

src/components/nodes/node.js
```diff
@@ -39,6 +41,11 @@ export default class Node {
     }
   }
 
+  clone() {
+    const diagramClone = cloneDeep(this.diagram);
+    return new Node(this.type, diagramClone.bpmnElement, diagramClone);
```
lisa-fehr: Didn't copy Notifications in the core application, for some reason.

lisa-fehr:
Thoughts?

**Error End Event** should have a unique Error Name.
**Message End Event** should have a unique Message Name.
**Intermediate Message Throw Event** should have a unique Message Name.

Screen Shot 2020-02-13 at 1 37 21 PM

chipit24:
I've updated the PR to set `errorRef` and `messageRef` properties to `null`.  This will ensure a new message or error element is created and that the old ref is not copied over.

But another thing I noticed was that if I copy a Message Start Event, the "Listen For Message" property isn't copied, and the instance type of the object has changed as well, from `ModdleElement` to `Base`. I'm not sure what implications that has, but it doesn't make me feel good about the cloned element.

I also started switching on element properties when doing clone. I think what I'll do moving forward it to extend the Node class and create different Node classes for the different element types, and implement clone methods specific to each (i.e. "replace conditional with polymorphism"). This will also allow us the freedom to handle complicated cases like the pool.

tests/unit/node.spec.js
```diff
@@ -8,10 +8,26 @@ describe('Node', () => {
...
+
+  it('should set cloned event definition errorRef value to null', () => {
+    const definition = { eventDefinitions: [{ errorRef: {} }] };
+    const node = new Node('type', definition, { bpmnElement: definition });
+    const clonedNode = node.clone();
+
+    expect(clonedNode.definition.eventDefinitions[0].errorRef).toBeNull();
+  });
+
+  it('should set cloned event definition errorRef value to null', () => {
```
lisa-fehr:
```suggestion
  it('should set cloned event definition messageRef value to null', () => {
```

tests/e2e/specs/EventBasedGateway.spec.js
johnblackspear: Oof that scroll-to thing

lisa-fehr:
Not sure if it is me. If I copy an element, it doesn't copy as much.

IE: Copied a start timer with weekdays selected and an end date. Copied the copy and it didn't have the weekdays and end date.

chipit24:
@lisa-fehr So this was because the cloned elements created where instances of `Node` and not `TimerEventNode`. Should be fixed now.

src/components/nodes/node.js
```diff
@@ -41,18 +44,29 @@ export default class Node {
...
+    const diagramClone = cloneDeepWith(this.diagram, this.clearEventDefinitionRefs);
+    const clonedNode = new Node(this.type, diagramClone.bpmnElement, diagramClone);
     clonedNode.id = null;
     Node.diagramPropertiesToCopy.forEach(prop => clonedNode.diagram.bounds[prop] = this.diagram.bounds[prop]);
     clonedNode.definition.name = this.definition.name;
 
     return clonedNode;
   }
 
+  clearEventDefinitionRefs(value) {
```
lisa-fehr: This doesn't work well in core. Checking it out.

lisa-fehr:
@chipit24 @johnblackspear think I have it ironed out for copying task configuration again (except for Notifications in core). It will copy a copy of a task. I don't think it reverts any other fixes for Start Timer (copies time config) or End Message (Message Name is unique). Take a look when you have time.

lisa-fehr:
Found a bug:

Import a bpmn with a start timer and a task
- add config details to the task and copy it
- add timer config details to the start timer and copy it

The configs are not copied on the first copy. The second copy does work.

chipit24:
A get this error when copying an Intermediate Timer Event:
```
Uncaught (in promise) TypeError: Cannot read property 'isGeneric' of undefined
```
This seems to come from trying to serialize the BPMN data; so I guess the copy command incorrectly sets some BPMN definition property.

To reproduce:
1. Drag an Intermediate Timer Event onto the grid.
2. Expand the Timing Control accordion and change the Type from Duration to Date/Time or Cycle.
3. Copy the element.

src/components/modeler/Modeler.vue
```diff
@@ -539,7 +550,20 @@ export default {
         definition.set('name', '');
       }
 
-      store.commit('addNode', new Node(type, definition, diagram));
+      const node = this.getNodeType(type, definition, diagram);
+
+      store.commit('addNode', node);
+    },
+    getNodeType(type, definition, diagram) {
```
chipit24: Do you think a name like `getNode` or `createNode` would read better here? I saw "type" and my brain assumed this is going to return the node type, like the string `"processmaker-modeler-task"`.

src/components/nodes/taskNode.js
```diff
@@ -0,0 +1,19 @@
+import Node from './node';
+import cloneDeep from 'lodash/cloneDeep';
+
+export default class TaskNode extends Node {
+  clone(nodeRegistry, moddle, $t) {
+    let definition = nodeRegistry[this.type].definition(moddle, $t);
+    const diagramClone = cloneDeep(this.diagram);
```
chipit24: Is cloning the diagram like this required? I had issues with lodash's `clone` and `cloneDeep` methods as it didn't copy everything correctly. Do you need to copy more than what's specified in `Node.diagramPropertiesToCopy`?

src/components/modeler/Modeler.vue
lisa-fehr: I started with getNode, but had some doubts. Can switch it.

src/components/nodes/taskNode.js
lisa-fehr: With this cloning, it seems to get additional configurations on tasks. Like `DueIn`, `Users` & `Assignment Rules`. And for the `connector-send-email`, it copies the `Email Notifications`.

chipit24: Lodash's clone methods seem to modify the object in unexpected ways:

Screen Shot 2020-03-13 at 11 26 26 AM

You can see above that the regular element is referred to as `ModdleElement` and has some additional properties, but when cloned, it's now simply `Base` and is missing a few things from before. This may be related to the `isGeneric` error I experienced.

src/components/modeler/Modeler.vue
chipit24: Maybe `createNodeFromType`?

Issue #1098, opened 2020-02-07
Align and distribute shapes


johnblackspear:
# Enhancement: Align and distribute
## Description
We should be able to align and distribute shapes that are (multi-)selected on the page.

## Align
We should be able to
 - align to the left (all items line up with the minimum x value of the group of selected items)
 - align to the right
 - align to the top
 - align to the bottom
 - align centers along the vertical axis
 - align centers along the horizontal axis

## Distribute
Elements will be distributed by making sure the space between their centres is the same (distribute centre). This is how distribute is implemented in Camunda.

## Proposed alignment options:
![image](https://user-images.githubusercontent.com/28595383/74067135-6ee98d80-49f0-11ea-9048-230ad9fce798.png)

## Proposed distribution options:
![image](https://user-images.githubusercontent.com/28595383/74067163-7f016d00-49f0-11ea-950d-ae84fc0d12c6.png)

## Acceptance criteria
 * Create an icon for each of the alignment and distribution actions above (8 icon buttons in total).
    * These icons should be placed in the toolbar with the other buttons.
    * The buttons should be disabled until at least 2 shape are selected.
    * Remove the shortcut keys that currently map to these actions; use only the buttons.
 * Clicking the toolbar buttons should align / distribute the selected shapes as outlines in the sections above.
 * This feature shouldn't affect boundary events, flows (sequence flows, message flows, association flows), or lanes.
 * With pools:
    * If pools are included in the selection, apply align / distribute only to the pools, not the elements inside the pools.
    * If there are pool elements included in the selection, apply align / distribute relative to the elements inside each pool, not globally for all elements in every pool.

johnblackspear:
Some current issues with the WIP:

 - [ ] shapes without a `size()` breaks the bounding box logic

chipit24:
This is now complete! 🎊🇨🇦👨‍💻👩‍💻

PR #1108, opened 2020-02-12
Feature/shape metrics

johnblackspear:
Adds some basic shape metrics. 

This is used to build align and distribute functionality in a later PR.

Related to (but not fixing) #1098

### DIFF
--- a/src/components/nodes/utilities/shapeMetrics.js
+++ b/src/components/nodes/utilities/shapeMetrics.js
@@ -1,0 +1,15 @@
+const shapeBottom = (shape) => shape.position().y + shape.size().height;
+const shapeLeft = (shape) => shape.position().x;
+const shapeCenterX = (shape) => shape.position().x + (shape.size().width / 2);
+const shapeCenterY = (shape) => shape.position().y + (shape.size().height / 2);
+const shapeRight = (shape) => shape.position().x + shape.size().width;
+const shapeTop = (shape) => shape.position().y;
+
+module.exports = {
+  shapeBottom,
+  shapeLeft,
+  shapeCenterX,
+  shapeCenterY,
+  shapeRight,
+  shapeTop,
+};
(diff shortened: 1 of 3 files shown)


src/components/nodes/utilities/shapeMetrics.js
```diff
@@ -0,0 +1,15 @@
+const shapeBottom = (shape) => shape.position().y + shape.size().height;
+const shapeLeft = (shape) => shape.position().x;
+const shapeCenterX = (shape) => shape.position().x + (shape.size().width / 2);
+const shapeCenterY = (shape) => shape.position().y + (shape.size().height / 2);
+const shapeRight = (shape) => shape.position().x + shape.size().width;
+const shapeTop = (shape) => shape.position().y;
+
+module.exports = {
```
johnblackspear: @chipit24 @lisa-fehr @tuantruong625 
Is it OK to do this here or do we standardise on individual export statements?

chipit24: We don't use `module.exports` anywhere else in our source code. We've standardized on individual import / exports statements, what's knows as "ES modules". The `module.exports` import counterpart is `require`, it's what's known as CommonJS modules.

It's unorthodox to mix the two so let's stick with individual named ES module exports.

chipit24: Looks like JointJS has methods for calculating all this on it's `Rect` object: https://resources.jointjs.com/docs/jointjs/v3.1/geometry.html#g.Rect.prototype. The only thing is, we'd have to create a rectangle first, like `const rect = g.Rect(x, y, width, height)`.

If it's possible to use that, we could get rid of all this code 😄

johnblackspear: That's fairly easy to do - I'm going to keep the basic signature the same so that we can get to a basic version of align & distribute and then look at stripping out the first-pass code.

PR #1109, opened 2020-02-12
Add align and distribute utilities

### LINKED ISSUE #1098: Align and distribute shapes

johnblackspear:
Adds some basic align and distribute utilities that we can use on the front end.

Relies on PR https://github.com/ProcessMaker/modeler/pull/1108

Related to but does not fix #1098

### DIFF
--- a/tests/unit/customMatchers/toHaveBeenProgrammaticallyMoved/index.js
+++ b/tests/unit/customMatchers/toHaveBeenProgrammaticallyMoved/index.js
@@ -1,0 +1,36 @@
+import get from 'lodash/get';
+import passes from './predicate.js';
+
+const failMessage = (shape) => () => {
+  return this.utils.matcherHint('.toHaveBeenProgrammaticallyMoved', 'shape', '') +
+    '\n\nExpected:' +
+    '\n Shape to have been programmatically moved.' +
+    '\n\nReceived: ' +
+    (get(shape, 'translate.mock.calls', []).length === 0
+      ? '\n Shape has not been moved.'
+      : `\n Move has been called, but with dx==0, dy==0: ${this.utils.printReceived(shape.translate.mock.calls)}`);
+};
+
+const passMessage = (shape) => () => {
+  return this.utils.matcherHint('.not.toHaveBeenProgrammaticallyMoved', 'shape', '') +
+    '\n\nExpected:' +
+    '\n Shape to not have been programmatically moved.' +
+    '\n\nReceived: ' +
+    `\n Shape has been moved by: ${this.utils.printReceived(shape.translate.mock.calls)}`;
+};
+
+/**
+ * Matcher to ensure that a shape has been (programmatically) moved.
+ * @param shape
+ */
+export default function toHaveBeenProgrammaticallyMoved(shape) {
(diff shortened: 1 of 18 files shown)


tests/unit/customMatchers/toHaveBeenProgrammaticallyMoved/index.js
```diff
@@ -0,0 +1,37 @@
+import { matcherHint, printReceived } from 'jest-matcher-utils';
```
chipit24: Looking at the docs at https://jestjs.io/docs/en/expect#thisutils, it seems like these can be accessed through `this.utils`. Was that not working for you? Maybe it's only available after calling `expect.extend`?

johnblackspear:
I looked at the guide and it didn't dive into some I looked for (and I also want to test the predicate logic apart from the matcher return strings / `this.isNot` logic). So I dug into the jest code and also into the source here: https://github.com/jest-community/jest-extended/tree/master/src/matchers

I like their way of doing things, hence the structure here.

tests/unit/customMatchers/toHaveBeenProgrammaticallyMoved/index.js
johnblackspear: I'm pretty sure it would work too - it's basically doing the exact same thing, I've just been taking inspiration from the source mentioned below, and that was done using the direct import. Either way is fine, lemme know what you prefer style-wise.

package.json
```diff
@@ -84,6 +84,7 @@
     "eslint-plugin-vue": "^5.2.3",
     "i18next": "^15.0.8",
     "istanbul-lib-coverage": "^2.0.5",
+    "jest-matcher-utils": "^25.1.0",
```
chipit24: Since you removed the references to `jest-matcher-utils`, should this be uninstalled as well?

johnblackspear: Good catch, thanks

PR #1110, opened 2020-02-12
Feature/align shapes

johnblackspear:
This adds basic shape alignment functions.

Relies on https://github.com/ProcessMaker/modeler/pull/1109

Addresses part of #1098

### DIFF
--- a/src/components/nodes/utilities/shapeGroup.js
+++ b/src/components/nodes/utilities/shapeGroup.js
@@ -1,0 +1,28 @@
+/**
+ * Utility functions that operate on a collection of shapes
+ */
+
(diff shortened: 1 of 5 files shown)


src/components/nodes/utilities/shapeGroup.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * Utility functions that operate on a collection of shapes
+ */
+
+export function getBoundingBox(shapes) {
```
johnblackspear: This is definitely replaceable by https://resources.jointjs.com/docs/jointjs/v3.1/geometry.html#g.Rect.prototype

PR #1111, opened 2020-02-12
Feature/distribute shapes

johnblackspear:
Adds distribution of shapes to the utilities.

Relies on PR https://github.com/ProcessMaker/modeler/pull/1110

Addresses #1098 .

### DIFF
--- a/src/components/nodes/utilities/distribute.js
+++ b/src/components/nodes/utilities/distribute.js
@@ -57,7 +57,8 @@ export function distributeHorizontalCentersEvenly(shapes) {
 export function distributeHorizontalSpacingEvenly(shapes) {
   const bounds = getBoundingBox(shapes);
   const itemCount = shapes.length;
-  if (itemCount < 3) {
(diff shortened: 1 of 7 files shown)


src/components/nodes/utilities/distribute.js
```diff
@@ -0,0 +1,74 @@
+import { moveShapeLeftTo, moveShapeMiddleXTo, moveShapeMiddleYTo } from '@/components/nodes/utilities/shapeMovement';
+import { getBoundingBox } from '@/components/nodes/utilities/shapeGroup';
+
+export function distributeVerticalCentersEvenly(shapes) {
+  const bounds = getBoundingBox(shapes);
+  const itemCount = shapes.length;
+  if (itemCount < 3) {
```
lisa-fehr: Should this `3` be a variable explaining why that is the limit?

johnblackspear: Yeah, thanks :)

PR #1114, opened 2020-02-12
Add align and distribute options to toolbar

johnblackspear:
This adds the alignment and distribution functions to the toolbar.

image

From left to right:

1. align left
2. center horizontally
3. align right
4. align bottom
5. center vertically
6. align top
7. distribute horizontally
8. distribute vertically

Icons and layout to be confirmed.

### DIFF
--- a/tests/e2e/support/constants.js
+++ b/tests/e2e/support/constants.js
@@ -7,6 +7,11 @@ export const defaultViewportDimensions = {
   height: 768,
 };
 
+/**
+ * Node types are strings returned by
+ * shape.component.node.type - they are not the string
+ * same as the shape type.
+ */
 export const nodeTypes = {
   startEvent: 'processmaker-modeler-start-event',
   messageStartEvent: 'processmaker-modeler-message-start-event',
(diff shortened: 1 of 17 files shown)


tests/e2e/support/constants.js
```diff
@@ -7,6 +7,11 @@ export const defaultViewportDimensions = {
   height: 768,
 };
 
+/**
```
johnblackspear: Using this file feels a bit strange to me - I don't want to have these strings proliferate all over the place, but keeping this in  the `tests` dir is also wrong. Where would we like to move this to?

chipit24: I like the idea of moving it. Maybe it can go into `src/components/nodes/nodeTypes.js`.

This might tie in nicely with the refactor I was thinking about for https://github.com/ProcessMaker/modeler/pull/1099#issuecomment-587941206. If we extract this to a "node map", it can be used in the tests, it can be used to create the various node classes I'm thinking of (if we specify a type and a class property, not just the string), and it can be used to replace the string literals that are spread all over the place.

There are some files where this ID is exported, e.g.:
https://github.com/ProcessMaker/modeler/blob/c5c61d2e427ccf9186b501bf3f3d027c42f045d9/src/components/nodes/pool/index.js#L5

But I found that keeping it there caused circular dependency issues; so moving these IDs to some central file to be used by all would be great.

johnblackspear: Wanna pair on this to make sure we align our two refactors? I think this would be the last change and then this PR can come out of draft.

PR #1115, opened 2020-02-13
Clear sequence flow default name

### LINKED ISSUE #1102: New Sequence Flow
image

Can we make the default value empty? It gets really busy fast with all the "New Sequence Flow"

chipit24:
Fixes #1102.

Screen Shot 2020-02-13 at 10 35 51 AM

### DIFF
--- a/src/components/nodes/sequenceFlow/index.js
+++ b/src/components/nodes/sequenceFlow/index.js
@@ -11,7 +11,7 @@ export default {
   control: false,
   definition(moddle) {
     return moddle.create('bpmn:SequenceFlow', {
-      name: '',
+      name: null,
     });
   },
   diagram(moddle) {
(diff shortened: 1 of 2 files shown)


src/components/nodes/sequenceFlow/index.js
```diff
@@ -65,7 +65,7 @@ export default {
           items: [
             {
               component: 'FormInput',
-              config: nameConfigSettings,
+              config: { ...nameConfigSettings, validation: null },
```
lisa-fehr: Should these still be required fields? The Issue did not specify.

chipit24: I just confirmed with Alan and nope, we should leave it optional; which makes sense from a UX perspective; less friction for the user as they aren't given an invalid element by default.

PR #1116, opened 2020-02-13
Fix allowed message flows

### LINKED ISSUE #1105: Message Flow are not available for some elements
# Message Flow are not available for some elements
## Description
Currently there are elements that could contain message elements but can not be connected using the message flows.

## Reproduction steps
Insert two pools, one pool with and End Message Event, the other one with and Start Message Event.

Try to draw a message flow between those elements.

End Message Flow Event --> Start Message Event

## Expected behavior
Any throw element (ex. Intermediate throw and End Event) with a Message Event Definition, must have a Message Flow button in its crown.

Any catch element (ex. Intermediate catch, boundary and Start Events) with a Message Event Definition, can be a valid target for a message flow.

image

## Actual behavior
Not all the elements like End with message have the Message Flow

image

chipit24:
Fixes #1105.

Screen Shot 2020-02-13 at 2 19 51 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,13 +5,17 @@
 version: 2.1
 orbs:
   # use Cypress orb from CircleCI registry
-  cypress: cypress-io/cypress@1.10.0
+  cypress: cypress-io/cypress@1.14.0
+executors:
+  with-chrome:
+    docker:
+      - image: 'cypress/browsers:node12.8.1-chrome80-ff72'
 workflows:
   build:
     jobs:
       # runs tests using Chrome browser in custom executor (Cypress docker image)
       - cypress/run:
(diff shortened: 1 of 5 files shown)


chipit24:
Another case of tests passing locally for failing on CI. The two failing tests are from `MessageFlows.spec.js`:
```
Cannot connect to itself
Cannot connect to invalid message flow targets
```

chipit24:
So these tests are passing for me locally both through the CLI and through the GUI:
Screen Shot 2020-02-18 at 10 10 54 AM
Screen Shot 2020-02-18 at 10 15 43 AM

The only difference I can think of is that the CI is running Chrome 76 (headless) while I'm running Chrome 80 locally. Going to update the Circle CI config to use the latest Chrome.

PR #1122, opened 2020-02-18
Add validation to Sub Process dropdowns

sanjacornelius:
<h2>Changes</h2>

 - Require a Process and a process Start Event to be selected when using the SubProcess node.
- Display error message when required dropdown fields have no selection.

Ref [processmaker/#2843](https://github.com/ProcessMaker/processmaker/issues/2843)

### DIFF
--- a/src/components/nodes/subProcess/SubProcessFormSelect.vue
+++ b/src/components/nodes/subProcess/SubProcessFormSelect.vue
@@ -2,22 +2,30 @@
   <div>
     <form-multi-select
       label="Process"
+      name="Process"
       helper="Select which Process this element calls"
       v-model="selectedProcess"
+      :showLabels="false"
+      :allow-empty="false"
       :disabled="processList.length === 0"
       :options="processList"
       optionContent="name"
       class="p-0 mb-2"
+      :validation="'required'"
     />
 
     <form-multi-select
       label="Start Event"
+      name="StartEvent"
       v-if="selectedProcess"
       v-model="selectedStartEvent"
       :disabled="startEventList.length === 0"
+      :allow-empty="false"
+      :showLabels="false"
       :options="startEventList"
       optionContent="name"
       class="p-0 mb-2"
+      :validation="'required'"
     />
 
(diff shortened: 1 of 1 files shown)


src/components/nodes/subProcess/SubProcessFormSelect.vue
```diff
@@ -2,30 +2,38 @@
...
+      name="Process"
       helper="Select which Process this element calls"
       v-model="selectedProcess"
+      :showLabels="false"
+      :allow-empty="false"
       :disabled="processList.length === 0"
       :options="processList"
       optionContent="name"
       class="p-0 mb-2"
+      :validation="'required'"
```
chipit24: Since you're passing in a string, you don't need to explicitly bind a string value.
```suggestion
      validation="required"
```

src/components/nodes/subProcess/SubProcessFormSelect.vue
```diff
@@ -2,30 +2,38 @@
...
     />
 
     <a
       v-if="selectedProcess"
       :href="`/modeler/${selectedProcess.id}`"
       target="_blank"
     >
-      Open Process
+      {{ this.$t('Open Process') }}
```
chipit24: `this` isn't required in Vue templates, and we always omit it; I think it's worth keeping that consistent.
```suggestion
      {{ $t('Open Process') }}
```

src/components/nodes/subProcess/SubProcessFormSelect.vue
```diff
@@ -2,30 +2,38 @@
...
+      name="StartEvent"
       v-if="selectedProcess"
       v-model="selectedStartEvent"
       :disabled="startEventList.length === 0"
+      :allow-empty="false"
+      :showLabels="false"
       :options="startEventList"
       optionContent="name"
       class="p-0 mb-2"
+      :validation="'required'"
```
chipit24: Same here.
```suggestion
      validation="required"
```

Issue #1120, opened 2020-02-18
Add Download BPMN button


chipit24:
Add a button to the validation footer to allow downloading the current process as a BPMN file.

Acceptance criteria:
 * Move the current "Download XML" button from the title bar to the left side of the validation bar footer.
 * The button should display "Download BPMN", and should prompt the user to save the current process an a `.bpmn` file.

chipit24:
Todo: Add missing download functionality to `ModelerApp.vue` in PM core app.

johnblackspear:
Thanks for taking this, Rob. AFAIK this is now done - please check Ryan's PR on Friday (https://github.com/ProcessMaker/processmaker/pull/2938)

PR #1124, opened 2020-02-19
Move download button to validation footer

chipit24:
Related issue: #1120.

**Note:** Another PR is still required in the processmaker core repo to update the container `ModelerApp.vue` component there as well.

### DIFF
--- a/src/ModelerApp.vue
+++ b/src/ModelerApp.vue
@@ -15,7 +15,7 @@
         </div>
       </b-card-header>
       <b-card-body class="overflow-hidden position-relative p-0 vh-100">
-        <modeler @set-xml-manager="xmlManager = $event" @validate="validationErrors = $event" @warnings="warnings = $event"/>
+        <modeler ref="modeler" @set-xml-manager="xmlManager = $event" @validate="validationErrors = $event" @warnings="warnings = $event"/>
       </b-card-body>
       <validation-status
         :validation-errors="validationErrors"
--- a/src/ModelerApp.vue
+++ b/src/ModelerApp.vue
@@ -9,19 +9,19 @@
 
         <div class="ml-auto">
           <b-btn variant="secondary" size="sm" v-b-modal="'uploadmodal'" class="mr-2">
-            <i class="fas fa-upload mr-1" />
+            <i class="fas fa-upload mr-1"/>
             {{ $t('Upload XML') }}
-          </b-btn>
-          <b-btn variant="secondary" size="sm" data-test="downloadXMLBtn" @click="download">
-            <i class="fas fa-download mr-1" />
-            {{ $t('Download XML') }}
           </b-btn>
         </div>
       </b-card-header>
       <b-card-body class="overflow-hidden position-relative p-0 vh-100">
-        <modeler ref="modeler" @validate="validationErrors = $event" @warnings="warnings = $event" />
+        <modeler @set-xml-manager="xmlManager = $event" @validate="validationErrors = $event" @warnings="warnings = $event"/>
       </b-card-body>
-      <validation-status :validation-errors="validationErrors" :warnings="warnings" />
+      <validation-status
+        :validation-errors="validationErrors"
+        :warnings="warnings"
+        :xml-manager="xmlManager"
+      />
     </b-card>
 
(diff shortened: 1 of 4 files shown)


src/components/validationStatus/ValidationStatus.vue
```diff
@@ -1,83 +1,83 @@
...
+      class="mr-auto"
+      :disabled="!xmlManager"
+      variant="secondary"
+      size="sm"
+      data-test="downloadXMLBtn"
+      @click="xmlManager && xmlManager.download()"
+    >
+      <i class="fas fa-download mr-1"/>
+      {{ $t('Download BPMN') }}
+    </b-btn>
```
chipit24: Most of the changes in this template are whitespace changes (I removed a couple of `div`s that increased nesting but provided no real value). If you're reading this on GitHub, you can hide the whitespace changes to make it easier to review:

Screen Shot 2020-02-19 at 1 16 11 PM

I've commented on the important bit that's changed: adding the new download button.

src/ModelerApp.vue
```diff
@@ -9,19 +9,19 @@
...
           </b-btn>
-          <b-btn variant="secondary" size="sm" data-test="downloadXMLBtn" @click="download">
-            <i class="fas fa-download mr-1" />
-            {{ $t('Download XML') }}
-          </b-btn>
         </div>
       </b-card-header>
       <b-card-body class="overflow-hidden position-relative p-0 vh-100">
-        <modeler ref="modeler" @validate="validationErrors = $event" @warnings="warnings = $event" />
+        <modeler @set-xml-manager="xmlManager = $event" @validate="validationErrors = $event" @warnings="warnings = $event"/>
```
chipit24: This is a pattern I've used before in react: it gets around having to use a ref to reach into the component's methods by emitting a "controller" object that gives the parent control over things in the child component.

src/components/modeler/XMLManager.js
```diff
@@ -0,0 +1,29 @@
...
+          window.xml = xml;
+          return;
+        }
+
+        const file = new File([xml], 'bpmnProcess.xml', { type: 'text/xml' });
+        FilerSaver.saveAs(file);
+      }
+    });
+  }
+}
```
chipit24: This class can be added to, but I've kept it minimal for now.

johnblackspear: Like that you've extracted this

src/components/modeler/XMLManager.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  getDefinitionsFromXml(xmlString) {
+    return new Promise((resolve, reject) => {
+      this.#moddle.fromXML(xmlString, (err, definitions) => {
+        if (err) {
+          reject(err);
+        }
+
+        definitions.exporter = 'ProcessMaker Modeler';
```
johnblackspear: I'm wondering if we should move these types of strings to a central place (thing of my abuse of the constants file in the `e2e` directory as well). I know the same string is in `ModelerApp.vue`, for instance.

Also, should this string be passed through `$t()` here or is this always going to be an  untranslated exporter value?

chipit24: If `definitions.exporter` is suppose to be the name of the app, and be kept in sync with the page title in `ModelerApp.vue`, then yes, I think it makes sense to pull out the string and keep it in a single place.

However, core currently uses a different `ModelerApp.vue` file and has no reference to a "ProcessMaker Modeler" string. I'm also not sure how this string is used or if it should be translated. I'll clarify with Alan, but I think it's fine keeping it the way it is for now.

johnblackspear: 👍

chipit24: Looking at the spec it's support to identify the name and version of the app that produced the XML:

Screen Shot 2020-02-20 at 2 15 26 PM

So I don't think these values should be translated, but it might be worth pulling them from our `package.json` file to help with debugging. That way, if we come across an issue and have the XML, we can see something like `exporter="@processmaker/modeler" exporterVersion="0.18.4"`.

@velkymx thoughts on this?

PR #1126, opened 2020-02-19
Implements customizable status bar and elements highlight

### LINKED ISSUE #1117: Process Element Decorations and Custom Status Bar
# Process Element Decorations and Custom Status Bar
## Description
Add support to display Process completion check:

- Show flow paths
- Highlight elements

Custom status bar:
image

caleeli:
Resolves #1117 

This PR includes:

- Add Modeler Property **decorations**, to customize presentation of elements in Designer
- Add **registerStatusBar** to add custom elements to validation status bar

image


src/ModelerApp.vue
```diff
@@ -19,9 +19,11 @@
         </div>
       </b-card-header>
       <b-card-body class="overflow-hidden position-relative p-0 vh-100">
-        <modeler ref="modeler" @validate="validationErrors = $event" @warnings="warnings = $event" />
+        <modeler ref="modeler" @validate="validationErrors = $event" @warnings="warnings = $event" :decorations="decorations" />
       </b-card-body>
-      <validation-status :validation-errors="validationErrors" :warnings="warnings" />
+      <validation-status ref="validationStatus" :validation-errors="validationErrors" :warnings="warnings" :owner="self">
+        <component v-for="(component, index) in validationBar" :key="`validation-status-${index}`" :is="component" :owner="self" />
```
chipit24: I think `index` is sufficient for the `key` property: `:key="index"`.

I would highly recommend removing the `owner` prop. Is there a specific reason you did this instead of the regular props / events flow? I think it's a very hacky way to step around the Vue component API to enable two-way data binding, which is exactly what Vue's component based design aims to avoid as it's very hard to reason about and prone to bugs. The `owner` prop seems to be used how `$parent` would be used, but is more obscure.

In this case, I don't see any problems with passing down a `validationBar` prop, and listening for something like an `@update-validation-bar` event to update the `validationBar` data.

src/mixins/highlightConfig.js
```diff
@@ -69,6 +77,11 @@ export default {
         this.shapeView.highlight(this.shapeBody, defaultHighlighter);
       }
     },
+    setBorderOutline(borderOutline) {
+      this.currentBorderOutline ? this.shapeView.unhighlight(this.shapeBody, this.currentBorderOutline) : null;
+      this.currentBorderOutline = borderOutline ? JSON.parse(JSON.stringify(borderOutline)) : null;
+      this.currentBorderOutline ? this.shapeView.highlight(this.shapeBody, this.currentBorderOutline) : null;
+    },
```
chipit24: I find this a little hard to read. Can you simply it? You might be able to take advantage of an early return, and replace two of the ternary operators with actual `if` statements.

src/components/validationStatus/ValidationStatus.vue
```diff
@@ -86,7 +90,7 @@ export default {
   components: {
     FontAwesomeIcon,
   },
-  props: ['validationErrors', 'warnings'],
+  props: ['validationErrors', 'warnings', 'owner'],
```
chipit24: The `owner` prop doesn't appear to be used in this component; did you mean to pass it to modeler instead?

src/ModelerApp.vue
caleeli: removed

src/mixins/highlightConfig.js
caleeli: Improved

src/components/validationStatus/ValidationStatus.vue
caleeli: removed

src/ModelerApp.vue
```diff
@@ -61,6 +63,11 @@ export default {
   },
   data() {
     return {
+      self: this,
```
chipit24, line 66: It doesn't look like `self` is used, in which case I think it should be removed.

src/ModelerApp.vue
```diff
@@ -61,6 +63,11 @@ export default {
...
+      validationBar: [],
+      decorations: {
+        borderOutline: {},
+      },
```
chipit24, line 70: It seems like `decorations` only contains `borderOutline`, in which case, why wrap that property? Can `borderOutline` be referenced and passed down directly?

src/ModelerApp.vue
```diff
@@ -15,13 +15,15 @@
...
+        <modeler ref="modeler" @set-xml-manager="xmlManager = $event" @validate="validationErrors = $event" @warnings="warnings = $event" :decorations="decorations" />
       </b-card-body>
-      <validation-status
+      <validation-status ref="validationStatus"
         :validation-errors="validationErrors"
         :warnings="warnings"
         :xml-manager="xmlManager"
-      />
+      >
+        <component v-for="(component, index) in validationBar" :key="`validation-status-${index}`" :is="component" />
```
chipit24, line 25: I left [a comment](https://github.com/ProcessMaker/modeler/pull/1126#discussion_r382108133) about the key, but looks like it was unchanged. Does simplifying it cause errors?

> I think `index` is sufficient for the `key` property: `:key="index"`.

```suggestion
        <component v-for="(component, index) in validationBar" :key="index" :is="component" />
```

src/mixins/highlightConfig.js
```diff
@@ -69,6 +77,15 @@ export default {
         this.shapeView.highlight(this.shapeBody, defaultHighlighter);
       }
     },
+    setBorderOutline(borderOutline) {
+      if (this.currentBorderOutline) {
+        this.shapeView.unhighlight(this.shapeBody, this.currentBorderOutline);
+      }
+      this.currentBorderOutline = borderOutline ? JSON.parse(JSON.stringify(borderOutline)) : null;
```
chipit24, line 84: `JSON.parse(JSON.stringify(someObject))` is used to create a clone, but in this case, is it necessary? If the prop isn't being mutated, but a new border object is passed down, you can take advantage of the 2nd watcher param (previous value) that is passed to the watcher.

If not, since we use lodash, I think it would be more readable to use lodash's `clone` or `deepClone` method instead.

src/ModelerApp.vue
caleeli: Well the idea for use "decorations" is to group any other styling decorations of the diagram (outside of the diagram bpmn definitions). I know at the moment this PR only provides borderOutline, but I think in the future, other decorations could be required, like:

- change task colors (in runtime)
- opacity (to show elements that were disabled or failed)
- add icons (to show that some tasks have uploaded files or comments)

src/mixins/highlightConfig.js
caleeli: Yes it is necessary because if the caller mutates the object dynamically (changing color or any other property) the unhighlight function will not remove the previous border highlight. Then this way if the caller changes any inner property of the border the style of the highlight will be changed safely.
Changed to `cloneDeep`

caleeli:
@chipit24 thanks so much for your comments

PR #1127, opened 2020-02-19
Add weekday configuration to Intermediate Timer Event

### LINKED ISSUE #1106: Can not attach intermediate events as boundary
# Can not attach intermediate events as boundary
## Description
In the current develop version, I can not attach intermediate events as boundary

## Reproduction steps
- Add a Task (normal user task)
- Drag an intermediate (timer or message) and try to attach it to the task

## Expected behavior
The user should be able to attach boundary events to the tasks

## Actual behavior
Can not attach intermediate events as boundary

image
image

## Other information
Current develop commit: f4f26bc1905e5cde2470c47811b22e44f5544304

### LINKED ISSUE #1107: Missing day of week parameter to configure the Intermediate Timer Event
# Missing day of week parameter to configure the Intermediate Timer Event  
## Description
The Cyclic Intermediate Timer Event has missing parameters

## Reproduction steps
Add a Intermediate Timer Event
- Go to the Timing control properties
- Select Type=Cycle
- Select week

## Expected behavior
It must have a property to select the week day
image

## Actual behavior
Missing week day selection in weekly cycles
image

lisa-fehr:
Closes #1107

Larger than expected :(


lisa-fehr:
Adding a test. But ready for review.

src/components/inspectors/TimerExpression.vue
```diff
@@ -327,31 +245,12 @@ export default {
       this.periodicity = 'week';
       this.endDate = DateTime.local().toUTC().toISO();
       this.ends = 'never';
-      this.weekdays.forEach(weekday => weekday.selected = false);
+      this.expression = null;
+      this.selectedWeekdays = [];
     },
     clickWeekDay(weekday) {
       weekday.selected = !weekday.selected;
     },
```
chipit24: This method doesn't appear to be used here anymore.

src/components/inspectors/WeekdaySelect.vue
```diff
@@ -0,0 +1,185 @@
...
+
+export default {
+  props: {
+    selectWeekdays: {
+      type: Array,
+      default: () => [],
+    },
+    startDate: {
+      type: String,
+      default: DateTime.local().toUTC().toISO().toString(),
```
chipit24: [`toISO`](https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html#instance-method-toISO) already returns a string so I think calling `toString` on it is redundant.
```suggestion
      default: DateTime.local().toUTC().toISO(),
```

src/components/inspectors/CycleExpression.vue
```diff
@@ -68,10 +83,25 @@ export default {
       if (this.periodicity.isTime) {
         return `R/PT${this.repeat}${this.periodicity.value}`;
       }
+      if (this.periodicity.isWeek && this.expression) {
+        return this.expression;
+      }
       return `R/P${this.repeat}${this.periodicity.value}`;
```
chipit24: The `cycleExpression` computed property only appears to be used for the watcher, and PHPStorm complains about this function being unused. But I think keeping the computed property makes this easier to read.

src/components/inspectors/CycleManager.js
```diff
@@ -0,0 +1,100 @@
...
+import { getIso8601FormattedDateString, getPeriod } from './TimeUtils';
+
+export default class CycleManager {
+  _startDate;
+  _repeat;
+  _periodicityValue;
+  _selectedWeekdays;
+  _endDate;
+  _ends;
+  _times;
```
chipit24: Instead of using `_` to denote private properties, we've used `#` in other classes (which will turn it into a "real" private property). Do you think that'll help here?

src/components/inspectors/WeekdaySelect.vue
```diff
@@ -0,0 +1,185 @@
...
+        {
+          day: 6,
+          initial: 'S',
+          selected: false,
+        },
+      ],
+      cycleManager: new CycleManager(this.startDate, this.repeat, this.periodicityValue, this.selectedWeekdays, this.endDate, this.ends, this.times),
+    };
+  },
+  watch: {
```
chipit24: It seems some of these props/watchers are set up to mirror the value on the cycle manager class. Do you think managing this class from the parent, and passing down the class instead of the individual props is possible / will help?

src/components/inspectors/TimerExpression.vue
lisa-fehr: Weird. It didn't grey out the method for me in phpstorm.

src/components/inspectors/WeekdaySelect.vue
lisa-fehr: I had trouble trying to pass cycleManager in as a prop. If that is what you were meaning. It often kept old copies of the values. Would need to mob on it.

src/components/inspectors/CycleManager.js
```diff
@@ -2,69 +2,69 @@ import { DateTime } from 'luxon';
...
   // istanbul ignore next
   set selectedWeekdays(value) {
-    this._selectedWeekdays = value;
+    this.#selectedWeekdays = value;
   }
 
   /**
    * For historic reasons we need to be liberal in how we accept various incoming
-   * dates. See https://en.wikipedia.org/wiki/Robustness_principle
+   * dates. See https://en.wikipedia.org/wiki/Robustness#principle
```
chipit24: Was this change intentional? Generally, hashes / fragments like this will scroll the page to the element with that ID (`principle` in this case). But the `/wiki/Robustness` page has no such element. Meanwhile, the `/wiki/Robustness_principle` page is more relevant.

src/components/inspectors/WeekdaySelect.vue
chipit24: Yeah, that's what I meant. But now that I'm thinking about it, it's not as clear cut as I thought. Changing a value on an instance of that class won't trigger any of Vue's reactivity hooks, and that's good, since it sidesteps the whole unidirectional flow of data between components and should be avoided anyway. That may be what you were experiencing.

To make it work, you'd have to create and pass down a new instance of `CycleManager` each time, which also doesn't sound right. I'm wondering if pulling back a bit and not using a class might be a better approach? Maybe just having a collection of utility functions? Not too sure, but right now this works and what I'm thinking of would require an unknown amount of extra work so let's get this merged.

src/components/inspectors/CycleManager.js
```diff
@@ -5,17 +5,17 @@ export default class CycleManager {
   #startDate;
   #repeat;
   #periodicityValue;
-  #selectedWeekdays;
-  #endDate;
+  _selectedWeekdays;
+  _endDate;
```
chipit24: @lisa-fehr Were these two properties causing tests to fail? I think having some properties use the underscore and others the hash makes it a little harder for me to read compared to having them all follow the same convention.

If there's issues with the private class property syntax then sticking with underscores is fine.

src/components/inspectors/CycleManager.js
johnblackspear: Nah I don't think that is intentional nor do I think we really need the comment, this was a placeholder for a discussion at some point :)

src/components/inspectors/CycleManager.js
lisa-fehr: Yeah. It is a bit strange. I can't see a direct alteration of the private value.

PR #1128, opened 2020-02-20
Fix BPMN error console delay

### LINKED ISSUE #1125: Delay in the BPMN Error Console
# Bug report issue
## Description
BPMN validation console has a delayed response when updating inspector configs such as the required 'Name' field.
## Reproduction steps
1. Toggle the 'Auto Validate' to on and open the error console.
2. Drag a Task onto the paper, notice the BPMN error count was immediately updated.
3. Remove the task name within the inspector 'Name' input to produce a BPMN validation error. Notice the 'Name' input field displays an error but the BPMN error console does not. 
3. Click off the inspector. Notice the BPMN error count is now updated to include the 'Name' input errors. 
4. Click back on the node to open the inspector, add a name to the 'Name' field. Notice the the 'Name' input field errors disappear, but the error in the BPMN error console does not.
5. Click off the inspector. Notice the error is no longer displayed in the BPMN console and the BPMN error count is now updated.

## Expected behavior
Updating any required inspector configs that throw a BPMN error, should update the BPMN console in realtime, without having to click off the inspector.
## Actual behavior
When configuring required fields, you need to click off the inspector to reflect updated BPMN errors in the BPMN console.

## Other information
<a href="https://www.loom.com/share/8016e3457131486483cd04a96a2b28eb"> <p>19 February, 2020 - Loom Recording - Watch Video</p>  </a>

sanjacornelius:
<h2>Changes</h2>
Emit an 'save-state' event when there are updates within the Form Renderer.

<h2>To Test</h2>

1. Toggle the 'Auto Validate' to on and open the error console.
2. Drag a Task onto the paper.
3. Remove the task name within the inspector 'Name' input to produce a BPMN validation error. 

**Expected Behavior**
Notice the 'Name' input field displays and the BPMN error count updates.

4. Add a name to the 'Name' field. 

**Expected Behavior**
Notice the the 'Name' input field errors disappear and the BPMN error count updates.

<a href="https://www.loom.com/share/d8d3bea34e474bb7a5c327b75741961a"> <p>Google Chrome - modeler - Watch Video</p>  </a>

Closes #1125

### DIFF
--- a/src/components/inspectors/InspectorPanel.vue
+++ b/src/components/inspectors/InspectorPanel.vue
@@ -173,6 +173,7 @@ export default {
     defaultInspectorHandler(value) {
       /* Go through each property and rebind it to our data */
       for (const key in omit(value, ['$type', 'eventDefinitions'])) {
+        this.updateState();
         if (this.highlightedNode.definition.get(key) !== value[key]) {
           this.setNodeProp(this.highlightedNode, key, value[key]);
         }


src/components/inspectors/InspectorPanel.vue
```diff
@@ -173,6 +173,7 @@ export default {
     defaultInspectorHandler(value) {
       /* Go through each property and rebind it to our data */
       for (const key in omit(value, ['$type', 'eventDefinitions'])) {
+        this.updateState();
```
sanjacornelius: @chipit24  While digging into the code saw that the 'save-state' emits when focusing out of the inspector.  Using that same method, I added it to this function whenever the form renderer is updated. Not sure if this is the best place to input this method, but that is the idea I had behind this.

chipit24: This is a great catch! But, I'm hesitant to have this merged without clarifying a few things:

Updating on every keystroke saves a new BPMN string to the undo/redo stack, so when you press "undo" after entering a name, it will undo the name you entered letter-by-letter. This was a choice we consciously made, as we wanted updates like this to be batched. Has this requirement changed? If not, we'll have to come up with a different solution to this problem. Maybe only make the additional call to `updateState` when the validation status of the form changes?

Also, where you have the `updateState` function call right now means it will only be executed for the default case, but it should be called for custom inspector handlers as well.

sanjacornelius: The requirements have not changed, so another solution is needed. I'll look into your suggestion for adding the `updateState` when the validation status of the form changes. Thank you!

sanjacornelius:
Closing. A fix for this is not needed, leaving current functionality as-is.

PR #1131, opened 2020-02-21
Double click edit label

### LINKED ISSUE #1118: Canvas Edit Labels
As a process designer, I'd like to be able double click an element and edit the label on the canvas.

chipit24:
Fixes #1118.

dbl-click-label

### DIFF
--- a/src/components/modeler/Modeler.vue
+++ b/src/components/modeler/Modeler.vue
@@ -704,7 +704,7 @@ export default {
       return shape.component != null;
     },
     setShapeStacking(shape) {
-      if (this.isRendering || shape === this.previouslyStackedShape) {
+      if (this.isRendering || (!shape.component.node.isType('processmaker-modeler-pool') && shape === this.previouslyStackedShape)) {
         return;
       }
 
--- a/src/components/modeler/Modeler.vue
+++ b/src/components/modeler/Modeler.vue
@@ -131,7 +131,7 @@ import Node from '@/components/nodes/node';
 import { addNodeToProcess } from '@/components/nodeManager';
 import moveShapeByKeypress from '@/components/modeler/moveWithArrowKeys';
 import setUpSelectionBox from '@/components/modeler/setUpSelectionBox';
-import focusNameInput from '@/components/modeler/focusOnDoubleClick';
+import focusNameInputAndHighlightLabel from '@/components/modeler/focusNameInputAndHighlightLabel';
 import XMLManager from '@/components/modeler/XMLManager';
 
 export default {
@@ -756,7 +756,7 @@ export default {
     this.paperManager = PaperManager.factory(this.$refs.paper, this.graph.get('interactiveFunc'), this.graph);
     this.paper = this.paperManager.paper;
 
-    this.paperManager.addEventHandler('cell:pointerdblclick', focusNameInput);
+    this.paperManager.addEventHandler('cell:pointerdblclick', focusNameInputAndHighlightLabel);
 
     this.handleResize();
     window.addEventListener('resize', this.handleResize);
--- a/src/components/modeler/Modeler.vue
+++ b/src/components/modeler/Modeler.vue
@@ -131,6 +131,7 @@ import Node from '@/components/nodes/node';
 import { addNodeToProcess } from '@/components/nodeManager';
 import moveShapeByKeypress from '@/components/modeler/moveWithArrowKeys';
 import setUpSelectionBox from '@/components/modeler/setUpSelectionBox';
+import focusNameInput from '@/components/modeler/focusOnDoubleClick';
 
 const version = '1.0';
 
@@ -181,6 +182,7 @@ export default {
       nodeTypes: [],
       breadcrumbData: [],
       activeNode: null,
+      previouslyStackedShape: null,
     };
   },
   watch: {
@@ -708,10 +710,11 @@ export default {
       return shape.component != null;
     },
     setShapeStacking(shape) {
-      if (this.isRendering) {
+      if (this.isRendering || shape === this.previouslyStackedShape) {
         return;
       }
 
+      this.previouslyStackedShape = shape;
       this.paperManager.performAtomicAction(() => ensureShapeIsNotCovered(shape, this.graph));
     },
   },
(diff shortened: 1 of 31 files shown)


src/mixins/focusOnDoubleClick.js
```diff
@@ -0,0 +1,48 @@
+let timeoutID;
+
+function waitToTriggerOpenAnimation() {
+  return new Promise(resolve => {
+    timeoutID = setTimeout(resolve);
+  });
+}
+
+async function focusNameInput(cellView) {
+  if (cellView.model !== this.shape) {
```
lisa-fehr: Works great when I just drag on a few elements, but when I connect to a sequence flow some don't allow doubleclick on the labels anymore. If I add a Text Annotation or End Event after connecting a sequence flow, they also don't seem to allow doubleclick.
doubleclick-lables


It'll say:
`cellView.model` is `processmaker.components.nodes.task.Shape` and `this.shape` is `standard.Link`
Or `focusNameInput` stops firing for some elements.

lisa-fehr: If I delete the sequence flow, I can doubleclick the other labels again.

src/mixins/focusOnDoubleClick.js
```diff
@@ -0,0 +1,48 @@
...
+    clearTimeout(timeoutID);
+    configurationAccordion.click();
+    await waitToTriggerOpenAnimation();
+  }
+
+  const labelElement = this.shapeView.selectors.label ||
+    Array.from(this.shapeView.selectors.root.children).find(el => el.classList.contains('labels'));
+  labelElement.style.outline = '1px dashed blue';
+  nameInput.addEventListener('blur', () => {
+    labelElement.style.outline = '';
```
lisa-fehr: This seems to error when doubleclicking a message flow.

src/mixins/focusOnDoubleClick.js
```diff
@@ -0,0 +1,48 @@
...
+  const configurationAccordion = document.getElementById('accordion-button-Configuration');
+  if (configurationAccordion.getAttribute('aria-expanded') === 'false') {
+    clearTimeout(timeoutID);
+    configurationAccordion.click();
+    await waitToTriggerOpenAnimation();
+  }
+
+  const labelElement = this.shapeView.selectors.label ||
+    Array.from(this.shapeView.selectors.root.children).find(el => el.classList.contains('labels'));
+  labelElement.style.outline = '1px dashed blue';
```
lisa-fehr: This seems to error when doubleclicking a message flow.

chipit24: Ah, I get this error:
```
focusOnDoubleClick.js Uncaught (in promise) TypeError: Cannot read property 'style' of undefined
```

It seems message flows don't have labels, so I'll return if `labelElement` is `undefined`.

src/mixins/focusOnDoubleClick.js
chipit24: I can't seem to reproduce this, even following the exact steps you took in your gif. Do you have a list of steps that I can follow to ensure this error shows up?

johnblackspear: I can also reproduce this:
 - drag a task (A) to the diagram
 - drag another task (B) to the diagram
 - connect A to B with a sequence flow

I can now double click to edit A but not B.

Deleting the sequence flow makes B editable via double click again.

chipit24: Ah, I can finally reproduce this, but only in Safari (v13); it works without issue for me in Chrome 80 and FF 72 and 73.

@lisa-fehr @johnblackspear what browsers / browser versions did you use to reproduce this?

johnblackspear:
I've been looking into this - seems that the sequence flow has a double click handler so that you can add control points, e.g.:




I *think* this is interfering somehow with the double click handler for label editing, looking into that now.

johnblackspear:
OK so it's very probable that it's the sequence flow doubleclick handler that is causing this. Changing the label edit trigger event to something else (say, `contextmenu`) resolves this.

So I'll be focusing on how to resolve the conflict between the various double click handlers.

src/mixins/focusOnDoubleClick.js
lisa-fehr: @chipit24 I saw it in Chrome Version 80.0.3987.122

chipit24:
I created a bare-bones example in CodeSandbox of two tasks connected with a link: https://codesandbox.io/s/cocky-surf-lbs76.

This example seems to work correctly in Safari, but for modeler, only the click handler is fired for the sequence flow target element (and the double click handler is ignored).

This doesn't confirm a whole lot, but there is something specific to modeler that's causing double click events to be ignored for link targets (only in Safari for me).

src/components/modeler/Modeler.vue
```diff
@@ -708,10 +710,11 @@ export default {
       return shape.component != null;
     },
     setShapeStacking(shape) {
-      if (this.isRendering) {
+      if (this.isRendering || shape === this.previouslyStackedShape) {
```
chipit24: If this is called multiple times for the same shape, due to a bug (possibly related to https://github.com/clientIO/joint/issues/1263), the double click event is not fired. This should also prevent multiple unnecessary calls to `toFront`, though I'm not sure if that had any performance implications.

johnblackspear: Good catch, Rob

chipit24: It seems `setShapeStacking` needs to be called more than once when dealing with lanes, so I had to add an extra check for that.

I posted an update to the issue I linked above; it seems multiple calls to `toFront` does indeed interfere with the emitting of the double click event.

chipit24:
The failing test on this branch (`UndoRedo.spec.js: "Correctly parses elements after redo"`) is passing for me locally both when running through the GUI and in headless mode. I'll re-run the tests and see if there's any difference, otherwise this'll be a pain to debug.

chipit24:
It looks like this is failing because on CI, for some reason, the start event doesn't get added for this test and it causes the IDs to be off by 1 in the XML. I'll look into why that is, but if all else fails I'll see if I can have it ignore the IDs when comparing the XML.

PR #1138, opened 2020-03-11
Change task to form task

### LINKED ISSUE #1135: Change Task to Form Task
Change the name of the task in the crown from Task to Form Task.

chipit24:
Fixes #1135.

Screen Shot 2020-03-11 at 10 09 00 AM


lisa-fehr:
Would the side menu still be **Task**, since it covers all task types?
formTaskQuestion

chipit24:
@lisa-fehr During standup yesterday I asked Alan and he said to re-name it to Form Task even in the controls list, if I remember correctly. But I think it makes sense to keep that as Task. I can double check during standup.

PR #1144, opened 2020-03-18
Sub Process name fixes

### LINKED ISSUE #1142: Sub Process Name is Difficult
This might already be fixed, but the naming in the inspector of a sub process is uncooperative. You can't type in the field.

### LINKED ISSUE #1143: Do not replace sub process name with selected process
If you type/paste in a name for a subprocess, it should not get overwritten if you select a process.
Saved subprocess names should not be overwritten after selecting a process as well.

Related #1142

lisa-fehr:
Do not allow the sub process name to be set as undefined
Closes #1142 

Do not change sub process name when selecting a process
Closes #1143 

Changes the behavior of this pr: https://github.com/ProcessMaker/modeler/pull/1052

### DIFF
--- a/src/components/nodes/subProcess/index.js
+++ b/src/components/nodes/subProcess/index.js
@@ -39,8 +39,8 @@ export default {
 
     setNodeProp(node, 'calledElement', currentConfig.calledElement);
 
-    if (currentConfig.name != value.name) {
-      if (oldConfig.name === value.name || !oldConfig.name) {
+    if (currentConfig.name !== value.name) {
+      if (currentConfig.name && (oldConfig.name === value.name || !oldConfig.name)) {
         // SubProcessFormSelect automatically updated the name so set the new name here
         setNodeProp(node, 'name', currentConfig.name);
       } else {
--- a/src/components/nodes/subProcess/index.js
+++ b/src/components/nodes/subProcess/index.js
@@ -6,6 +6,11 @@ import nameConfigSettings from '@/components/inspectors/nameConfigSettings';
 export const taskHeight = 76;
 export const id = 'processmaker-modeler-call-activity';
 
+function subProcessFormSelected(currentName, node) {
+  const previousConfigName = JSON.parse(node.definition.config).name;
+  return previousConfigName === currentName || !previousConfigName;
+}
+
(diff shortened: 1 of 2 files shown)


src/components/nodes/subProcess/index.js
```diff
@@ -39,8 +39,8 @@ export default {
 
     setNodeProp(node, 'calledElement', currentConfig.calledElement);
 
-    if (currentConfig.name != value.name) {
-      if (oldConfig.name === value.name || !oldConfig.name) {
+    if (currentConfig.name !== value.name) {
+      if (currentConfig.name && (oldConfig.name === value.name || !oldConfig.name)) {
```
chipit24: I find this block a little hard to read. I wonder if it can be pulled out into a function and then refactored.

lisa-fehr: I wasn't 100% sure what it is doing to give it a name.

johnblackspear: Random name suggestions, feel free to ignore:

`currentConfig.name !== value.name` ==> `nameChanged`

`currentConfig.name && (oldConfig.name === value.name || !oldConfig.name)` ==> `nameChangedByFormSelect`

Or something like that, not sure my suggestion is an improvement :)

src/components/nodes/subProcess/index.js
```diff
@@ -34,14 +39,12 @@ export default {
...
 
-    const oldConfig = JSON.parse(node.definition.config);
     const currentConfig = JSON.parse(value.config);
 
     setNodeProp(node, 'calledElement', currentConfig.calledElement);
 
     if (currentConfig.name !== value.name) {
-      if (currentConfig.name && (oldConfig.name === value.name || !oldConfig.name)) {
-        // SubProcessFormSelect automatically updated the name so set the new name here
+      if (subProcessFormSelected(value.name, node) && currentConfig.name) {
```
chipit24: I don't know if this helps with readability for me. I like what @johnblackspear suggested, taking this whole line and moving it into a function with a name like "nameChangedByFormSelect". The current name, for me, suggests that some form in a sub process was selected.

PR #1145, opened 2020-03-18
Show inspector when element is highlighted

### LINKED ISSUE #1139: Show Inspector on Focus
When the menus are hidden and I click on an element I'd like the inspector to slide out so I can use it. Then when I click off of the element (on the canvas) the menu should slide away.

chipit24:
Fixes #1139.

Video of feature: https://www.dropbox.com/s/udxbskut6r0y0jk/Mar-18-2020%2015-12-20__highlighting.mp4?dl=0.


tests/e2e/specs/Modeler.spec.js
```diff
@@ -568,4 +568,16 @@ describe('Modeler', () => {
           });
       });
   });
+
+  it('after collapsing panels, show inspector handel when element is highlighted', () => {
```
lisa-fehr:
```suggestion
  it('after collapsing panels, show inspector handle when element is highlighted', () => {
```

chipit24: Whoops, should be "inspector panel". Good catch!

PR #1147, opened 2020-03-19
Add color palette to crown

### LINKED ISSUE #1146: Colorize Elements
image

The color in the bubble is the BORDER / ICON color. The interior color should be 20-30% lighter (maybe more).

Example:

image

Clicking on the rainbow icon will show the color picker (there is already one in the core product)

image

Color information should be store as a custom PM4 attribute for the element

Colors that are defined in other elements in the current diagram will show in the recents palets here: 

image

chipit24:
Fixes #1146.

Remaining todos:
- [x] Reset node to previous colour instead of default colour after changing it to the valid (green) or invalid (red) node colour.
- [x] Make lanes added to pools the same colour as the parent pool.
- [x] Persist sequence flow colour (it currently resets after deselecting flow).
- [x] Fix pool and lane colour reset.
   * Change the pool colour, then click to clear the pool colour (the white circle). It doesn't clear.
- [x] Add tests.

Video of feature: https://www.dropbox.com/s/99z3at8jfdg6rs4/color_picker.mp4?dl=0

**Note:** In core, a warning is generated about the color attribute; this doesn't seem to cause any breaking changes, and will need to be fixed in https://github.com/ProcessMaker/processmaker-bpmn-moddle.


src/components/nodeColors.js
```diff
@@ -1,6 +1,72 @@
...
+export const intermediateColorStroke = '#fbbe02';
+
 export const endColor = '#fff1f2';
+export const endColorStroke = '#ed4757';
+
 export const poolColor = '#f7f7f7';
+
+export function getDefaultNodeColors(node, color) {
+  if (color) {
+    return { fill: Color(color).lighten(0.3).hex(), stroke: color };
```
johnblackspear: This is great - I've noticed that the contrast between text and the grey is not super high, so if you want consider upping the lightening to something like 40% or so

PR #1159, opened 2020-03-24
Fix crown hiding bug

### LINKED ISSUE #1158: Hide crown after pressing Sequence Flow button
## Description
The crown does not automatically close after selecting a crown action.

## Reproduction steps
1. Drag any element (e.g. a Start Event) onto the grid.
2. Highlight the element and select the Sequence Flow button in the crown.

## Expected behaviour
The crown should disappear after selecting the Sequence Flow button.

## Actual behaviour
The crown remains open.

## Software used
Latest standalone modeler `develop` branch, on Chrome 80 and macOS Catalina.

## Other information
It appears this issue was introduced in https://github.com/ProcessMaker/modeler/commit/3e457aa1c8c477384b375f066b993d19a2aa14e.

chipit24:
Fixes #1158.

This issue was due to the fact that adding a sequence flow caused the `highlightedShapes` array to change. Normally, clicking to add a sequence flow executes `@toggle-crown-state="showCrown = $event"` to hide the crown, but right after that, the sequence flow is added, and the watcher for `highlightedShapes` is run, causing `this.showCrown` to be set back to `true`.

### DIFF
--- a/src/components/crown/crownConfig/crownConfig.vue
+++ b/src/components/crown/crownConfig/crownConfig.vue
@@ -70,6 +70,7 @@ import poolLaneCrownConfig from '@/mixins/poolLaneCrownConfig';
 import { removeFlows } from '@/components/crown/utils.js';
 import pull from 'lodash/pull';
 import store from '@/store';
+import isEqual from 'lodash/isEqual';
 
 export default {
   components: {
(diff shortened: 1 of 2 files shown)


src/components/crown/crownConfig/crownConfig.vue
```diff
@@ -102,7 +103,11 @@ export default {
   },
   mixins: [poolLaneCrownConfig],
   watch: {
-    highlightedShapes() {
+    highlightedShapes(shapes, prevShapes) {
```
chipit24: For how this works, see https://vuejs.org/v2/api/#vm-watch. There is one catch:
> Note: when mutating (rather than replacing) an Object or an Array, the old value will be the same as new value because they reference the same Object/Array. Vue doesn’t keep a copy of the pre-mutate value.

Since we create a new array each time `highlightedShapes` changes, this works as expected.

PR #1160, opened 2020-03-25
Keep element name when switching type

### LINKED ISSUE #1150: Keep element name when changing type
When you change the element type via the "cog dropdown", the element should retain the same name.

lisa-fehr:
Closes #1150 

Should only switch the name when changing type if it matches a default name for that type.

### DIFF
--- a/src/components/nodes/eventBasedGateway/index.js
+++ b/src/components/nodes/eventBasedGateway/index.js
@@ -1,16 +1,19 @@
 import component from './eventBasedGateway.vue';
 import advancedAccordionConfig from '@/components/inspectors/advancedAccordionConfig';
+import defaultNames from '@/components/nodes/gateway/defaultNames';
+
+const id = 'processmaker-modeler-event-based-gateway';
 
 export default {
-  id: 'processmaker-modeler-event-based-gateway',
+  id,
   component,
   bpmnType: 'bpmn:EventBasedGateway',
   control: false,
   category: 'BPMN',
-  label: 'Event-Based Gateway',
+  label: defaultNames[id],
   definition(moddle, $t) {
     return moddle.create('bpmn:EventBasedGateway', {
-      name: $t('Event-Based Gateway'),
+      name: $t(defaultNames[id]),
     });
   },
   diagram(moddle) {
@@ -23,7 +26,7 @@ export default {
   },
   inspectorConfig: [
     {
-      name: 'Event-Based Gateway',
+      name: defaultNames[id],
       items: [
         {
           component: 'FormAccordion',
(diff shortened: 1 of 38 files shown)


src/components/nodes/eventBasedGateway/index.js
```diff
@@ -1,16 +1,19 @@
...
+const id = 'processmaker-modeler-event-based-gateway';
 
 export default {
-  id: 'processmaker-modeler-event-based-gateway',
+  id,
   component,
   bpmnType: 'bpmn:EventBasedGateway',
   control: false,
   category: 'BPMN',
-  label: 'Event-Based Gateway',
```
lisa-fehr: This one was either `Event-Based Gateway` or `Event Based Gateway`. I switched it to the crown item.

tests/unit/keepOriginalName.spec.js
```diff
@@ -0,0 +1,85 @@
...
+  });
+  it('Should switch default Gateway name', () => {
+    const node = new Node(
+      'processmaker-modeler-parallel-gateway',
+      { id: 'test', name: 'Exclusive Gateway' },
+      {},
+    );
+    expect(keepOriginalName(node)).false;
+  });
+});
```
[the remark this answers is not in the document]
lisa-fehr: Might take a bit of time to set this one up.

src/components/nodes/node.js
```diff
@@ -18,6 +26,26 @@ export default class Node {
     return this.type === type;
   }
 
+  isStartGroup() {
```
johnblackspear: Minor thing: I'm wondering if the concept of "group" will throw us later when we've forgotten the discussions around this. Would a clunkier but more specific name like `isStartEvent` or so work for you?

src/components/nodes/subProcess/index.js
```diff
@@ -13,10 +14,10 @@ export default {
   control: false,
   category: 'BPMN',
   icon: require('@/assets/toolpanel/subProcess.svg'),
-  label: 'Sub Process',
+  label: defaultNames[id],
```
johnblackspear: Just wondering - why is the label untranslated but the moddle name translated?

src/components/nodes/task/defaultNames.js
```diff
@@ -0,0 +1,8 @@
+const defaultNames = {
+  'processmaker-modeler-task': 'Form Task',
```
johnblackspear: :D +1, this one eh

src/components/nodes/subProcess/index.js
lisa-fehr: `$t` isn't available here. We would need another solution to translate them.

lisa-fehr:
Screen Shot 2020-03-30 at 2 52 47 PM
A solution for the edge case on 'Task with Marker' could be keeping the name if there are no default names.

PR #1163, opened 2020-03-27
Add SVG to Save event

johnblackspear:
This adds an SVG to the `saveBpmn` event.

It is necessary to save the SVG in core that is later used for the printable view.

There are some `!!url-loader!` calls that are used to embed icons in the SVG. These will likely be replaced later via the `updateIconColor` mixin that also encodes them using the data URI schema.

This is a requirement for:
https://github.com/ProcessMaker/processmaker/issues/2956 but does not address it. That is in a separate PR in Core.

### DIFF
--- a/src/ModelerApp.vue
+++ b/src/ModelerApp.vue
@@ -15,7 +15,7 @@
         </div>
       </b-card-header>
       <b-card-body class="overflow-hidden position-relative p-0 vh-100">
-        <modeler ref="modeler" @saveBpmn="saveBpmn" @set-xml-manager="xmlManager = $event" @validate="validationErrors = $event" @warnings="warnings = $event"/>
+        <modeler ref="modeler" @set-xml-manager="xmlManager = $event" @validate="validationErrors = $event" @warnings="warnings = $event"/>
       </b-card-body>
       <validation-status
         :validation-errors="validationErrors"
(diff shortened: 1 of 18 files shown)


src/ModelerApp.vue
```diff
@@ -15,7 +15,7 @@
         </div>
       </b-card-header>
       <b-card-body class="overflow-hidden position-relative p-0 vh-100">
-        <modeler ref="modeler" @set-xml-manager="xmlManager = $event" @validate="validationErrors = $event" @warnings="warnings = $event"/>
+        <modeler ref="modeler" @saveBpmn="saveBpmn" @set-xml-manager="xmlManager = $event" @validate="validationErrors = $event" @warnings="warnings = $event"/>
```
chipit24: Since you removed the function you should also remove the event listener that references it, as it's causing an error.
```suggestion
        <modeler ref="modeler" @set-xml-manager="xmlManager = $event" @validate="validationErrors = $event" @warnings="warnings = $event"/>
```

src/components/nodes/association/association.vue
```diff
@@ -19,7 +19,7 @@
 import { shapes } from 'jointjs';
 import linkConfig from '@/mixins/linkConfig';
 import get from 'lodash/get';
-import associationHead from '@/assets/association-head.svg';
```
johnblackspear: These are still needed until we decide on whether to expose these via the `updateIconColor` mixin

PR #1164, opened 2020-03-30
Update icon color when shape color changes

### LINKED ISSUE #1154: Change icon color when changing element color
Related to https://github.com/ProcessMaker/modeler/issues/1146.

Currently icons are pulled in as images with hard-coded colors. The color of these icons should change to the color selected for the element via the color picker crown dropdown.

chipit24:
Fixes #1154.

task_icon_changing_color

**Note:** There are two additional edge cases to consider which will require further investigation and work in separate issues:
 * Text annotation does not span the entire length of the text:
    Screen Shot 2020-03-30 at 4 26 25 PM
 * Arbitrary marker icons do not change color:
   Screen Shot 2020-03-30 at 4 33 30 PM

### DIFF
--- a/src/assets/intermediate-mail-alt.svg
+++ b/src/assets/intermediate-mail-alt.svg
@@ -1,6 +1,3 @@
 <svg width="30" height="18" viewBox="0 0 30 18" fill="none" xmlns="http://www.w3.org/2000/svg">
-<rect width="30" height="18" fill="#FABD2D"/>
-<path d="M29.5 0.5H0.5V17.5H29.5V0.5Z" stroke="#FFF4D3"/>
-<path d="M1.23334 1.26471L15 10" stroke="#FFF4D3" stroke-linecap="square"/>
-<path d="M28.7667 1.26471L15 10" stroke="#FFF4D3" stroke-linecap="square"/>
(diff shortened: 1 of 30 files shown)


src/assets/intermediate-mail-alt.svg
```diff
@@ -1,6 +1,3 @@
 <svg width="30" height="18" viewBox="0 0 30 18" fill="none" xmlns="http://www.w3.org/2000/svg">
-<rect width="30" height="18" fill="#FABD2D"/>
-<path d="M29.5 0.5H0.5V17.5H29.5V0.5Z" stroke="#FFF4D3"/>
-<path d="M1.23334 1.26471L15 10" stroke="#FFF4D3" stroke-linecap="square"/>
-<path d="M28.7667 1.26471L15 10" stroke="#FFF4D3" stroke-linecap="square"/>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M28.2506 1H1.7494L15 9.40784L28.2506 1ZM15 10.5922L15.1543 10.6901L15.3623 10.3623L29 1.70884V17H1V1.70881L14.6377 10.3623L14.8457 10.6901L15 10.5922Z" fill="#FABD2D"/>
```
chipit24: For this icon and the one below it (`message-end-event.svg`), I removed the outline as the stroke and fill would become the same color and end up rendering a solid rectangle.

I did this using Figma:
Screen Shot 2020-03-30 at 4 37 12 PM

johnblackspear: 👍

PR #1172, opened 2020-04-03
Fix sequence flow validation and crown button rendering

### LINKED ISSUE #1162: End Signal Event should not have a sequence flow crown option
Screen Shot 2020-03-26 at 2 51 18 PM
Screen Shot 2020-03-26 at 2 54 07 PM

chipit24:
Fixes #1162.

### DIFF
--- a/src/components/nodes/association/index.js
+++ b/src/components/nodes/association/index.js
@@ -11,8 +11,11 @@ export default {
   control: false,
   definition(moddle) {
     return moddle.create('bpmn:Association', {
-      associationDirection: `${ direction.none }`,
+      associationDirection: `${direction.none}`,
     });
+  },
+  allowOutgoingSequenceFlow() {
+    return false;
   },
   inspectorConfig: [
     {
--- a/src/components/nodes/sequenceFlow/sequenceFlow.vue
+++ b/src/components/nodes/sequenceFlow/sequenceFlow.vue
@@ -106,7 +106,7 @@ export default {
         this.targetIsNotALane() &&
         this.targetIsInSamePool() &&
         this.targetIsNotSource() &&
-        this.allowOutgoingFlow() &&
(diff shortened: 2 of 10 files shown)


src/components/nodes/association/index.js
```diff
@@ -11,9 +11,12 @@ export default {
   control: false,
   definition(moddle) {
     return moddle.create('bpmn:Association', {
-      associationDirection: `${ direction.none }`,
+      associationDirection: `${direction.none}`,
```
chipit24, line 14: PHPStorm did this. It's not part of our ESLint rules but we should add it: https://eslint.org/docs/rules/template-curly-spacing.

I slightly prefer no spaces, as I find I interpret them as actual spaces which will be rendered when skimming over template strinsgs.

src/components/nodes/sequenceFlow/sequenceFlow.vue
```diff
@@ -130,9 +130,9 @@ export default {
     targetIsNotSource() {
       return this.targetNode.id !== this.sourceNode.id;
     },
-    allowOutgoingFlow() {
-      return this.sourceConfig.allowOutgoingFlow == null ||
-        this.sourceConfig.allowOutgoingFlow(this.targetNode);
+    allowOutgoingSequenceFlow() {
```
chipit24, line 133: This bit was confusing for me initially. `allowOutgoingFlow` didn't actually seem to be defined in any of the element configs (not in modeler, PM core, or any one of the connectors); so it wasn't actually doing anything. Because of that I decided it was safe to re-name it to something more accurate and useful.

src/components/nodes/association/index.js
chipit24: We decided to add the rule and disallow spaces.

PR #1184, opened 2020-04-09
Implements default gateway flow

### LINKED ISSUE #1182: Default geteway flow
# Default gateway flow
## Description
A Sequence Flow that has an Exclusive, Inclusive Gateway as its source can be defined as default. Such a Sequence Flow will have a marker to show that it is a default flow. A default outgoing Sequence Flow MUST be drawn with a slash marker at the beginning of the connector as seen in the figure.

image

caleeli:
Resolves #1182

This PR includes:
- Add a crown button for sequence flow to set it as default. This button is enabled only for outgoing flows of gateways (exclusive and inclusive)
- The default assignment can be removed clicking in the same button
- Add a marker for the sequence flow assigned as default
- One gateway can have only one default flow
image
image

### DIFF
--- a/src/components/modeler/Modeler.vue
+++ b/src/components/modeler/Modeler.vue
@@ -253,6 +253,9 @@ export default {
   methods: {
     defaultFlow(flow) {
       const source = flow.definition.sourceRef;
+      if (source.default && source.default.id === flow.id) {
+        flow = null;
+      }
       store.commit('setDefaultFlow', { source, flow });
     },
     copyElement(node, copyCount) {
--- a/src/components/modeler/Modeler.vue
+++ b/src/components/modeler/Modeler.vue
@@ -96,7 +96,7 @@
         @setTooltip="tooltipTarget = $event"
         @replace-node="replaceNode"
         @copy-element="copyElement"
-        @default-flow="defaultFlow"
+        @default-flow="toggleDefaultFlow"
       />
     </b-row>
   </span>
@@ -251,12 +251,12 @@ export default {
     },
   },
   methods: {
-    defaultFlow(flow) {
+    toggleDefaultFlow(flow) {
       const source = flow.definition.sourceRef;
       if (source.default && source.default.id === flow.id) {
         flow = null;
       }
-      store.commit('setDefaultFlow', { source, flow });
+      source.set('default', flow);
     },
     copyElement(node, copyCount) {
       const clonedNode = node.clone(this.nodeRegistry, this.moddle, this.$t);
--- a/src/components/modeler/Modeler.vue
+++ b/src/components/modeler/Modeler.vue
@@ -96,6 +96,7 @@
         @setTooltip="tooltipTarget = $event"
         @replace-node="replaceNode"
         @copy-element="copyElement"
+        @default-flow="defaultFlow"
       />
     </b-row>
   </span>
@@ -250,6 +251,10 @@ export default {
     },
   },
   methods: {
+    defaultFlow(flow) {
+      const source = flow.definition.sourceRef;
+      store.commit('setDefaultFlow', { source, flow });
+    },
     copyElement(node, copyCount) {
       const clonedNode = node.clone(this.nodeRegistry, this.moddle, this.$t);
       const yOffset = (node.diagram.bounds.height + 30) * copyCount;
--- a/src/components/nodes/gateway/gateway.vue
+++ b/src/components/nodes/gateway/gateway.vue
@@ -42,6 +42,11 @@ export default {
     'isRendering',
   ],
   mixins: [highlightConfig, portsConfig, hideLabelOnDrag],
+  created() {
+    const flow = this.node.definition.default || null;
+    delete this.node.definition.default;
+    this.$set(this.node.definition, 'default', flow);
+  },
   data() {
     return {
       shape: null,
@@ -89,7 +94,6 @@ export default {
 
     this.shape.addTo(this.graph);
     this.shape.component = this;
-    this.$set(this.node.definition, 'default', null);
   },
 };
 </script>
--- a/src/components/nodes/gateway/gateway.vue
+++ b/src/components/nodes/gateway/gateway.vue
@@ -89,6 +89,7 @@ export default {
 
     this.shape.addTo(this.graph);
     this.shape.component = this;
+    this.$set(this.node.definition, 'default', null);
   },
 };
 </script>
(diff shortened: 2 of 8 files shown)


src/components/nodes/node.js
```diff
@@ -25,6 +25,15 @@ export default class Node {
     return types.includes(this.definition.$type);
   }
 
+  canBeDefaultFlow() {
+    const validSources = [
+      'bpmn:ExclusiveGateway',
+      'bpmn:InclusiveGateway',
+    ];
+    return this.definition.$type === 'bpmn:SequenceFlow'
+      && validSources.indexOf(this.definition.sourceRef.$type) > -1;
```
lisa-fehr, line 34: Alternative:
```suggestion
      && validSources.includes(this.definition.sourceRef.$type);
```

src/components/nodes/sequenceFlow/sequenceFlow.vue
```diff
@@ -77,6 +89,17 @@ export default {
         if (newNameLabel !== this.nameLabel) {
           this.nameLabel = newNameLabel;
         }
+        this.defaultFlow = this.node.definition.sourceRef
+          && this.node.definition.sourceRef.default
+          && this.node.definition.sourceRef.default.id === this.node.definition.id;
```
lisa-fehr, line 94: Used 3 times. Could become a method.
```
isDefaultFlow() {
  return this.node.definition.sourceRef
          && this.node.definition.sourceRef.default
          && this.node.definition.sourceRef.default.id === this.node.definition.id;
}
```

caleeli: Fixed

src/components/modeler/Modeler.vue
```diff
@@ -250,6 +251,13 @@ export default {
     },
   },
   methods: {
+    defaultFlow(flow) {
```
johnblackspear, line 254: Could you rename this to `toggleDefaultFlow` or similar? 

Or at least extract `source.default && source.default.id === flow.id` into a labeled expression that communicate the toggling intent :)

src/components/nodes/gateway/gateway.vue
```diff
@@ -42,6 +42,11 @@ export default {
     'isRendering',
   ],
   mixins: [highlightConfig, portsConfig, hideLabelOnDrag],
+  created() {
+    const flow = this.node.definition.default || null;
+    delete this.node.definition.default;
```
johnblackspear, line 47: I'm a little confused why you're deleting and re-adding it - could you give some overview why this needs to be done this way?

src/components/nodes/sequenceFlow/sequenceFlow.vue
```diff
@@ -68,6 +68,18 @@ export default {
         });
       },
     },
+    defaultFlow: {
+      get() {
+        return this.shape.attr('line').sourceMarker['stroke-width'] > 0;
+      },
```
johnblackspear, line 74: This getter doesn't seem to be called anywhere - and I think that we may be confused in future about whether we should use the `defaultFlow` computed property or the `isDefaultFlow()` method to determine whether a flow is the default. Could this be removed?

src/components/nodes/sequenceFlow/sequenceFlow.vue
```diff
@@ -68,6 +68,18 @@ export default {
...
+      set(value) {
+        this.shape.attr('line', {
+          sourceMarker: {
+            'stroke-width': value ? 2 : 0,
+          },
+        });
+      },
+    },
```
johnblackspear, line 82: If you get rid of the getter then this can be its own method, removing the need for the computed property.

src/components/nodes/sequenceFlow/sequenceFlow.vue
```diff
@@ -77,11 +89,23 @@ export default {
         if (newNameLabel !== this.nameLabel) {
           this.nameLabel = newNameLabel;
         }
+        this.defaultFlow = this.isDefaultFlow();
```
johnblackspear, line 92: See comments above regarding the computed property, this could be a call to whatever method handles the stroke at the flow origin

src/components/nodes/sequenceFlow/sequenceFlow.vue
```diff
@@ -137,11 +161,22 @@ export default {
...
       }]);
     },
+    createDefaultFlowMarker() {
+      const isDefault = this.isDefaultFlow();
+      this.shape.attr('line', {
+        sourceMarker: {
+          'type': 'polyline',
+          'stroke-width': isDefault ? 2 : 0,
+          points: '2,6 6,-6',
+        },
```
johnblackspear, line 171: This could be at the heart of the method that you have instead of the computed property. I'd also suggest inlining the `isDefault` const.

src/store.js
```diff
@@ -52,6 +52,10 @@ export default new Vuex.Store({
     allowSavingElementPosition: state => state.allowSavingElementPosition,
   },
   mutations: {
+    setDefaultFlow(state, { source, flow }) {
+      source.set('default', flow);
+      makeDefinitionPropertyReactive(source, 'default', flow);
+    },
```
johnblackspear, line 58: This doesn't need to be on the store - since it doesn't read or set any state in the store. You can basically inline the `source.set` in Modeler.vue where you commit (i.e. `store.commit('setDefaultFlow', { source, flow });`)

src/components/modeler/Modeler.vue
caleeli: Renamed to "toggleDefaultFlow"

src/components/nodes/gateway/gateway.vue
caleeli: Is to fix a weird issue when loading a process with a gateway with a defined "default" property, it is not reactive, then the user can not change the default flow

src/components/nodes/sequenceFlow/sequenceFlow.vue
caleeli: Removed

src/components/nodes/sequenceFlow/sequenceFlow.vue
caleeli: Computed removed, changed to a `setDefaultMarker` method

src/components/nodes/sequenceFlow/sequenceFlow.vue
caleeli: Fixed

src/store.js
caleeli: Removed from store, moved to Modeler.vue

caleeli:
@johnblackspear thanks for your comments, I updated the PR with the fixes

PR #1202, opened 2020-04-17
Validation of signal reference

caleeli:
Requires: https://github.com/ProcessMaker/bpmnlint-plugin/pull/6

This PR includes:

- Add required validation for signal selector

image

### DIFF
--- a/src/components/inspectors/SignalSelect.vue
+++ b/src/components/inspectors/SignalSelect.vue
@@ -1,7 +1,7 @@
 <template>
   <div>
     <label class="typo__label">{{ label }}</label>
-    <div class="d-flex">
+    <div class="d-flex" :class="{invalid}">
       <multiselect
         :value="selectedOption"
         @input="change"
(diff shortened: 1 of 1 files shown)


src/components/inspectors/SignalSelect.vue
```diff
@@ -347,3 +353,9 @@ export default {
     padding: 0px
   }
 </style>
+
+<style>
+  .invalid .multiselect__tags {
+    border-color:#f04124!important
```
chipit24, line 359: This border color is different from the border color of other invalid elements in the inspector panel:
Screen Shot 2020-04-20 at 10 23 58 AM

It looks like the other inputs have an invalid border color of `#dc3545`.

The invalid style that applies the above red color comes from vue-form-elements:
https://github.com/ProcessMaker/vue-form-elements/blob/711586cc154f1312f4cd22d8d226b216ae6d28d1/src/components/FormHtmlViewer.vue#L27-L30

PR #1210, opened 2020-04-21
Move element configs from core to modeler

### LINKED ISSUE #1200: Move node configuration extensions from core into modeler
**Description**
The [`registerInspectorExtension`](https://github.com/ProcessMaker/modeler/blob/develop/src/components/InspectorExtensionManager.js) method is used to hook into the modeler initialization process and add custom properties to the inspector panel for elements. This is intended to add custom behaviour that should not be available to all modeler instances; the intended use-case would be to add custom properties for connectors.

However, this method is currently used in core's [`initialLoad.js`](https://github.com/ProcessMaker/processmaker/blob/develop/resources/js/processes/modeler/initialLoad.js) file to add inspector properties that can just as easily be added directly in modeler.

**Benefits**
The benefit of keeping these properties defined in modeler's element configs is:
 * It makes adding validation easier, as validation rules can be explicitly added to bpmnlint and it will work consistently across modeler instances.
 * Having the properties be part of the initial, default config, co-located with the actual component, allows easier refactoring and modification of the properties.
 * Allows fully testing the properties, and how the components work with them.

**Solution**
The process for moving an extension over from core to modeler is quite simple:
1. Remove the `registerInspectorExtension` line from `initialLoad.js` in core.
2. Copy the inspector config lines into the element's "index.js" config file.
3. Copy cover any inspector config form elements into modeler's `src/components/inspectors` folder.

**Estimate**
There are currently up to 18 of these `registerInspectorExtension` calls that can be moved over into modeler. It's possible to move many or all of these at once, though it's not necessary to move them all, or even to move them all together in one PR. At least two PRs would be required to complete this move, one in modeler for the addition, and one in core for the removal. Estimate for completion is 2 days.

chipit24:
Fixes #1200.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,8 @@ The [entry point](https://webpack.js.org/configuration/entry-context/#entry) for
 
 This event is fired before the modeler and [BpmnModdle](https://github.com/bpmn-io/bpmn-moddle) are set up and mounted. Listeners to this event are passed an object with three methods: `registerInspectorExtension`, `registerBpmnExtension`, and `registerNode`.
 
+You call the `registerInspectorExtension` method to hook into the modeler initialization process to add custom properties to the inspector panel for elements. This is intended to add custom behaviour that should **not** be available to all modeler instances; the intended use-case is to add custom properties for connectors.
+
 ```javascript
 import bpmnExtension from '@processmaker/processmaker-bpmn-moddle/resources/processmaker.json';
 import { intermediateMessageCatchEvent } from '@/components/nodes';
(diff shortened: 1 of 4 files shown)


chipit24:
Closing: This issue has turned out to be much more involved than initially anticipated. The following additional work would mean https://github.com/ProcessMaker/modeler/issues/1200 could take 1 week or more to fully complete, meaning we won't be available to finish it:

 * Inspector config properties are registered exclusively on the base start event. However, the other start events extend off this config, and to ensure these additions remain only on the base start event, a separate, minimal start event would have to be created, similar to tasks.
 * All custom inspector additions reference a custom component that will need to be ported over to modeler:
    * To keep PRs small enough to properly review and test, these would have to be brought over one-by-one in separate PRs.
    * Most of these components require additional work to fix lintinig and styling issues, and minor bugs.
 * Like with any change, bugs could be introduced, and that is something we want to minimize at this time.

PR #1213, opened 2020-04-22
Add reset for shift key combos

### LINKED ISSUE #1169: Investigate: Sometimes cursor is stuck as crosshair
# Bug report issue
## Description
When interacting with the diagram to add boundary events or change colors, the cursor can stay stuck in a select crosshair.
## Reproduction steps
- Press "shift + command + 4" to take a screenshot of the diagram.
- Move the cursor around or drag an element on the diagram.
## Expected behaviour
You should not get stuck with the crosshair cursor.
## Actual behaviour
You cannot click to restore the normal drag cursor.
## Software used
Lastest Modeler
## Other information
crosshair-bug

lisa-fehr:
Closes #1169

### DIFF
--- a/tests/e2e/specs/Multiselect.spec.js
+++ b/tests/e2e/specs/Multiselect.spec.js
@@ -123,7 +123,7 @@ describe('Multiselect', () => {
     });
   });
 
-  it('should not move multiple shapes when shift + combo is held down', () => {
+  it('should not move multiple shapes when shift + combo was held down', () => {
     [
       {combo: 'meta', task1Position: { x: 100, y: 100 }},
       {combo: 'ctrl', task1Position: { x: 400, y: 100 }},
--- a/tests/e2e/specs/Multiselect.spec.js
+++ b/tests/e2e/specs/Multiselect.spec.js
@@ -122,4 +122,45 @@ describe('Multiselect', () => {
       });
     });
   });
+
+  it('should not move multiple shapes when shift + combo is held down', () => {
+    [
+      {combo: 'meta', task1Position: { x: 100, y: 100 }},
+      {combo: 'ctrl', task1Position: { x: 400, y: 100 }},
+      {combo: 'alt', task1Position: { x: 700, y: 100 }},
+    ].forEach(({combo, task1Position}) => {
+      const task2Position = { x: task1Position.x + 100, y: task1Position.y + 100 };
+      const translateAmount = 200;
+      const newTask1Position = { x: task1Position.x + translateAmount, y: task1Position.y + translateAmount };
+
+      dragFromSourceToDest(nodeTypes.task, task1Position);
+      dragFromSourceToDest(nodeTypes.task, task2Position);
+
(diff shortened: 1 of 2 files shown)


tests/e2e/specs/Multiselect.spec.js
```diff
@@ -122,4 +122,45 @@ describe('Multiselect', () => {
...
+      {combo: 'alt', task1Position: { x: 700, y: 100 }},
+    ].forEach(({combo, task1Position}) => {
+      const task2Position = { x: task1Position.x + 100, y: task1Position.y + 100 };
+      const translateAmount = 200;
+      const newTask1Position = { x: task1Position.x + translateAmount, y: task1Position.y + translateAmount };
+
+      dragFromSourceToDest(nodeTypes.task, task1Position);
+      dragFromSourceToDest(nodeTypes.task, task2Position);
+
+      cy.get('.paper-container').click();
```
chipit24, line 139: I realize this was taken from the above test, and the same applies there as well, but since the `.paper-container` selector is used here first, I think it would make more sense to specify the alias here.

```suggestion
      cy.get('.paper-container').as('paperContainer').click();
```

tests/e2e/specs/Multiselect.spec.js
```diff
@@ -122,4 +122,45 @@ describe('Multiselect', () => {
...
+
+      cy.get('body').type(`{shift}{${ combo }}`, { release: false });
+      cy.get('.paper-container').as('paperContainer').trigger('mousedown', 'topLeft');
```
chipit24, line 142: And then this line can become:

```suggestion
      cy.get('@paperContainer').trigger('mousedown', 'topLeft');
```

tests/e2e/specs/Multiselect.spec.js
```diff
@@ -122,4 +122,45 @@ describe('Multiselect', () => {
...
+          const { x, y } = elements.find(el => el.get('id') === $task.attr('model-id')).position();
+          expect({ x, y }).to.eql(newTask1Position);
+        });
+      });
+      getElementAtPosition(task2Position, nodeTypes.task).then($task => {
+        getGraphElements().then(elements => {
+          const { x, y } = elements.find(el => el.get('id') === $task.attr('model-id')).position();
+          expect({ x, y }).to.eql(task2Position);
+        });
+      });
```
chipit24, line 163: When I comment out your fix (lines 46-49 in `setUpSelectionBox.js`), this test fails, however, it doesn't fail due to the assertion, but because of a type error:
```
TypeError: Cannot read property 'attr' of null
```

So if this behaviour is broken in the future, and this test fails the same way, it may not be obvious why exactly its broken or how to fix it. I think if the fix is commented out, the test should fail with an assertion.

To keep this test small and focused, maybe you don't have to add elements to the page, maybe trigger a <kbd>shift</kbd> key press with the combo buttons, a mouse down + move (but no mouse up), and then assert the cursor is the default one, and that no selection box is visible.

PR #1218, opened 2020-04-28
Element Description syncs to similar element

### LINKED ISSUE #1201: Copy and Paste Element Description
# Bug report issue
## Description
If you edit the description on a copied element, it updates the description on the element copied.
## Reproduction steps
**Scenario 1**
- Add a Start Event
- Add an Advanced > Description
- Press Copy
- Edit the Description on the copy
- View the original Description

Apr-16-2020 17-01-17

**Scenario 2**
- Add a Task
- Add two Boundary Events
- Update the Advanced > Description on one Boundary Event
- View the Description on the other Boundary Event

## Expected behaviour
Description should be unique for original and copy in Scenario 1.
Description should be unique between two Boundary Events on the same task in Scenario 2.
## Actual behaviour
The copy's Description syncs with the original in Scenario 1, and Description syncs with the first and second Boundary Events in Scenario 2.
## Software used
Current Modeler: 0.23.0

lisa-fehr:
Closes #1201

### DIFF
--- a/src/components/inspectors/InspectorPanel.vue
+++ b/src/components/inspectors/InspectorPanel.vue
@@ -12,7 +12,7 @@
         :style="{ height: parentHeight }"
       >
         <vue-form-renderer
-          :key="config[0].name"
+          :key="highlightedNode._modelerId"
           v-if="highlightedNode"
           :data="data"
           @update="updateDefinition"


lisa-fehr:
I'd like to make a test. But this error doesn't seem to be reproducible in a cypress test.

PR #1241, opened 2020-10-08
Move replace logic into migration class

chipit24:


chipit24:
Closed as fork is no longer required. Use https://github.com/ProcessMaker/modeler/pull/1242.

PR #1252, opened 2020-11-10
Fix Colors

### LINKED ISSUE #1250: Color Picker Invalid BPMN
attribute 'color' is not supported

**Steps to Reproduce:**

Create process
Create task
Change task color with pallete color
Try start new Request

**Current Behavior:**

In Process Modeler, adding a color to a task using the new Color Palette option, will throw a bpmn validation error about the COLOR tag. This will prevent the process to be started from the Request button and show the validation warning icon in the Process list. This behavior obviously undermines the whole Color option.

Side note, a process ´broken´this way can still be started via a web entry on a start event. So looks like its just our UI performing validation and preventing the start.

vehikl-jacare:
Safeguards some behaviour related to #1250, but the fix is done in https://github.com/ProcessMaker/processmaker-bpmn-moddle/pull/40.


vehikl-jacare:
Removed WIP label as the main fix is done outside of modeler.

PR #1267, opened 2020-11-18
Add validation to data store and object

vehikl-jacare:
Sequence flow connections should not be allowed to/from data object and data stores. These BPMN elements require data input/output associations.

Screen Shot 2020-11-18 at 3 37 07 PM
Screen Shot 2020-11-18 at 3 37 24 PM


vehikl-jacare:
WIP: Add tests

PR #1269, opened 2020-11-19
Modeler validation throws "Element is missing ID." when adding a pool control

marcoAntonioNina:
Fixed [FOUR-1773](https://processmaker.atlassian.net/browse/FOUR-1773)

- Add id by default in LaneSet, this ID has no other functionality only for validations.

### DIFF
--- a/src/components/nodes/pool/pool.vue
+++ b/src/components/nodes/pool/pool.vue
@@ -182,6 +182,7 @@ export default {
     createLaneSet() {
       const laneSet = this.moddle.create('bpmn:LaneSet');
       this.laneSet = laneSet;
+      this.laneSet.set('id', 'laneset_0');
       this.containingProcess.get('laneSets').push(laneSet);
     },
     pushNewLane(definition = Lane.definition(this.moddle, this.$t)) {
--- a/src/components/nodes/pool/pool.vue
+++ b/src/components/nodes/pool/pool.vue
@@ -182,7 +182,9 @@ export default {
     createLaneSet() {
       const laneSet = this.moddle.create('bpmn:LaneSet');
       this.laneSet = laneSet;
-      this.laneSet.set('id', 'laneset_0');
+      const generator = this.$parent.nodeIdGenerator;
+      const [laneSetId] = generator.generate();
+      this.laneSet.set('id', laneSetId);
       this.containingProcess.get('laneSets').push(laneSet);
     },
     pushNewLane(definition = Lane.definition(this.moddle, this.$t)) {
--- a/src/components/nodes/pool/pool.vue
+++ b/src/components/nodes/pool/pool.vue
@@ -67,6 +67,7 @@ export default {
     'planeElements',
     'isRendering',
     'paperManager',
+    'nodeIdGenerator',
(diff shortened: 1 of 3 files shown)


src/components/nodes/pool/pool.vue
```diff
@@ -182,6 +182,7 @@ export default {
     createLaneSet() {
       const laneSet = this.moddle.create('bpmn:LaneSet');
       this.laneSet = laneSet;
+      this.laneSet.set('id', 'laneset_0');
```
vehikl-jacare, line 185: This ID should be unique, and may be the source of the duplicate ID and unresolved reference errors. We have a module that generates unique IDs (`NodeIdGenerator.js`) that we use for generating unique IDs for all other nodes; you could try using that here as well.

vehikl-jacare (changes requested):
We tested this fix against the following diagram:
Screen Shot 2020-11-19 at 1 50 02 PM

Importing that diagram into Camunda caused the following errors:
> [/Users/vehikl/Desktop/with_lane_id_fix.xml] unparsable content <bpmn:laneSet> detected
	line: 24
	column: 4
	nested error: duplicate ID <laneset_0> [ warning ]
[/Users/vehikl/Desktop/with_lane_id_fix.xml] unresolved reference <node_11> [ warning ]
[/Users/vehikl/Desktop/with_lane_id_fix.xml] unresolved reference <node_12> [ warning ]
[/Users/vehikl/Desktop/with_lane_id_fix.xml] no bpmnElement referenced in <bpmndi:BPMNShape id="node_11_di" /> [ warning ]
[/Users/vehikl/Desktop/with_lane_id_fix.xml] no bpmnElement referenced in <bpmndi:BPMNShape id="node_12_di" /> [ warning ]

Please also add a test for this, to prevent the bug from re-surfacing.

src/components/nodes/pool/pool.vue
caleeli:
```
import NodeIdGenerator from '@/NodeIdGenerator';
const generator = new NodeIdGenerator(this.definitions);
const [laneSetId] = generator.generate();
```

src/components/nodes/pool/pool.vue
```diff
@@ -182,6 +182,9 @@ export default {
     createLaneSet() {
       const laneSet = this.moddle.create('bpmn:LaneSet');
       this.laneSet = laneSet;
+      const generator = this.$parent.nodeIdGenerator;
```
vehikl-jacare, line 185: Instead of using `this.$parent` could you please pass the generator in as a prop?

vehikl-jacare (changes requested):
Please see the comment re `this.$parent`, and also please add a test for the `id` behaviour.

You are also welcome to assign to us and we can add a test.

caleeli:
+ Node generator was added as a property.
+ Added test to check LaneSet IDs

boliviacoca:
@vehikl-jacare please feel free to let us know if looking good.

Thanks for your kind help,
Kind regards

boliviacoca (approved):

PR #1285, opened 2020-12-04
Refresh inspector after the node is migrated

caleeli:
Fixes: https://processmaker.atlassian.net/browse/FOUR-2406

This PR changes:
- Refreshes the inspector after a node type migration completes, to prevent the user from seeing the name blink

ChangeNode mp4

### DIFF
--- a/src/components/inspectors/InspectorPanel.vue
+++ b/src/components/inspectors/InspectorPanel.vue
@@ -50,7 +50,6 @@ import get from 'lodash/get';
 import cloneDeep from 'lodash/cloneDeep';
 import Process from './process';
 import isString from 'lodash/isString';
-import isEqual from 'lodash/isEqual';
 
 Vue.component('FormText', renderer.FormText);
 Vue.component('FormInput', FormInput);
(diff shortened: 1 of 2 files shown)


vehikl-jacare (commented):
Testing locally, this doesn't appear to fix the issue with the name flashing.

You may be able to look into using `this.paperManager.performAtomicAction` to batch actions related to rendering the jointjs shapes.

caleeli:
Hi @vehikl-jacare, thanks for your comment, oh the chage fix the issue in the inspector, but I didn't see the flashing in the diagram. Could you send me a suggestion of how to use `this.paperManager.performAtomicAction` to fix that.

caleeli:
Hi @vehikl-jacare, I've added a different change to fix the name flashing in the diagram, could you check it please.

vehikl-jacare:
Hi @caleeli, we've implemented a solution for the inspector flashing that works for both single and multi-select. 

Canvas flashing is still happening, and we think it requires a more involved solution and could be its own ticket.

caleeli:
@vehikl-jacare do you have a ticket or PR to fix the flashing at inspector?
@boliviacoca could we divide this issue in 2?

boliviacoca:
I think we can @caleeli, please proceed.

caleeli:
A new issue was created to fox the name flashing: https://github.com/ProcessMaker/modeler/issues/1289

caleeli:
I've tested this branch but the name is still flashing in the inspector when the start event type is changed. Ex

`Custom Name -> Start Timer Event -> Custom Name`

NameFlashing mp4

vehikl-jacare:
The issue of the name flashing in the inspector is fixed on develop. The flashing in the diagram is fixed by #1301.

PR #1303, opened 2020-12-14
Update to nvm 14 (in line with core)

vehikl-jacare:
This needs to be the same as Processmaker Core


vehikl-jacare:
No longer needed

PR #1305, opened 2020-12-18
Custom task icons

### LINKED ISSUE #1249: Task Icons
We'd like to able to change the icon icon for any task with a custom one. This icon will replace the icon in the top left of the task element and this functionality needs to be available to Form, Manual, Call Activity and Service tasks.

vehikl-jacare:
Fixes #1249.

Adds the ability to add custom icons on:
 - Form Task
 - Manual Task
 - Script Task
 - Call Activity (Sub Process).

It requires https://github.com/ProcessMaker/processmaker-bpmn-moddle/pull/43 to emit valid BPMN, as the `customIcon` attribute needs to be properly namespaced.

Screen Shot 2020-12-18 at 1 50 04 PM

Screen Shot 2020-12-18 at 5 54 22 PM

Screen Shot 2020-12-18 at 5 54 39 PM

Screen Shot 2020-12-18 at 5 55 07 PM


package.json
```diff
@@ -93,6 +94,7 @@
     "husky": "^4.3.5",
     "i18next": "^15.0.8",
     "istanbul-lib-coverage": "^2.0.5",
+    "js-yaml-loader": "^1.2.2",
```
vehikl-jacare, line 97: This is already specified in `dependencies` and can safely be removed from here.

src/components/nodes/task/task.vue
```diff
@@ -111,6 +116,9 @@ export default {
     },
   },
   methods: {
+    resetCustomIconName() {
+      this.setCustomIconName('');
+    },
```
vehikl-jacare, line 121: This looks like it's the same `resetCustomIconName` method that's defined in the `customIcon` mixin and can safely be removed from here.

src/mixins/updateIconColor.js
```diff
@@ -6,26 +6,47 @@ export default {
...
+  computed: {
+    icon: {
+      cache: false,
+      get() {
+        if (this.node.definition.get('customIcon')) {
+          const customIcon = this.node.definition.get('customIcon');
+          return atob(customIcon);
+        }
+        return this.nodeIcon;
+      },
```
vehikl-jacare, line 18: We may not be to set `cache: false` if we switch from `.get` to referencing the reactive property directly. Vue might not be aware of changes to the node definition properties if they are referenced through that `get` method.

src/mixins/updateIconColor.js
```diff
@@ -6,26 +6,47 @@ export default {
...
+    },
+  },
   watch: {
     'node.definition.color': {
       handler() {
-        this.setIconColor();
+        this.updateIconColor();
+      },
+      deep: true,
```
vehikl-jacare, line 26: `color` is a string value and `deep: true` might not be needed.

src/components/crown/crownButtons/crownColorDropdown.vue
```diff
@@ -82,18 +125,44 @@ export default {
...
-      store.commit('updateNodeProp', { node: this.node, key: 'color', value: undefined });
-      this.$emit('save-state');
+    resetNodeStyling() {
+      this.resetNodeColor();
+      this.resetCustomIcon();
     },
     setNodeColor(color) {
-      Vue.set(this.node.definition, 'color', color);
       store.commit('updateNodeProp', { node: this.node, key: 'color', value: color });
+      Vue.set(this.node.definition, 'color', color);
```
vehikl-jacare, line 134: Setting the property so it's reactive to Vue may already happen in the `updateNodeProp` store mutation and might not be required here.

vehikl-jacare: `Vue.set` cannot be safely removed until the `pm:customIcon` property has been added to the BPMN extensions.

PR #1321, opened 2021-01-06
Stop bootstrap from disabling focus on modal with tinymce

nolanpro:
Fixes https://processmaker.atlassian.net/browse/FOUR-2636
- Also make the inspector rich text editor 100% since it was only taking up half the width


vehikl-jacare (approved):
The TinyMCE rich text editor runs within an iframe and as far as we understand this change allows the iframe to receive focus. 

Looks good, and thanks - though we are still seeing some unrelated intermittent CI failures that we are working to resolve. Most of these are timeout related due to new animations.

boliviacoca (approved):

PR #1329, opened 2021-01-26
FOUR-2718: Signal catch: Inspector for signal start event is showing all process signals

danloa:
Fixes [https://processmaker.atlassian.net/browse/FOUR-2718](https://processmaker.atlassian.net/browse/FOUR-2718)

The option of edit/insert signals is available just in intermediate throw signal event and end signal event.


caleeli (approved):
It works

boliviacoca (approved):

PR #1346, opened 2021-04-12
Ethos Events Package requirements

gusys:
The Process element are exposed to extend InspectorsExtension for Ellucian Ethos Events Package


gusys:
This pull request is no longer necessary, the schema configuration for the Ellucian Ethos Events package was moved to the process configuration section

PR #1374, opened 2021-08-18
Make undo bring back full changes to flowchart, not inspector settings

nolanpro:
Fixes http://tickets.pm4overflow.com/tickets/590

Kapture 2021-08-18 at 16 01 41


nolanpro:
The fix is working but the tests are not passing. Moved to backlog

PR #1375, opened 2021-09-06
 [Loops] Change "Loop Characteristics" option

marcoAntonioNina:
Fixes [ticket 693](http://tickets.pm4overflow.com/tickets/693)

- Change loop radio buttons to dropdown Loop Mode
- Change labels

### DIFF
--- a/src/components/inspectors/LoopCharacteristics.vue
+++ b/src/components/inspectors/LoopCharacteristics.vue
@@ -1,7 +1,7 @@
 <template>
   <div>
     <b-form-group :label="$t('Loop Mode')">
-      <b-form-select id="" v-model="loopType">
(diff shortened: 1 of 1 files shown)


src/components/inspectors/LoopCharacteristics.vue
```diff
@@ -190,6 +185,9 @@ export default {
         }
       },
     },
+    loopType(value) {
```
caleeli, line 188: Please do not use watchers. Is it possible to change it to @change event?

marcoAntonioNina: changed

caleeli (approved):

PR #1380, opened 2021-09-15
[Loops] Add Loop option to Loop Modes

marcoAntonioNina:
Fixes [ticket 695](http://tickets.pm4overflow.com/tickets/695)

- add options to loop standard
- option loopMaximum
- option loopCondition
image


src/components/inspectors/LoopCharacteristics.vue
```diff
@@ -1,10 +1,22 @@
+/* eslint-disable no-console */
```
caleeli, line 1: Remove this debug line

PR #1437, opened 2022-04-13
Bugfix/four 5628 : Error when redesigning a new process in an existing one

AbdelilahLbardi:
## Issue & Reproduction Steps

1 - Have a process that contains in its design a majority of all the controls or create a new process that has at least the following controls: tasks, gateways, pool with lines, start and end event.
2 - Open the process and remove all controls, after that create a new process design and save it.
3 - Reload the page of the design process.

Expected behavior: 
Lanes to be deleted when the Pool get deleted.

Actual behavior: 
Deleting a pool does not deleted related Lanes.

## Solution
- Defining the deletion of `Pool`'s `Lanes` on `removeNodesFromPool` within the `Modeler`.

## How to Test
Test the steps above

## Related Tickets & Packages
-  None

## Regression
- I suspect that the regression was done in the following change: https://github.com/ProcessMaker/modeler/commit/6bcca1c58b571c537c15874671899a66c3796039

## Code Review Checklist

### DIFF
--- a/src/components/modeler/Modeler.vue
+++ b/src/components/modeler/Modeler.vue
@@ -881,7 +881,7 @@ export default {
           });
         }
         let laneSets = node.definition.processRef.laneSets;
-        if (laneSets != null && laneSets.length > 0) {
+        if (!!laneSets && laneSets.length > 0) {
           laneSets[0].lanes.forEach(lane => {
             const nodeToRemove = this.nodes.find(n => n.definition === lane);
             if (nodeToRemove) {
--- a/src/components/modeler/Modeler.vue
+++ b/src/components/modeler/Modeler.vue
@@ -880,6 +880,15 @@ export default {
             }
           });
         }
+        let laneSets = node.definition.processRef.laneSets;
+        if (laneSets != null && laneSets.length > 0) {
+          laneSets[0].lanes.forEach(lane => {
+            const nodeToRemove = this.nodes.find(n => n.definition === lane);
+            if (nodeToRemove) {
+              this.removeNode(nodeToRemove);
+            }
+          });
+        }
       }
     },
     handleResize() {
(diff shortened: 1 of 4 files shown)


AbdelilahLbardi:
Pushing tests changes first because I expect them to fail which happened [here](https://app.circleci.com/pipelines/github/ProcessMaker/modeler/3342/workflows/2ee9d86a-633d-4eb6-be62-40bed8e60133/jobs/3979)

src/components/modeler/Modeler.vue
```diff
@@ -880,6 +880,15 @@ export default {
             }
           });
         }
+        let laneSets = node.definition.processRef.laneSets;
+        if (laneSets != null && laneSets.length > 0) {
```
josechirivella, line 884:
```suggestion
        if (laneSets !== null && laneSets.length > 0) {
```

josechirivella (changes requested):

src/components/modeler/Modeler.vue
AbdelilahLbardi: Hey @josechirivella, thanks for the review. I did not test the type to avoid getting back to it in case `lanseSets` is set to `undefined` instead of `null`. What do you think?

josechirivella: If we are worried for the value to be of falsy type such as `null` `undefined` or `false` we can always do `!!`
https://betterprogramming.pub/javascript-bang-bang-i-shot-you-down-use-of-double-bangs-in-javascript-7c9d94446054
such as
```suggestion
        if (!!laneSets && laneSets.length > 0) {
```

caleeli (approved):

josechirivella (approved):

ryancooley (approved):

PR #1440, opened 2022-05-19
Bugfix/four-5628 : Error when redesigning a new process in an existing one

AbdelilahLbardi:
## Issue & Reproduction Steps

- Import the client process from FOUR-5628
- You should be able to see some console errors
- When deleting the pool, you will get an error.

Expected behavior:

- Process getting displayed with no errors.
- Deletion works fine

Actual behavior: 

- When importing the process you will get console errors
- You will not be able to delete the pool

## Solution
- Tweaking the XML manager file to fix broken lanes that has no shape within.

## How to Test
Test the steps above

## Related Tickets & Packages
- 

## Code Review Checklist

### DIFF
--- a/src/components/modeler/XMLManager.js
+++ b/src/components/modeler/XMLManager.js
@@ -32,7 +32,6 @@ export default class XMLManager {
   }
 
   cleanBrokenReferences(definitions) {
-    
     const { rootElements, diagrams } = definitions;
     const removed = [];
 
--- a/src/components/modeler/XMLManager.js
+++ b/src/components/modeler/XMLManager.js
@@ -32,6 +32,7 @@ export default class XMLManager {
   }
 
   cleanBrokenReferences(definitions) {
+    
     const { rootElements, diagrams } = definitions;
     const removed = [];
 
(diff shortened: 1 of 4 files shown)


src/components/modeler/XMLManager.js
```diff
@@ -47,6 +48,7 @@ export default class XMLManager {
           }
           return true;
         });
+        element.laneSets[0].lanes = element.laneSets[0].lanes.filter(node => !!node.flowNodeRef);
```
caleeli, line 51: There are two problems:

- This code is inside `if (element.$type === 'bpmn:Process' && element.flowElements) {` but should be outside because LaneSets are at the same level of Process Elements.
- Also take in could that the diagram could have more than one pool, so the LaneSets could be more than one.

caleeli (changes requested):
@AbdelilahLbardi please review the comment

agustinbusso:
@AbdelilahLbardi Also would be nice to have a test for this, you can take a look into **FixBrokenSequenceFlows.spec.js**:

Maybe you can create some processes, export it, open the BPMN, break it and check after uploadXml the process is fixed.  A process with more than one pool it would be a great use case. Also with single lane and multiple lanes.

caleeli (approved):

agustinbusso (approved):

ryancooley (approved):

PR #1450, opened 2022-08-17
FOUR-6606: Dimensions of Some Modeler Crown-Config Icons are Inconsistent

mcraeteisha:
# Issue
Ticket: [FOUR-6606](https://processmaker.atlassian.net/browse/FOUR-6606)

Numerous icons in the Modeler crown-config menu have varying heights. This makes the icons noticeably smaller and inconsistent with their neighbors.

# Solution
Updated heights of all icons to 15px (if they were not already at that height).

# How to Test
1. Update Modeler branch to `bugfix/FOUR-6606`
2. Link Modeler to Core
3. Open local environment, create a new process
4. Add the following elements to your process: Start Event, Form Task, Exclusive Gateway (with a Default Flow), Text Annotation, Data Object. This will allow you to see all of the different crown icons.
5. Open your Inspector to see that heights of icons have been set to 15px. Note: some icons were already set to a height of 15px, although the height is not listed as a style attribute. Inspecting the element should still show a height of 15px:
Screen Shot 2022-08-17 at 12 11 04 PM

# Related Tickets & Packages
- [FOUR-5110](https://processmaker.atlassian.net/browse/FOUR-5110) | Design Bugs

# Code Review Checklist
- [ ]  I have pulled this code locally and tested it on my instance, along with any associated packages.
- [ ]  This code adheres to [ProcessMaker Coding Guidelines](https://github.com/ProcessMaker/processmaker/wiki/Coding-Guidelines).
- [ ]  This code includes a unit test or an E2E test that tests its functionality, or is covered by an existing test.
- [ ]  This solution fixes the bug reported in the original ticket.
- [ ]  This solution does not alter the expected output of a component in a way that would break existing Processes.
- [ ]  This solution does not implement any breaking changes that would invalidate documentation or cause existing Processes to fail.
- [ ]  This solution has been tested with enterprise packages that rely on its functionality and does not introduce bugs in those packages.
- [ ]  This code does not duplicate functionality that already exists in the framework or in ProcessMaker.
- [ ]  This ticket conforms to the PRD associated with this part of ProcessMaker.

### DIFF
--- a/src/components/crown/crownButtons/addLaneAboveButton.vue
+++ b/src/components/crown/crownButtons/addLaneAboveButton.vue
@@ -9,7 +9,6 @@
       :src="laneAboveIcon"
       aria-hidden="true"
       class="lane-above-icon"
-      alt=""
     >
   </crown-button>
 </template>
--- a/src/components/crown/crownButtons/addLaneAboveButton.vue
+++ b/src/components/crown/crownButtons/addLaneAboveButton.vue
@@ -2,10 +2,16 @@
   <crown-button
     id="lane-above-button"
     aria-label="Add lane above icon"
-    :src="laneAboveIcon"
     v-on="$listeners"
     :width="25"
-  />
+  >
+    <img
+      :src="laneAboveIcon"
+      aria-hidden="true"
+      class="lane-above-icon"
+      alt=""
+    >
+  </crown-button>
 </template>
 
 <script>
@@ -21,3 +27,8 @@ export default {
   },
 };
 </script>
+<style>
+  .lane-above-icon {
+    height: 15px;
+  }
+</style>
--- a/src/components/crown/crownButtons/addLaneAboveButton.vue
+++ b/src/components/crown/crownButtons/addLaneAboveButton.vue
@@ -8,7 +8,6 @@
     <img
       :src="laneAboveIcon"
       aria-hidden="true"
-      class="lane-above-icon"
     >
   </crown-button>
 </template>
@@ -26,8 +25,3 @@ export default {
   },
 };
 </script>
-<style>
-  .lane-above-icon {
-    height: 15px;
-  }
-</style>
--- a/src/components/crown/crownButtons/crownButton.vue
+++ b/src/components/crown/crownButtons/crownButton.vue
@@ -40,8 +40,12 @@ export default {
   margin-top: 0;
   display: flex;
 }
-img, i {
+img {
   margin: 0px 10px;
   height: 15px;
+}
+i {
+  margin: 0px 10px;
+  font-size: 15px;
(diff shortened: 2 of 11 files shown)


src/components/crown/crownButtons/addLaneAboveButton.vue
```diff
@@ -2,10 +2,16 @@
...
-    :src="laneAboveIcon"
     v-on="$listeners"
     :width="25"
-  />
+  >
+    <img
+      :src="laneAboveIcon"
+      aria-hidden="true"
+      class="lane-above-icon"
+      alt=""
```
josechirivella, line 12: This is a global comment. If the alt values are going to be empty, please remove.

josechirivella (commented):
Looks good but added a non-blocker comment

src/components/crown/crownButtons/addLaneAboveButton.vue
mcraeteisha: Empty alt values have been removed.

agustinbusso (commented):
Looks good. There are only some details that maybe we can improve. For example the gear icon have a height of 16px instead 15 and the lane icons does not have the real 15px dimensions because is limited by the width. There are 4 things I would do it:

**(Please watch the video)**

1- The first thing I would do it is to center the elements in the main container
2- The second thing I would do it is two generic classes one for `<i>` elements and another for `` elements. So if we need to add any change in the future for all crown icons we only need to change those two classes.
3- The other thing I would do it is to remove the width of the elements so the height is 
the attribute that is going to define the dimension and is not going to be limited by the height.
4- And the last I would do it is to remove the vertical margin in each item.


https://user-images.githubusercontent.com/90727999/185223049-f199c8a3-098e-47b7-a717-4bf4f690f795.mov

src/components/crown/crownButtons/crownButton.vue
```diff
@@ -41,7 +41,7 @@ export default {
   display: flex;
 }
 img, i {
```
agustinbusso, line 43: One last minor change: for `<i>` elements you should use font-size instead height. Notice the following screenshot, the pintor palette have a height of 16px:
Screen Shot 2022-08-17 at 16 59 35

agustinbusso (changes requested):

agustinbusso:
Looks great!

https://user-images.githubusercontent.com/90727999/185233957-d5666719-1a1c-4547-87fc-8a1395eddae9.mov

agustinbusso (approved):

hvanlear (commented):
Looks great @mcraeteisha @ryancooley! Thanks again

ryancooley (approved):

PR #1477, opened 2023-03-28
FOUR-7496: ZoomIn/Out Hotkeys

TinCodes:
## Feature
In order to provide better controls for the modeler canvas, hotkeys have been introduced. In this particular feature, the user can zoom the canvas in and out by using the keyboard. 

## User Story
As a user designer,
I want to zoom in zoom out using hotkeys
so that adjust my view easily.

## Acceptance Criteria
Given the user opens a BPMN process
When user press “Control” + “+” (“Command” + “+”)
Then the view zooms in

Given the user opens a BPMN process
When user press “Control” + “-” (“Command” + “-”)
Then the view zooms out

## Changes Made
- Hotkeys folder added containing mixins to handle all hotkey shortcuts.
- `ZoomInOut.js` file added to handle all logic regarding the Ctrl + "+" and Ctrl + "-" shortcuts.
- Encapsulated the keydown event listener from `Modeler.vue` to it's own mixin `ZoomInOut.js`
- Added canvas control variables to `Modeler.vue`.
- Created tests for zooming in and out with standard "+" key and Numeric Pad "+" key.
- Toolbar detects changes made in the paperManager scale property. (Zoom percentage)

## Related Tickets & Packages
- [FOUR-7496](https://processmaker.atlassian.net/browse/FOUR-7496)

## Code Review Checklist

[FOUR-7496]: https://processmaker.atlassian.net/browse/FOUR-7496?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/main.js
```diff
@@ -10,13 +10,15 @@ import 'bootstrap/dist/css/bootstrap.css';
...
 import Multiselect from '@processmaker/vue-multiselect/src/Multiselect';
+import VueMousetrapPlugin from 'vue-mousetrap/vue2';
 
 import '@fortawesome/fontawesome-free/css/all.min.css';
 
 Vue.use(BootstrapVue);
 Vue.use(VueDeepSet);
 Vue.use(VueI18Next);
 Vue.use(ScreenBuilder);
+Vue.use(VueMousetrapPlugin);
```
josechirivella, line 21: What is the purpose of the MouseTrap

src/components/hotkeys/main.js
```diff
@@ -0,0 +1,39 @@
+import ZoomInOut from './zoomInOut';
+import store from '@/store';
+import moveShapeByKeypress from '@/components/modeler/moveWithArrowKeys';
```
josechirivella, line 3: this could be a circular dependency since that function is created in `Modeler.vue` and we are also importing this file into `Modeler.vue`, hence, the circular dependency.
Consider moving this function to this file

josechirivella (changes requested):

src/components/hotkeys/main.js
TinCodes: `moveShapeByKeypress` is created in a separate file that `Modeler.vue` _was_ importing, but since `hotkeys/main.js` is the only file using it now then I agree that `moveWithArrowKeys.js` should be moved elsewhere

src/main.js
TinCodes: to be corrected asap

package-lock.json
josechirivella, line 1: Did you commit the package-lock.json changes since you removed the package?

josechirivella (changes requested):

package-lock.json
TinCodes: Seems like I didn't... Should I do it in a new commit?

caleeli: since we are not going to use any additional hotkeys library, please rollback the `package-lock.json`

josechirivella (approved):

caleeli (approved):

PR #1487, opened 2023-04-06
FOUR-7482: Boundary Event in task data connector move everywhere in the selection

rodriquelca:
## Issue & Reproduction Steps

1. Open modeler 
3. Drag task form 
5. Select Boundary to task 
7. Select the task form

Expected behavior: 
Boundary Event in task or data connector should not move everywhere in the selection
Actual behavior: 
Boundary Event in task or data connector move everywhere in the selection
## Solution
- Add some validations for boundary events movements.
## How to Test
Test the steps above

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-7883
## Bloqued By
- https://processmaker.atlassian.net/browse/FOUR-7833
## Code Review Checklist
- [ ] I have pulled this code locally and tested it on my instance, along with any associated packages.
- [ ] This code adheres to [ProcessMaker Coding Guidelines](https://github.com/ProcessMaker/processmaker/wiki/Coding-Guidelines).
- [ ] This code includes a unit test or an E2E test that tests its functionality, or is covered by an existing test.
- [ ] This solution fixes the bug reported in the original ticket.
- [ ] This solution does not alter the expected output of a component in a way that would break existing Processes.
- [ ] This solution does not implement any breaking changes that would invalidate documentation or cause existing Processes to fail.
- [ ] This solution has been tested with enterprise packages that rely on its functionality and does not introduce bugs in those packages.
- [ ] This code does not duplicate functionality that already exists in the framework or in ProcessMaker.
- [ ] This ticket conforms to the PRD associated with this part of ProcessMaker.

PR #1490, opened 2023-04-13
Validate the selection of boundary events

caleeli:
## Issue & Reproduction Steps
Issue when selecting a boundary event inside a pool.

https://github.com/ProcessMaker/modeler/pull/1487#pullrequestreview-1377626134

Expected behavior: 
A boundary event could only be selected alone.


ana-mauricio (approved):
The Fix is correct

PR #1512, opened 2023-04-21
FOUR-7834 Modified the behavior of client entering / leaving the paper

josechirivella:
## Issue & Reproduction Steps

Expected behavior: 
Paper panning should be stopped when Client tries to move the paper outside of it

Actual behavior: 
Paper panning is not stopped when client moves outside of the paper causing the cells to be moved to 0

## Solution
- Added a new Boolean state to set if the client has left the paper or not.
- Removed the setting the client {X,Y} from the mouse on every iteration since I noticed was causing a lot of delay and creating a long pool of Vue Changes

## How to Test
Test the steps above

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-7834

## Code Review Checklist

### DIFF
--- a/src/components/hotkeys/main.js
+++ b/src/components/hotkeys/main.js
@@ -46,7 +46,6 @@ export default {
       if (event.code === 'Space') {
         this.isGrabbing = true;
         this.paperManager.addEventHandler('blank:pointermove', (event, x, y) => {
-          // console.log({ x: event.clientX, y: event.clientY });
           if (!this.canvasDragPosition) {
             const scale = this.paperManager.scale;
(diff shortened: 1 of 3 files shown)


src/components/hotkeys/main.js
```diff
@@ -41,11 +46,12 @@ export default {
       if (event.code === 'Space') {
         this.isGrabbing = true;
         this.paperManager.addEventHandler('blank:pointermove', (event, x, y) => {
+          // console.log({ x: event.clientX, y: event.clientY });
```
TinCodes, line 49: Remove `console.log()` please

TinCodes (approved):

rodriquelca (approved):

PR #1525, opened 2023-04-28
FOUR-8326

rodriquelca:
## Issue & Reproduction Steps

Expected behavior: 

Actual behavior: 

## Solution
- the boundary events will be cloned in the final process to avoid this js issue

## How to Test

1. create a process
3. add  a task
5. attach a boundary 
7. select the boludary with shift key
9. select the task with shift key 
10. open the developer tools
11. clone or duplicate the selected shapes

https://user-images.githubusercontent.com/1401911/235174954-dae20aa1-1ea7-4801-9953-fa9e5da9a67a.mov

teds shapes

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-8326

## Code Review Checklist
- [x] I have pulled this code locally and tested it on my instance, along with any associated packages.
- [x] This code adheres to [ProcessMaker Coding Guidelines](https://github.com/ProcessMaker/processmaker/wiki/Coding-Guidelines).
- [x] This solution fixes the bug reported in the original ticket.

### DIFF
--- a/src/mixins/cloneSelection.js
+++ b/src/mixins/cloneSelection.js
@@ -42,7 +42,7 @@ export default {
             clonedDataOutputAssociations.push(clonedFlow);
             clonedNodes.push(clonedFlow);
           } else {
-            //validate boundary events and collect in clonedBoundaryEvents array
+            // Validate boundary events and collect in clonedBoundaryEvents array
             const clonedElement = this.cloneElementAndCalculateOffset(node);
             if (node.definition && node.definition.$type !== 'bpmn:BoundaryEvent') {
               clonedNodes.push(clonedElement);
@@ -52,7 +52,7 @@ export default {
           }
         });
       }
-      //attach the cloned boundary events alway at the end off the cloned nodes
+      // Sets the clonedBoundaryEvents at the end of the array
       clonedNodes = [...clonedNodes, ...clonedBoundaryEvents];
       this.connectClonedFlows(clonedFlows, clonedNodes);
       this.connectClonedDataInputAssociations(clonedDataInputAssociations, clonedNodes);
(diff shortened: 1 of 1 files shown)


ana-mauricio:
The fix is correct @rodriquelca

caleeli (approved):

src/mixins/cloneSelection.js
```diff
@@ -42,12 +42,18 @@ export default {
...
             clonedNodes.push(clonedFlow);
           } else {
+            // Validate boundary events and collect in clonedBoundaryEvents array
             const clonedElement = this.cloneElementAndCalculateOffset(node);
-            clonedNodes.push(clonedElement);
+            if (node.definition && node.definition.$type !== 'bpmn:BoundaryEvent') {
+              clonedNodes.push(clonedElement);
+            }  else {
+              clonedBoundaryEvents.push(clonedElement);
+            }
```
TinCodes, line 51: This validation could be done out side the `else` statement, as another `else if` maybe

src/mixins/cloneSelection.js
```diff
@@ -42,12 +42,18 @@ export default {
...
           }
         });
       }
-
+      // Sets the clonedBoundaryEvents at the end of the array
+      clonedNodes = [...clonedNodes, ...clonedBoundaryEvents];
```
TinCodes, line 56: Does the order affect how the cloned elements are connected afterwards? I think that maybe it is not necessary to put it at the end of the array and would be better to check the code that connects flows before adding every clone to the canvas.

Thoughts? @josechirivella @caleeli?

caleeli: @rodriquelca commented me the boundary clone needs the task clone to exists before to attach it.

PR #1531, opened 2023-05-02
FOUR-8360 Unable to pan while hovering on a pool / cell

josechirivella:
## Issue & Reproduction Steps

Expected behavior: 
User is expected to move the paper while pressing Space even if they click or drag over an element

Actual behavior: 
User wasn't able to drag elements while hovering them

## Solution
- Added comments in the code

## How to Test
Test the steps above
* Gran an element to the paper, preferably a Pool
* Start pressing space and hold it
* Start dragging from inside the pool and the paper should move without moving the pool itself

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-8360

## Code Review Checklist


src/components/modeler/setUpSelectionBox.js
josechirivella, line 1: removed this file since this function wasn't even used anymore since Selection.vue is doing this work now

rodriquelca (approved):

PR #1667, opened 2023-09-08
FOUR-8066 Process Testing part 1

caleeli:
## Process Testing part 1

- Run a process in an isolated environment.
- Save scenarios
- Run scenarios
- Run processes from modeler


caleeli:
A test was added to cover half of the changes: tests/unit/components/toolbar/toolbarExtraMenu.spec.js
But it was counted as new code to be covered.
Code added to Modeler.vue was not been able to test because require a global configuration to enable extra packages.

PR #1677, opened 2023-09-25
FOUR-10728 Check if the config starts with { before JSON parse it

caleeli:
Modeler is always parsing the pm:config as a JSON string. But when a Signal Start Event is used, it uses the pm:config as string and not as a JSON.

## Issue & Reproduction Steps

- Create a process with Signal Start Event
- Set the request variable in the inspector of the signal start event
- Save the process
- When reopened it will throw the exception.

## Solution
- Validate if the pm:config is a JSON string before to parse it

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-10728
- https://processmaker.atlassian.net/browse/FOUR-10713

## Code Review Checklist

### DIFF
--- a/src/components/modeler/Modeler.vue
+++ b/src/components/modeler/Modeler.vue
@@ -850,7 +850,7 @@ export default {
 
       this.removeUnsupportedElementAttributes(definition);
 
-      const config = definition.config && definition.config.startsWith('{') ? JSON.parse(definition.config) : {};
+      const config = definition.config && _.isJSON(definition.config) ? JSON.parse(definition.config) : {};
       const type = config?.processKey || parser(definition, this.moddle);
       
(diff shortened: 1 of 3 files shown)


src/components/modeler/Modeler.vue
```diff
@@ -850,7 +850,7 @@ export default {
 
       this.removeUnsupportedElementAttributes(definition);
 
-      const config = definition.config ? JSON.parse(definition.config) : {};
+      const config = definition.config && definition.config.startsWith('{') ? JSON.parse(definition.config) : {};
```
danloa, line 853: lodash  _.isJSON( value ) can be an improvement here

danloa (approved):

PR #1757, opened 2023-12-01
FOUR-11972 Getting the correct value assigned for eventDefinitions

josechirivella:
## Issue & Reproduction Steps

Expected behavior: 
Boundary Timer Event should keep and show the value previously set or default in reload

Actual behavior: 
Boundary Timer Event wasn't getting the correct value previously set or default since we were assuming that in index 1, the `timeDefinition` was going to be there. 

## Solution
- Getting the correct eventDefinitions type. Using the `omit` from lodash to omit all the objects except the possibilities of timer objects. There could be an argument to use `pick` instead.

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-11972
- https://processmaker.atlassian.net/browse/FOUR-12517

## Code Review Checklist

ci:deploy


tokensPM:
QA server was successfully deployed https://ci-d6e2fe00e3.eng.processmaker.net

PR #1842, opened 2024-06-18
FOUR-16765: The API calls in modeler is execute twice

henryjonathanquispe:
## Issue & Reproduction Steps

1. Go to modeler 
2. Create a process and a task with Screens and Display Next Assigned Task to Task Assignee enable
3. Open the debuger
4. Click on any task and see th calls in Network

## Solution
- Add validation to Inspector Panel in Modeler

## How to Test
- Go to modeler 
- Create a process and a task with Screens and Display Next Assigned Task to Task Assignee enable
- Open the browser debugger
- Click on any task and see the calls in Network

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-16765

## Code Review Checklist

ci:next


src/components/inspectors/InspectorPanel.vue
```diff
@@ -91,10 +92,12 @@ export default {
     };
   },
   watch: {
-    highlightedNode() {
-      document.activeElement.blur();
-      this.prepareData();
-      this.prepareConfig();
+    highlightedNode(nVal, oVal) {
+      if (!isEqual(nVal, oVal)){
```
rodriquelca, line 96: add space before  {

src/components/inspectors/InspectorPanel.vue
```diff
@@ -91,10 +92,12 @@ export default {
     };
   },
   watch: {
-    highlightedNode() {
-      document.activeElement.blur();
-      this.prepareData();
-      this.prepareConfig();
+    highlightedNode(nVal, oVal) {
```
rodriquelca, line 95: use newValue and oldValue for better understand

src/components/inspectors/InspectorPanel.vue
caleeli: Why are you using `isEqual`
If the `nVal` and `oVal` are objects what happens if they are pointing to the same object? isEqual wil always return `true`.
Does it could happen? If that is the case the blur and prepare actions will not be executed in that case.

henryjonathanquispe: I am using isEqual because the comparation is deep with this function,
The modeler fires two selection events twice, once when the mouse presses down and up,
This triggers calls to the inspector twice.
document.activeelement is related to the body and not to the modeler

caleeli (approved):
The change looks good, but seems needs a rebase, since some tests are failing

PR #1939, opened 2025-09-08
FOUR-25964: There is a loading in the message in the process modeler in multi tenant

rodriquelca:
## Issue & Reproduction Steps

Expected behavior: 
There is not should be a loading on the IA message
Actual behavior: 
There is a loading in the AI ​​message in the process modeler
## Solution
to support tenant aware and server validation by hostname 
- improve microservise ai 
- improve the package ai 
- improve ai wellcome message component 
## How to Test
-  see the related ticket

## Related Tickets & Packages
- https://processmaker.atlassian.net/browse/FOUR-25964

## Code Review Checklist


agustinbusso (changes requested):
Probably we should send the server in the backend side, to avoid any security problem and for promptsessionid add a change just in one place

